PR #13, opened 2018-04-02
Feature/12 mention feature type

reckart:
Subject, predicate and object can now be extracted in Annotation Page (Subject and Object share a featureSupport and featureEditor while predicate has its own featureSupport and featureEditor)


inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/PropertyFeatureSupport.java
```diff
@@ -0,0 +1,193 @@
...
+import de.tudarmstadt.ukp.inception.kb.graph.KBHandle;
+import de.tudarmstadt.ukp.inception.kb.graph.KBProperty;
+
+@Component
+public class PropertyFeatureSupport
+    implements FeatureSupport<Void>
+{
+    private static final Logger LOG = LoggerFactory.getLogger(PropertyFeatureSupport.class);
+    private static final String PREDICATE_KEY = "Property: Predicate";
+    private static final String FACT_PREDICATE_PREFIX = "fact-predicate:";
```
reckart: @matchlesswei Given that we use the prefix "kb:" for concept features, how about using `kb-property` here - after all, what we can choose from here are properties (not predicates).

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/PropertyFeatureSupport.java
```diff
@@ -0,0 +1,193 @@
...
+import de.tudarmstadt.ukp.inception.kb.KnowledgeBaseService;
+import de.tudarmstadt.ukp.inception.kb.graph.KBHandle;
+import de.tudarmstadt.ukp.inception.kb.graph.KBProperty;
+
+@Component
+public class PropertyFeatureSupport
+    implements FeatureSupport<Void>
+{
+    private static final Logger LOG = LoggerFactory.getLogger(PropertyFeatureSupport.class);
+    private static final String PREDICATE_KEY = "Property: Predicate";
```
reckart: How about we use "KB: Property" here. I'd also change the concept feature support to generate "KB: Concept".

reckart:
I did a bit of local testing and observed the following:
* the subject/object slots auto-arm - this is actually nice :)
* the slot re-arm even if I explicitly disarm them - maybe it would nice to have some way to say "there is nothing to fill into this slot" and then the slot should no longer auto-arm - we could open a separate issue for that.
* the subject/object slots cannot be cleared after they have been assigned - WDYT? How about adding a trashcan icon next to the slot for clearing it? Fix it in this PR or open a new one for it?
* minor layout issues in the feature detail editor
* if both, the subject and object slots are empty, and one is armed, I can click on the other and it also arms (i.e. both end up being orange). If I re-load the page, only the one clicked latest remains orange. This means the arming-mechanism works ok, but when arming a slot the entire detail panel needs to be redrawn, not just the editor into which was clicked (call `getOwner()` in the FeatureDetailEditor to fetch the entire panel for redrawing). I think this should be easy to fix and still go into the present PR.

reckart:
* In the Subject/Object dropdowns, it is possible to choose classes or instances. If I choose a class, then the values is rendered as "ERROR" in annotation bubble in the main annotation area.

matchlesswei:
@reckart thanks for the changes.

> In the Subject/Object dropdowns, it is possible to choose classes or instances. If I choose a class, then the values is rendered as "ERROR" in annotation bubble in the main annotation area.

The value is rendered as "ERROR" because this value is not allowed in the linked layer. Would it be better for me to change the code so that the dropdown in Subject/Object only shows values that are allowed in the linked layer?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/PropertyFeatureSupport.java
matchlesswei: changed.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/PropertyFeatureSupport.java
matchlesswei: changed.

matchlesswei:
@reckart when I tried mvn clean install, I failed in inception-ui-kb because:

```
[WARNING] The POM for eu.openminted.share.annotations:omtd-share-annotations-api:jar:3.0.2.2-SNAPSHOT is missing, no dependency information available
[ERROR] Failed to execute goal on project inception-ui-kb: Could not resolve dependencies for project de.tudarmstadt.ukp.inception.app:inception-ui-kb:jar:0.3.0-SNAPSHOT: Failure to find eu.openminted.share.annotations:omtd-share-annotations-api:jar:3.0.2.2-SNAPSHOT in http://zoidberg.ukp.informatik.tu-darmstadt.de/artifactory/public-snapshots was cached in the local repository, resolution will not be reattempted until the update interval of ukp-oss-snapshots has elapsed or updates are forced 
```
Should I add the dependency to POM?

reckart:
I have added missing Maven repos to the pom - that should fix the dependency problem.

reckart:
Regarding the "ERROR" problem: I think the best thing would be to "steal" the feature editor from the linked layer. I.e. if we link to the feature "KBItems" on the layer "LinkedEntity" then we should obtain a feature support for "KBItems" and use `createEditor` to retrieve the feature editor for it and then embed this editor into the SubjectObjectFeatureEditor. WDYT?

matchlesswei:
> the subject/object slots cannot be cleared after they have been assigned - WDYT? How about adding a trashcan icon next to the slot for clearing it? Fix it in this PR or open a new one for it?

I'm working on this right now.
But If I add a icon here, I cannot click to enable and disable "select to fill". @reckart Am I setting the trash bin in wrong position?
```
<div wicket:id="content">
        <div class="form-group">
          <div class="col-sm-10">
            <span wicket:id="label" class="form-control text-nowrap feature-editor-slot-label">
            </span>
          </div>
            <div style="position: relative; top:0.15em">
                <button wicket:id="removeLabel" type="submit" class="btn-xs btn-danger"
                    value="-">
                    <i class="fa fa-trash" aria-hidden="true"></i>
                </button>
            </div>
        </div>
```

reckart:
I'd suggest this:

```
        <div class="form-group">
          <div class="col-sm-12">
            <div class="input-group">
              <span wicket:id="label" class="form-control text-nowrap feature-editor-slot-label"></span>
              <span class="input-group-btn">
                <button wicket:id="removeLabel" type="button" class="btn btn-default">
                  <i class="fa fa-trash" aria-hidden="true"></i>
                </button>
              </span>
            </div>
          </div>
        </div>
```

Note that the button is `type="button"`, not `type="submit"` and in Wicket you should use an LambdaAjaxLink (not a LambdaAjaxButton) to handle the click.

reckart:
Let's do the trashcan thingy and then we merge this. I have opened separate issues for the remaining two comments I had.

matchlesswei:
Thanks! if the button is inside `<span>`, it works. But if it is inside `<div>`, it doesn't work. Why do they have this difference?

reckart:
> Thanks! if the button is inside <span>, it works. But if it is inside <div>, it doesn't work. Why do they have this difference?

I do not have the slightest idea!

matchlesswei:
The trashcan is done.

PR #24, opened 2018-04-06
Feature/18 skipped suggestions come back

reckart:


inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -92,11 +93,23 @@ public RecommendationDifference generateRecommendationWithLowestDifference(Docum
...
+    {
+        getRecommendationFromRecommendationModel();
+        removeRejectedOrSkippedAnnotations(false, null);
+        if (listOfRecommendationsForEachToken.size() > 0) {
+            return true;
+        }
+        else {
+            return false;
+        }
+    }
```
reckart: This could be done as 

```
return !listOfRecommendationsForEachToken.isEmpty();
```

The same applies to other places where something like this happens:

```
if (condition) {
  return true;
}
else {
  return false;
}
```

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -174,36 +187,48 @@ private boolean isAlreadyInCleanList(List<AnnotationObject> cleanRecommendationL
...
     {
         List<LearningRecord> records = learningRecordService.getAllRecordsByDocumentAndUserAndLayer
             (annotatorState.getDocument(), annotatorState.getUser().getUsername(), selectedLayer);
         for (List<AnnotationObject> recommendations: listOfRecommendationsForEachToken) {
             recommendations.removeIf(recommendation -> doesContainRejectedOrSkippedRecord(records,
-                recommendation));
+                recommendation, filterSkippedRecommendation, learnSkippedRecommendationTime));
         }
         listOfRecommendationsForEachToken.removeIf(recommendationsList ->
             recommendationsList.size() == 0);
```
reckart: Using `isEmpty()` is generally considered better than `== 0` to check if a collection is empty.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -174,36 +187,48 @@ private boolean isAlreadyInCleanList(List<AnnotationObject> cleanRecommendationL
...
     {
         boolean flag = false;
         for (LearningRecord record: records) {
-            if ((record.getUserAction().equals("rejected") || record.getUserAction().equals
-                ("skipped")) && record.getSourceDocument().getName()
-                .equals(aRecommendation.getDocumentName()) && record.getOffsetTokenBegin() ==
-                aRecommendation.getOffset().getBeginToken() && record.getOffsetTokenEnd() ==
-                aRecommendation.getOffset().getEndToken() && record.getAnnotation().equals
-                (aRecommendation.getAnnotation())) {
+            if ((record.getUserAction().equals("rejected") || (
```
reckart: Better introduce constants for these string values instead of repeating literals wherever they are used.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -174,36 +187,48 @@ private boolean isAlreadyInCleanList(List<AnnotationObject> cleanRecommendationL
...
+                .getAnnotation().equals(aRecommendation.getAnnotation())) {
                 flag = true;
                 break;
             }
         }
         return flag;
     }
 
+    private boolean needFilterByTime (Date learnSkippedTime, LearningRecord record) {
+        if (learnSkippedTime != null && learnSkippedTime.compareTo(record.getActionDate()) == 1) {
```
reckart: Could be written as 

```
return learnSkippedTime != null && learnSkippedTime.compareTo(record.getActionDate()) == 1;
```

But I am not sure what this code is actually doing. A comment would be nice. Also, reordering might allow for skipping the null check, e.g. if it is guaranteed that `record.getActionDate()` always returns non-null.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -380,8 +407,9 @@ public AnnotationObject generateRecommendationWithLowestConfidence(JCas aJcas)
         if (recommendations.size() > 0) {
             return recommendations.get(0);
         }
-        else
+        else {
             return null;
+        }
```
reckart: Consider using `Optional.of(recommendations.get(0)` or even `recommendations.stream().findFirst()` which would allow for omitting the check if the collection is empty.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.html
```diff
@@ -32,20 +32,27 @@ <h3 class="panel-title">Session</h3>
...
               <div class="col-sm-9">
-                <select class="form-control"  wicket:id="selectLayer"></select>
+                <select class="form-control" wicket:id="selectLayer"></select>
                 <span class="form-control" wicket:id="showSelectedLayer"></span>
               </div>
             </div>
           </div>
           <div class="panel-footer text-right">
-            <input wicket:id="layerSelectionButton" type="submit" class="btn btn-primary" value="Start"/>
+            <input wicket:id="layerSelectionButton" type="submit" class="btn btn-primary"
```
reckart: Usually we use a `<button type"submit"...>` in stead of an `input`. But practically, it probably does not matter.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -302,12 +315,43 @@ private Label createNoRecommendationLabel() {
             @Override
             public boolean isVisible()
```
reckart: Do not overwrite `isVisible`. Instead implement `onConfigure()`:

```
                @Override
                protected void onConfigure()
                {
                    super.onConfigure();
                    setVisible(startLearning && !hasUnseenRecommendation && !hasSkippedRecommendation);
                }
```

We even have a `LambdaBehavior` now which could be used without even creating an anonymous subclass:

```
Label noRecommendation = new Label("noRecommendationLabel", "There are no further suggestions.")
  .add(LambdaBehavior.onConfigure(_this -> _this.setVisible(startLearning && !hasUnseenRecommendation && !hasSkippedRecommendation));
```

When a component can become invisible, it is usually a good idea to have it output a placeholder, otherwise its visibility status cannot be changed via AJAX:

```
Label noRecommendation = new Label("noRecommendationLabel", "There are no further suggestions.")
  .add(LambdaBehavior.onConfigure(_this -> _this.setVisible(startLearning && !hasUnseenRecommendation && !hasSkippedRecommendation));
noRecommendation.setOutputMarkupPlaceholderTag(true);
```

Same applies to other places where isVisible() is overwritten.

reckart:
@matchlesswei I made some comments on the code. Since I opened the PR, GitHub doesn't let me *request changes*.

matchlesswei:
@reckart okay, I'll change the codes.

reckart:
@matchlesswei be sure to check this branch out freshly! I changed it's history when I updated your committer ID.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
reckart: Groovy also uses "it". I'm a bit torn between generic names like "_this" or "it" and using a more speaking name like "component".

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: have added a enum type for LearningRecordUserAction and replaced all the user action string values.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: have added a comment and set ActionDate nullable = false

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: have changed to use Optional to omit check if the collection is empty

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: have changed `== 0` to `isEmpty()`

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: have changed to directly return boolean condition to avoid if... else...

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -257,10 +282,12 @@ private boolean fromMoreThanOneClassifier(List<AnnotationObject> recommendationL
                 numberOfOpenNLPClassifier++;
             }
         }
-        if (numberOfOpenNLPClassifier >= 1 && numberOfStringMatchingClassifer >= 1)
+        if (numberOfOpenNLPClassifier >= 1 && numberOfStringMatchingClassifer >= 1) {
```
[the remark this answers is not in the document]
matchlesswei: have changed.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -92,11 +93,23 @@ public RecommendationDifference generateRecommendationWithLowestDifference(Docum
...
 
         return calculateDifferencesAndReturnLowestDifference ();
     }
 
+    public boolean hasRecommendationWhichIsSkipped()
+    {
+        getRecommendationFromRecommendationModel();
+        removeRejectedOrSkippedAnnotations(false, null);
+        if (listOfRecommendationsForEachToken.size() > 0) {
+            return true;
```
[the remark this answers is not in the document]
matchlesswei: have changed.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -70,8 +71,8 @@ public ActiveLearningRecommender(RecommendationService recommendationService,
         this.learningRecordService = recordService;
     }
 
-    public RecommendationDifference generateRecommendationWithLowestDifference(DocumentService
-                                                                              documentService)
+    public RecommendationDifference generateRecommendationWithLowestDifference(
+        DocumentService documentService, Date learnSkippedRecommendationTime)
```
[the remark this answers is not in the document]
matchlesswei: have changed.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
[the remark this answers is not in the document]
matchlesswei: have changed.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -174,36 +187,48 @@ private boolean isAlreadyInCleanList(List<AnnotationObject> cleanRecommendationL
...
         boolean flag = false;
         for (LearningRecord record: records) {
-            if ((record.getUserAction().equals("rejected") || record.getUserAction().equals
-                ("skipped")) && record.getSourceDocument().getName()
-                .equals(aRecommendation.getDocumentName()) && record.getOffsetTokenBegin() ==
-                aRecommendation.getOffset().getBeginToken() && record.getOffsetTokenEnd() ==
-                aRecommendation.getOffset().getEndToken() && record.getAnnotation().equals
-                (aRecommendation.getAnnotation())) {
+            if ((record.getUserAction().equals("rejected") || (
+                record.getUserAction().equals("skipped") && filterSkippedRecommendation)
```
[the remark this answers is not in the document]
matchlesswei: have moved the condition to a new method

reckart:
@matchlesswei please push the changes here

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
matchlesswei: changed

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
matchlesswei: I think in my case, Optional is not the most necessary. I don't want to avoid null because AL sidebar needs it to check whether to show the suggestion or no suggestion label. 

I can also use Optional.isPresent to check, but then there is a serializable problem if I use it in AL sidebar.

matchlesswei:
@reckart I think it could be merged.

reckart:
Look ok. I'll merge it.

Issue #30, opened 2018-04-10
Pre-disambiguate concepts in dropdown list


reckart:
When annotating an ontology concept in the text, optionally re-order the concepts in the dropdown based on the similarity of the concept description with the text surrounding the occurrence to be annotated.

* [ ]  make use of concept description
* [ ]  make use of context of annotated occurrences of the concept

reckart:
@rcffc what would be the milestone for this one?

PR #31, opened 2018-04-10
#30 - Pre-disambiguate concepts in dropdown list

reckart:
- Introduced new concept linking module
- Added flag to KB to indicate whether it supports entity linking
- Call out to concept linking code in ConceptFeatureEditor if the KB supports it

- [ ] ability to change "support entity linking" flag after a KB has been created.


reckart:
@rcffc I fixed the build, but I didn't yet get around to testing in the running application. How does the code behave if the resources it tried to load from the classpath are not present?

rcffc:
@reckart It will just ignore those parts of the code.

reckart:
It is not possible to change the "supportConceptLinking" flag after the KB has been saved.

reckart:
I opened a separate issue for editing the concept linking flag: https://github.com/inception-project/inception/issues/40

PR #49, opened 2018-04-12
Feature/48 filter candidates

reckart:
- Replace combobox by an auto-complete field

### DIFF
--- a/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
+++ b/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
@@ -260,7 +260,9 @@ public class ConceptLinkingService
         List<Token> mentionContext = getMentionContext(mentionSentence, splitMention,
             MENTION_CONTEXT_SIZE);
 
-        candidates.parallelStream().forEach(l -> {
+
+        List<CandidateEntity> result = new ArrayList<>((candidates));
+        result.parallelStream().forEach(l -> {
             String wikidataId = l.getIRI().replace(WIKIDATA_PREFIX, "");
 
             if (entityFrequencyMap != null && entityFrequencyMap.get(wikidataId) != null) {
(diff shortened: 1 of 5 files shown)


reckart:
I think we have to add a parameter to the KB listXXX methods in order to limit the number of returned concepts. For a large/remove KB, there is a very significant delay.

rcffc:
That is true. However, there is another big problem. In contrast to entity linking in german, the number of candidates in english is much larger. Therefore we might not get the correct entity in the candidate generation step. For example, the correct entity Q142 for France is only retrieved when we set the limit to ~2500 candidates, which leads to a retrieval time of 2 seconds. (At the moment the limit is set to 1000 candidates).

rcffc:
We might have to fall back on concept filtering for for remote KBs to make the response time acceptable...

reckart:
@rcffc maybe we can also cache data

reckart:
@rcffc caching would need to be somewhat persistent because it is also needed during rendering

rcffc:
I am currently evaluating the time of the ranking pipeline steps - will post an update soon.

rcffc:
The main culprit of the overall slowness was the retrieval of the semantic signature for the computation of two ranking features. After removing the querying of the semantic signature, the time for ranking drops almost to zero. 
I have also tested candidate retrieval time with (10, 100, 1000, 2500, 10000) candidates, which yielded (0.3s, 1.6s, 2s, 2.1s, 2.7s).

reckart:
Cool :)

Not sure what you think, but I feel we do not need hundreds of suggestions for the user to scroll through in the autocomplete. So even if the backend can quickly retrieve a large number, IMHO it would be sufficient to limit to number actually sent to the frontend.

rcffc:
If the number of candidates affects the response time, I would agree.

rcffc:
I am thinking that the AutoCompleteBoxField should come "preloaded" with the marked mention. Do you think this is possible?

reckart:
> I am thinking that the AutoCompleteBoxField should come "preloaded" with the marked mention. Do you think this is possible?

Hm. Sounds like a good idea. Might be a bit tricky to implement though. I'd suggest you open a separate issue and explain the expected behavior.

rcffc:
Jenkins, which check has failed?

reckart:
@rcffc There is a link "details" which takes you to the logging output.

reckart:
Correction, it takes you to the build. Then you need to check the "console output".

```
[INFO] --- maven-dependency-plugin:3.0.2:analyze-only (default) @ inception-ui-kb ---
[WARNING] Used undeclared dependencies found:
[WARNING]    com.googlecode.wicket-jquery-ui:wicket-jquery-ui-core:jar:7.10.1:compile
```

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.0.2:analyze-only (default) on project inception-ui-kb: Dependency problems found -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :inception-ui-kb
```

reckart:
@rcffc wrong dependency added ;)

PR #51, opened 2018-04-14
Feature/50 statement editor with validation

reckart:


Rentier:
@reckart  What is the status of this? It is still marked as WIP.

reckart:
This is the last state @mbugert was working. For the moment, it is orphaned.

reckart:
Compiles again - doesn't work at all.

reckart:
@naveen2507 any progress here on your machine that you didn't push?

naveen2507:
@reckart Nothing to push on this branch, as we need changes for queries before that as we discussed. I am working on this branch [https://github.com/inception-project/inception/tree/feature/322-refactor-sparql-queries](https://github.com/inception-project/inception/tree/feature/322-refactor-sparql-queries) for changes.

naveen2507:
@reckart Can you review this. Can't add you as reviewer here as you are the author for this PR.

reckart:
I don't seem to be getting far in testing this PR because I cannot create statements: https://github.com/inception-project/inception/issues/422

naveen2507:
@reckart Issue is fixed, you can test it now.

Rentier:
Please add extensive unit tests whenever you change the `KnowledgeBaseService`.

naveen2507:
@reckart @Rentier Can any one of you review the PR as its a major change which would be needed for changes in other PRs. Merging the branch would double up the work for review .

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -174,25 +175,25 @@ public KnowledgeBaseServiceRemoteTest(TestConfiguration aConfig)
...
-//            kb_wikidata_reified.setDescriptionIri(vf.createIRI("http://schema.org/description"));
-//            kbList.add(kb_wikidata_reified);
-//        }
+        // {
+        // ValueFactory vf = SimpleValueFactory.getInstance();
+        //
+        // KnowledgeBase kb_wikidata_reified = new KnowledgeBase();
+        // kb_wikidata_reified.setName("Wikidata (official/reified)");
+        // kb_wikidata_reified.setType(RepositoryType.REMOTE);
+        // kb_wikidata_reified.setReification(Reification.WIKIDATA);
```
Rentier: Please delete commented code. If we need it back, then we can ask git for it.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -256,6 +266,23 @@ public void thatPropertyListCanBeRetrieved()
 
     }
 
+    @Test
+    public void thatParentListCanBeRetireved()
```
Rentier: `retrieved`

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -256,6 +266,23 @@ public void thatPropertyListCanBeRetrieved()
...
+    {
+        KnowledgeBase kb = sutConfig.getKnowledgeBase();
+        
+        long duration = System.currentTimeMillis();
+        Set<KBHandle> parentList = sut.getParentConceptList(kb, sutConfig.getTestIdentifier(), true);
+        duration = System.currentTimeMillis() - duration;
+
+        System.out.printf("Parent List retrieved : %d%n", parentList.size());
```
Rentier: Please use a logger for this or remove it.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -256,6 +266,23 @@ public void thatPropertyListCanBeRetrieved()
...
+        System.out.printf("Time required        : %d ms%n", duration);
+        parentList.stream().limit(10).forEach(h -> System.out.printf("   %s%n", h));
+
+        assertThat(parentList).as("Check that parent list is not empty").isNotEmpty();
+
```
Rentier: Whitespace

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.java
```diff
@@ -199,6 +216,14 @@ public ExistingStatementGroupFragment(String aId) {
                     this::actionPropertyLinkClicked);
             propertyLink.add(new Label("property", groupModel.bind("property.uiLabel")));
             form.add(propertyLink);
+
+            // TODO what about handling type intersection when multiple range statements are
```
Rentier: Can you solve this TODO? If not, open an issue.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.java
```diff
@@ -166,10 +169,24 @@ public NewStatementGroupFragment(String aId) {
...
+                Set<KBHandle> parentConceptList = kbService.getParentConceptList(
+                        groupModel.getObject().getKb(), bean.getInstance().getIdentifier(), true);
+                
+                for (KBHandle parent : parentConceptList) {
+                    properties.addAll(kbService.listDomainProperties(groupModel.getObject().getKb(),
+                            parent.getIdentifier(), true, true));
+                }
+
+                // Kept here to avoid fail case scenario for WikiData
+                if (properties.isEmpty()) {
```
Rentier: Please explain what it is doing in addition to why

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/editor/StatementEditor.java
```diff
@@ -0,0 +1,369 @@
...
+                target.focusComponent(editMode.getFocusComponent());
+            }
+            content = editMode;
+        } else {
+            content = new ViewMode(CONTENT_MARKUP_ID, statement);
+        }
+        add(content);
+    }
+
+    protected void actionEdit(AjaxRequestTarget aTarget) {
```
Rentier: These are not DKPro Style braces.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueType.java
```diff
@@ -0,0 +1,95 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.ui.kb.value;
+
+import java.io.Serializable;
+
+public class ValueType
```
Rentier: Please document what this class is doing.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupport.java
```diff
@@ -0,0 +1,63 @@
...
+import de.tudarmstadt.ukp.inception.ui.kb.value.editor.ValuePresenter;
+
+/**
+ * A {@link ValueTypeSupport} reports if it supports a given datatype (string, int, etc., identified
+ * by IRIs). It provides Wicket components for presenting and editing values of supported datatypes.
+ */
+public interface ValueTypeSupport
+    extends BeanNameAware
+{
+    String getId();
```
Rentier: Javadoc

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupport.java
```diff
@@ -0,0 +1,63 @@
...
+    
+    List<ValueType> getSupportedValueTypes();
```
Rentier: Javadoc

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupport.java
```diff
@@ -0,0 +1,63 @@
...
+    
+    boolean accepts(KBStatement aStatement, KBProperty aProperty);
```
Rentier: Javadoc

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupport.java
```diff
@@ -0,0 +1,63 @@
...
+
+    boolean accepts(String range, Optional<KBObject> rangeKbObject);
```
Rentier: Javadoc

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupport.java
```diff
@@ -0,0 +1,63 @@
...
+     * 
+     * @param aId
+     *            Wicket markup id received by the editor instances
+     * 
+     * @return a {@link ValueEditor} instance
+     */
+    ValueEditor createEditor(String aId, IModel<KBStatement> aModel, IModel<KBProperty> aDatatype,
+            IModel<KnowledgeBase> kbModel);
+
+    ValuePresenter createPresenter(String aId, IModel<KBStatement> aModel,
```
Rentier: Javadoc

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+            List<ValueType> types = valueSupport.getSupportedValueTypes();
+            types.stream().forEach(allTypes::add);
+        }
+
+        allTypes.sort(comparing(ValueType::getUiName));
+
+        return allTypes;
+    }
+    
+    default List<ValueType> getRangeTypes(String range, Optional<KBObject> rangeKbObject)
```
Rentier: Why do you move so much logic into an interface? Defaults should not be use like non-abstract methods in abstract classes.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+    
+    @EventListener
+    public void onContextRefreshedEvent(ContextRefreshedEvent aEvent)
+    {
+        init();
+    }
+    
+    public void init()
+    {
+        List<ValueTypeSupport> fsp = new ArrayList<>();
```
Rentier: What is fsp?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+        String datatype = null;
+        
+        if (aStatement.getValue() != null) {
+            Class<?> clazz = aStatement.getValue().getClass();
+            IRI type = DefaultDatatypeMapper.getDatatypeURI(clazz);
+            datatype = type != null ? type.stringValue() : null;
+        }
+        
+        if (datatype == null && aProperty != null) {
+            if (aProperty.getRange() != null) {
```
Rentier: You can move that in one if.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+    
+    private String getDataType(KBStatement aStatement, KBProperty aProperty)
+    {
+        String datatype = null;
+        
+        if (aStatement.getValue() != null) {
+            Class<?> clazz = aStatement.getValue().getClass();
+            IRI type = DefaultDatatypeMapper.getDatatypeURI(clazz);
+            datatype = type != null ? type.stringValue() : null;
```
Rentier: `datatype` is already `null`, no need to make it more null.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+            datatype = XMLSchema.STRING.stringValue();
+        }
+        
+        return datatype;
+    }
+    
+    @Override
+    public ValueTypeSupport getValueSupport(KBStatement aStatement, KBProperty aProperty)
+    {
+        ValueTypeSupport support = null;
```
Rentier: Assign it directly in line128.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Override
+    public ValueTypeSupport getValueSupport(ValueType type)
+    {
+        ValueTypeSupport support = null;
+                
+        for (ValueTypeSupport s : getValueSupports()) {
+            if (s.getSupportedValueTypes().contains(type)) {
+                support = s;
+                break;
```
Rentier: You can return here and then throw the exception without a null check.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistryImpl.java
```diff
@@ -0,0 +1,166 @@
...
+        // Determine the data type
+        String datatype = getDataType(aStatement, aProperty);
+        
+        support = supportCache.get(datatype);
+        
+        for (ValueTypeSupport s : getValueSupports()) {
+            if (s.accepts(aStatement, aProperty)) {
+                support = s;
+                supportCache.put(datatype, s);
+                break;
```
Rentier: You can return here and then throw the exception without a null check.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/BooleanLiteralValueEditor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.wicket.Component;
+import org.apache.wicket.markup.html.form.CheckBox;
+import org.apache.wicket.model.CompoundPropertyModel;
+import org.apache.wicket.model.IModel;
+
+import de.tudarmstadt.ukp.clarin.webanno.support.lambda.LambdaAjaxFormComponentUpdatingBehavior;
+import de.tudarmstadt.ukp.inception.kb.graph.KBStatement;
+
+public class BooleanLiteralValueEditor extends ValueEditor
```
Rentier: Not DKPro style

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/BooleanLiteralValueEditor.java
```diff
@@ -0,0 +1,49 @@
...
+    
+    public BooleanLiteralValueEditor(String aId, IModel<KBStatement> aModel)
+    {
+        super(aId, CompoundPropertyModel.of(aModel));
+        
+        value = new CheckBox("value");
+        value.setOutputMarkupId(true);
+        value.add(new LambdaAjaxFormComponentUpdatingBehavior("change", t -> t.add(getParent())));
+        add(value);
+
```
Rentier: Whitespace

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/BooleanLiteralValueEditor.java
```diff
@@ -0,0 +1,49 @@
...
+        
+        value = new CheckBox("value");
+        value.setOutputMarkupId(true);
+        value.add(new LambdaAjaxFormComponentUpdatingBehavior("change", t -> t.add(getParent())));
+        add(value);
+        
+    }
+
+    @Override
+    public Component getFocusComponent() {
```
Rentier: DKPro style

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/BooleanLiteralValuePresenter.java
```diff
@@ -0,0 +1,35 @@
...
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.ui.kb.value.editor;
+
+import org.apache.wicket.markup.html.basic.Label;
+import org.apache.wicket.model.IModel;
+
+import de.tudarmstadt.ukp.inception.kb.graph.KBStatement;
+
+public class BooleanLiteralValuePresenter extends ValuePresenter
```
Rentier: DKPro style

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/IRIValueEditor.java
```diff
@@ -0,0 +1,116 @@
...
+import de.tudarmstadt.ukp.inception.kb.model.KnowledgeBase;
+
+public class IRIValueEditor
+    extends ValueEditor
+{
+    private static final long serialVersionUID = -1646737090861147804L;
+    
+    private @SpringBean KnowledgeBaseService kbService;
+    
+    AutoCompleteTextField<KBHandle> value;
```
Rentier: Why is it not private?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseService.java
```diff
@@ -399,4 +411,15 @@ RepositoryImplConfig getKnowledgeBaseConfig(KnowledgeBase kb)
     boolean statementsMatchSPO(KnowledgeBase akb, KBStatement mockStatement);
 
     Optional<KBObject> readKBIdentifier(Project aProject, String aIdentifier);
+
+    List<KBHandle> getParentConceptsForConcept(KnowledgeBase aKB, String aIdentifier,
+            boolean aAll)
+        throws QueryEvaluationException;
+    
+    Set<KBHandle> getParentConceptList(KnowledgeBase aKB, String aIdentifier, boolean aAll)
```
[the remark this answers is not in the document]
naveen2507: For making it a distinct list

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -686,7 +684,47 @@ public KBHandle update(KnowledgeBase kb, UpdateAction aAction)
...
+            String QUERY = SPARQLQueryStore.PROPERTYLIST_DOMAIN_DEPENDENT;
+            TupleQuery tupleQuery = conn.prepareTupleQuery(QueryLanguage.SPARQL, QUERY);
+            tupleQuery.setBinding("aDomain", vf.createIRI(aDomain));
+            tupleQuery.setBinding("pLABEL", kb.getLabelIri());
+            tupleQuery.setIncludeInferred(aIncludeInferred);
 
+            return evaluateListQuery(tupleQuery, aAll);
+        });
+        
+        resultList.sort(Comparator.comparing(KBObject::getUiLabel));
```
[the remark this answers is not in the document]
naveen2507: May be we could discuss the pros and cons for it. In general, right now we don't do the sorting in SPARQL

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -767,6 +805,90 @@ public boolean hasChildConcepts(KnowledgeBase aKB, String aParentIdentifier, boo
...
+        return resultList;
+    }
+    
+    
+    @Override
+    public Set<KBHandle> getParentConceptList(KnowledgeBase aKB, String aIdentifier, boolean aAll)
+        throws QueryEvaluationException
+    {
+        Set<KBHandle> parentConceptList = new HashSet<KBHandle>();
+        if (aIdentifier != null) {
```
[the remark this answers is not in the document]
naveen2507: It will not go into the block for nesting if ```aIdentifier == null```

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
```diff
@@ -43,5 +43,58 @@
             , "    FILTER(LANG(?l) = \"\" || LANGMATCHES(LANG(?l), \"en\"))"
             , "  }"
             , "}"
-            , "LIMIT 10000");  
+            , "LIMIT 10000");
```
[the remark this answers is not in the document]
naveen2507: Its in the other PR. I would need to take care of this once, the PR approves.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/ValueTypeSupportRegistry.java
naveen2507: It was as per the structure defined. And in any case the default method doesn't need any specific implementation and are pretty generic here.

naveen2507:
@reckart can you review this PR for 0.5.0

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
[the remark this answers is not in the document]
reckart: @Rentier The sorting in SPARQL won't be easy because `getUiLabel` is not trivial. If there is a label, it returns that, otherwise, it returns a suffix of the IRI. Even if we implement this logic in SPARQL, the SPARQL engine won't have a chance of optimizing it using indexes. It may even be slower than sorting locally because the SPARQL engine may first apply the sorting and then apply limiting, meaning it would have to sort all data, not only the returned data. Although I generally agree that sorting should be one in queries and not afterwards, in this particular case, I think it is actually better to sort after the query completed.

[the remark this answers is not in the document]
reckart: @naveen2507 could you please add comment to the sort statement explaining why we do the local sort instead of sorting in the query?

reckart:
@naveen2507 I am trying to test this, but it is hardly possible due to #421. Even in the Wine ontology I cannot create subclasses, add statements under an existing property or add a statement with a new property. How am I supposed to test it?

naveen2507:
@reckart I am working on #421 (for now, we don't delete the property on update and let it just be there in the property list.) . Till then you can test this issue if you have a old project before we got rid of inferencer. Otherwise give me half an hour to commit the code for #421  and then may be you can test this project.

naveen2507:
image

I just checked its working with the old configuration . Added `hasColor` for `IceWine`.

reckart:
When I edit a string literal property (e.g. the "label"), the dropdown field for the value type shows "Choose one" instead of "String".

reckart:
The "KB resource" value editor doesn't seem to be working:

- In the Wine ontology, I created a subclass of the "WineGrape" called "SuperWineGrape"
- then I edit the "label" property and change it to the "KB Resource" type
- 1) auto-completion doesn't seem to work at all. E.g. Typing "Reg" doesn't show a dropdown with the suggestion "Region"
- 2) if I leave "Reg" in the edit field and then click outside it, an Internal Error is generated

The "Boolean value" editor doesn't seem to work either.

- Instead of using "KB Resource", change the value to "Boolean" and try saving the property. That also generates an Internal Error

Neither does the "Numeric editor" seem to work.

- Set the type to "Numeric", enter some value (e.g. "12"), and try saving.
- First type pressing save, nothing happens.
- Second type pressing save, the editor disappears but the value remains unchanged

The "String value" editor also seems incomplete. 

- Choose "String" as the type and enter "en" as language and save
- The editor disappears, but the language doesn't seem to have been set on the property value

Note that some of these actions seem to be creating subclasses of "WineGrape" which show up next to "SuperWineGrape", e.g. I see "12", "true" and "false" there now. But they are actually all the same (i.e. have the same ID). Clicking on any of them as well as clicking on "SuperWineGrape" now displays the class with multiple "label" statements with the respective values "true", "false", "12" and "SuperWineGrape". 

Clicking on "true" or "false" nicely brings up the boolean value editor and the dropdown is set to boolean.

Clicking on "12", however, doesn't seem to bring up the numeric editor and the dropdown says "Choose one".

reckart:
Testing 9cf8f03:

**Editing value is possible without value type being set**
* actual: e.g. when adding a new statement, the value type dropdown shows "choose one", but there is already an auto-complete editor for concepts/instances.
* expected: if INCEpTION does not recognize the value type, I guess there should either be no editor at all (only the value should be displayed in a non-editable fashion) or we should use a simple string editor. Better suggestions?

**Change string feature type to numeric feature type**
* expected: when I change the dropdown from "string" to "numeric", the string editor field should be replaced by a numeric field - any string contents from the field should be cleared and the value should be initialized with 0. When saving the value and pressing "edit" again, the dropdown shows "Numeric" and the number editor comes back.
* actual: whatever value the string field had remains and I can even save it. When I press "edit" again, the dropdown shows "String".

**KB resource suggestions seem off**
* actual: I start entering something, e.g. `winery` and the dropdown comes up and shows options, but it seems to always show **everything** from the ontology, not filtered at all by what I have written.
* expected: when I write "winery", the dropdown should only list concepts that contain that string in their label. We already have code in the `ConceptFeatureEditor` which not only offers a filtered dropdown, but also displays tooltips with the concept descriptions on the bottom of the screen as the user hovers the mouse over an item in the dropdown. Can we re-use this code?
* note: I think this is not critical and can be spun off into a separate issue.

**Editing new KB resource-valued statement does not work as expected**
* expected: when pressing "edit" on KB-resource-valued statement, the editor comes up with the type "kb resource" selected in the dropdown and the value pre-filled into the editor
* actual: when pressing "edit", the type dropdown shows "choose one" and there is no value in the editor.

**Trying to change string value to KB resource value does not work**
* expected: I edit an existing string-valued property, change the type to "KB resource", select a concept, and save. The new KB resource value should be displayed instead of the original string value.
* actual: the original string value is displayed. The KB value doesn't seem to have been saved.

**Trying to have KB resource which does not exist is odd**
* expected: when I try to enter something that does not exist into the "KB resource" editor and try to save, I should get an error message or field validation message that the value is not valid and nothing else should happen.
* actual: I edit an existing string property and then I choose the "KB resource" type and start entering something that does not exist, e.g. "foo". When I press the "save" button, the auto-suggestion field clears. When I press "save" again, the editor goes away and the UI displays the original property. When I press "edit" again, I edit the original string value again.

MdeBoer95:
> Editing new KB resource-valued statement does not work as expected

For most cases this is fixed. Just when the Concept is *Class* the readKBIdentiefier method does return a KBObject and so the AutocompleteField is cleared (see 5283cc5). So what is the best way to check if the value is one of the pre-defined IRIs ? (e.g. *Class*, *Subclass*, *Type* etc.)
I am currently checking this with an if - else if -...  block.

> Trying to change string value to KB resource value does not work

The values are saved but the editor/ presenter is somehow no correctly redrawn or redrawn with the wrong model. You can switch from some String value to Resource, and safe a new Resource as value. Then select another concept in the tree an switch back to the concept you just edited. In this case you can see that the value has actually been saved correctly.
I also observed that if the ValueType is restricted to String and Resource the presenter is correctly rerendered and shows that new value. If the ValueType is not restricted it does not work. (Switch between any ValueTypes will always display the old value as long as the page/panel is not reloaded).

There are also more cases where the editor shows an old value.
Edit a property where the ValueType is restricted to String and Resource (e.g. adjacentRegion in the wine ontology). Select any Resource and save. Now edit again, switch the type to String, enter a string and save again. If you click on edit now, the Resource editor will come up filled with the old Resource value.
Some of these problems already existed before this PR and I am not sure which of these problems are caused by the changes of this PR. Also I could not find out the cause of these so far.

### Another Issue:  editing type statement doesn't work ###
When I try to edit the type statement it throws
*java.lang.NullPointerException at de.tudarmstadt.ukp.inception.ui.kb.value.editor.IRIValueEditor$1.getChoices(IRIValueEditor.java:78) ~[classes/:?]*
I will try to fix it.

naveen2507:
> For most cases this is fixed. Just when the Concept is Class the readKBIdentiefier method does return a KBObject and so the AutocompleteField is cleared (see 5283cc5). So what is the best way to check if the value is one of the pre-defined IRIs ? (e.g. Class, Subclass, Type etc.)
I am currently checking this with an if - else if -... block.

You can make use of `hasImplicitNamespace(String s)` in KBSI.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
```diff
@@ -323,9 +323,10 @@ public void nonempty_WithEmptyKnowledgeBase_ShouldReturnTrue() {
         List<KBHandle> listProperties = sut.listProperties(kb, true);
         Stream<String> listIdentifier = listProperties.stream().map(KBHandle::getIdentifier);
         String[] expectedProps = { kb.getSubclassIri().stringValue(),
-                kb.getLabelIri().stringValue(), kb.getDescriptionIri().stringValue() };
+            kb.getLabelIri().stringValue(), kb.getDescriptionIri().stringValue(),
```
naveen2507: @MdeBoer95 Configure DKPro checkstyle on ur IDE.

MdeBoer95:
@naveen2507 When I add multiple values to the comment property the concept will appear multiple times on the concept tree. This does not happen on master so it might be different from #440. Any idea? The problem exists at least since a7962b2.

naveen2507:
> @naveen2507 When I add multiple values to the comment property the concept will appear multiple times on the concept tree. This does not happen on master so it might be different from #440. Any idea? The problem exists at least since a7962b2.

Yeah, it should happen because we have added description to the query with `DISTINCT` on line https://github.com/inception-project/inception/blob/a7962b22b9ce91f3511efc7407f2c3e1525e8a08/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java#L826

We need to figure out the way to sort this out. I will take a look into this.. @reckart  Can we take this issue along with #440 as it seems to be a similar issue with multiple labels in Wikidata, it appears several times.

reckart:
@naveen2507 @MdeBoer95 I am actually not sure it is a problem that an item appears multiple times in the tree if it has multiple labels. If the user adds multiple labels to an item, she surely has a reason for it. If some labels should not appear in the tree, that I would expect that the users resorts to e.g. `skos:prefLabel` and `skos:altLabel`, configures the mapping such that the `skos:prefLabel` is used for the tree, and puts all but the primary labels into `skos:altLabel`.

naveen2507:
@reckart the SPARQL query attempts to retreive elements with `DISTINCT` label and description and that's why it will return two different KB handles for it. Here is an example : So when we try to retrieve root elements which includes Region with two descriptions . The list gives two elements for Region: 

`
[KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/food#Grape,name=Grape], KBHandle[identifier=http://www.ukp.informatik.tu-darmstadt.de/inception/1.0#node1cllf5hr3x1,name=My Wine], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/food#PotableLiquid,name=PotableLiquid], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/wine#Region,name=Region], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/wine#Region,name=Region], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/wine#VintageYear,name=VintageYear], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/wine#WineDescriptor,name=WineDescriptor], KBHandle[identifier=http://www.w3.org/TR/2003/PR-owl-guide-20031209/wine#Winery,name=Winery]]`

naveen2507:
@MdeBoer95 I will add a method which filters out the elements duplicated

reckart:
@naveen2507 As I said: I'm not sure this is really a problem. I'd say it might be expected behavior. If there are multiple labels, then the item maybe **should** appear multiple times. The user should make sure that there are not multiple labels.

But maybe you can explain #440 bit more, because there it seems to be quite odd. Mind that in #440 it is not that an item shows up under multiple different labels, but rather multiple times under the **same** label.

MdeBoer95:
@reckart @naveen2507 But I think it should not appear multiple times when it has multiple values for the *comment* property. So at least that should be fixed.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/editor/StatementEditor.java
```diff
@@ -310,16 +317,31 @@ public void actionQualifierChanged(AjaxQualifierChangedEvent event)
...
+                rangeValue = property.getObject().getRange();
+            }
+
             valueTypes = valueTypeRegistry.getAllTypes();
             if (rangeValue != null) {
-                Optional<KBObject> rangeKBHandle = kbService.
-                        readKBIdentifier(kbModel.getObject().getProject(), rangeValue);
+                Optional<KBObject> rangeKBHandle = kbService
+                        .readKBIdentifier(kbModel.getObject().getProject(), rangeValue);
                 if (rangeKBHandle.isPresent()) {
```
naveen2507: Why do you need this? It will create issues if the range is String or INTeger . As We already cover that in every liter support. Any specific reason for this?

reckart:
> @reckart @naveen2507 But I think it should not appear multiple times when it has multiple values for the comment property. So at least that should be fixed.

I definitely agree with that. Only the identifier and label should be taken into consideration. I guess I broke the thing regarding the comments when I fixed the tooltips in #481

naveen2507:
> I definitely agree with that. Only the identifier and label should be taken into consideration. I guess I broke the thing regarding the comments when I fixed the tooltips in #481

@reckart  @MdeBoer95 We can add a method to it, but it will be computationally expensive in case of Wikidata or large KBs. Best idea would be to figure out something with SPARQL. But I am not sure .. how exactly we can do that. I would suggest let it be an open issue in #440 (I will add the details for the issue there) and @reckart you can review this PR now.

reckart:
@MdeBoer95 @naveen2507 Nice work!!! 🥇 

Actually, I cannot review the branch because I opened it. But I approve it! I found a few things, but will open separate issues for them. With the final commit here, I also added a few visual improvements.

So if anybody feels like approving this officially, please go ahead. Otherwise I'll use admin power to merge the branch once it is up-to-date and Jenkins gives the thumbs up.

naveen2507:
I guess, I can approve it.

naveen2507:
@reckart Once you merge it, we can also close #422 this issue.

Issue #54, opened 2018-04-16
Append Wikidata description to KBHandle Label


rcffc:
To make the choice of the correct entity easier for the user a description could be displayed like in this picture: 
![screenshot from 2018-04-16 12-50-28](https://user-images.githubusercontent.com/13070150/38805114-f667e7ba-4174-11e8-9222-0d328e919a7c.png)

As far as I know, this cannot be done with the Virtuoso server and the Wikidata Query Api therefore has to be used to get the descriptions.

rcffc:
However there is a bit off a delay for getting the descriptions. It might be faster to fire a query to get the descriptions for all IRIs in one go, but I do not know how to do that.

reckart:
The detail sidebar has to be quite large if you put the descriptions into the dropdown in the was it is done in the screenshot.

It seems to be possible to make the dropdown larger than the original input field to accommodate for more information: https://docs.telerik.com/kendo-ui/api/javascript/ui/autocomplete/configuration/autowidth

rcffc:
A disadvantage of autowidth: Some UiLabels are very long. To fit in all of it into one row will make the sidebar very large. @Rentier  suggested to use the tooltip that is deplayed (e.g. for POS tags) at the bottom in the middle. Where can I find this class?

reckart:
Options like "autoWidth" can be added by overwriting this:

```
	@Override
	public void onConfigure(JQueryBehavior behavior)
	{
                super.onConfigure(behavior); // IMPORTANT
		behavior.setOption("autoWidth", true);
        }
```

reckart:
Check out the "newTemplate" method in de.tudarmstadt.ukp.clarin.webanno.api.annotation.feature.editor.KendoAutoCompleteTextFeatureEditor

Also check out the KendoAutoCompleteTextFeatureEditor.html file which includes a JavaScript that picks up the extra data sent to the UI by the custom template and renders is as a **JQuery tooltip**.

rcffc:
Perfect.

PR #66, opened 2018-04-19
#54 - Append Wikidata description to KBHandle Label

rcffc:
Get description directly while retrieving candidate entities instead of querying wikidata


inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -71,6 +71,7 @@ public static TupleQuery generateCandidateQuery(RepositoryConnection conn, List<
             "      {",
             "        ?e2 ?labelpredicate ?altLabel.",
             "        ?altLabel bif:contains '?entityLabel'. ",
+            "        ?e2 schema:description ?description.",
```
reckart: Ok for now. Please open an issue to make the description IRI configurable in the KB settings.

Issue #63, opened 2018-04-19
Recommender recommondations not reset on document reset


rcffc:
![reset_recommondations](https://user-images.githubusercontent.com/13070150/38981423-b1436b76-43bf-11e8-9acc-9ceb6472771d.gif)

reckart:
@rcffc mind that if you reset a document and there is a second document in the project, then it is still possible that the recommendations on the reset document are generated based on annotations in the other document.

PR #67, opened 2018-04-19
#63 - Recommender recommondations not reset on document reset

rcffc:
It will clear the predictions now, but with the current implementation it also clears active recommenders and trained models for other projects. Rather the activeRecommenders and trainedModels should be saved per Project.

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
@@ -238,9 +238,8 @@ public class RecommendationServiceImpl
     {
         String userName = aEvent.getDocument().getUser();
         Project project = aEvent.getDocument().getProject();
-        clearPredictionsForProject(userName, project);
-        triggerTrainingAndClassification(aEvent.getDocument().getUser(),
-            aEvent.getDocument().getProject());
+        clearState(userName);
+        triggerTrainingAndClassification(userName, project);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -285,7 +296,17 @@ private void clearState(String aUsername)
...
-        
+
+    private void clearPredictionsForProject(String aUsername, Project project)
+    {
+        synchronized (states) {
+            RecommendationUserState state = getState(aUsername);
+            state.clearActivePredictions(project);
+            state.clearIncomingPredictions(project);
+            state.clearTrainedModels();
+            state.clearActiveRecommenders();
```
reckart: Do we need such fine-grained control over the clearing? Would a `state.clear(project)` not suffice?

reckart: The RecommendationUserState expects that the user is only actively working on one project at a time. IMHO it would be sufficient to clear/rebuild the entire recommender state when a document is reset. But I thing having one method call instead of four would be sufficient - or do we need fine-grained control over which parts of the state are cleared? I included the "project" parameter because you had it in `state.clearActivePredictions(project);`.

reckart: I would like to avoid keeping recommender states for multiple projects in memory because they would easily fill up memory.

rcffc: Ok, addressed in latest commit.

reckart:

reckart:

PR #69, opened 2018-04-19
Bug/64 remove predictions when recommender removed

rcffc:


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -358,5 +372,40 @@ public void putTrainedModel(Recommender aRecommender, Object aModel)
         {
             trainedModels.put(aRecommender.getId(), aModel);
         }
+
+        public void removePredictions(Recommender aRecommender, Project aProject)
```
reckart: The recommender has a reference to the project. No need to pass the project separately.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -358,5 +372,40 @@ public void putTrainedModel(Recommender aRecommender, Object aModel)
...
+                = activeRecommenders.mapIterator();
+
+            while (it.hasNext()) {
+                AnnotationLayer layer = it.next();
+                Recommender rec = it.getValue();
+                if (!rec.equals(aRecommender)) {
+                    newActiveRecommenders.put(layer, rec);
+                }
+            }
+            setActiveRecommenders(newActiveRecommenders);
```
reckart: Normally, we only set the incoming model and let the rendering code turn the incoming model into the active model. If it is really valid to set the active model directly here, there should be a comment explaining why this is necessary.

reckart:

reckart:
@rcffc dependency problems - please check the jenkins logs

reckart:

reckart:

reckart:

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -373,26 +373,28 @@ public void putTrainedModel(Recommender aRecommender, Object aModel)
...
+                activePredictions.put(aRecommender.getProject().getId(), active);
             }
 
             // Remove trainedModel
             trainedModels.remove(aRecommender.getId());
 
-            // Remove from activeRecommenders map
+            // Remove from activeRecommenders map.
+            // We have to do this, otherwise training and prediction continues for the
+            // recommender when a new task is triggered.
```
reckart: Why would we not want the training and prediction to continue? 

If a document is reset, we need to clear the model and the current predictions, but then a new training/prediction run needs to be triggered. If we remove the recommender form the active map, this doesn't happen, does it?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
reckart: Never mind - wrong issue.

Issue #71, opened 2018-04-20
Entity feature editor does not ignore disabled KBs


reckart:
It seems as if the autocomplete field in the concept feature editor does not skip disabled KBs. So if I have a big remote **disabled** KB in my project and a small local **enabled** one, it still takes ages to show the auto-complete values. If I delete the remote KB, things are fast.

PR #74, opened 2018-04-20
#71 - Entity feature editor does not ignore disabled KBs

rcffc:


reckart:

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
```diff
@@ -191,6 +191,10 @@ private JCas getEditorCas(AnnotationActionHandler aHandler) throws IOException
             else {
                 // If no specific KB is selected, collect instances from all KBs
                 for (KnowledgeBase kb : kbService.getKnowledgeBases(project)) {
```
reckart: Better to use getEnabledKnowledgeBases().

Issue #75, opened 2018-04-20
Entity Linking does not return as much candidates as it could


rcffc:
The problem is calling limit() in a stream before distinct().

PR #76, opened 2018-04-20
#75 - Entity Linking does not return as much candidates as it could

rcffc:


reckart:

Issue #91, opened 2018-04-22
Make ConceptLinkingService fields configurable


rcffc:
Commit https://github.com/inception-project/inception/commit/751f99a7126baf82632f9588777eda362d5c9753 in #93 introduces a way to let an admin configure the cache size of ConceptLinkingService. One could specify these values for the other fields of the class in a similar fashion.

reckart:
@rcffc is this complete?

PR #98, opened 2018-04-23
#91 - Make ConceptLinkingService fields configurable

rcffc:


inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -122,6 +122,12 @@ public void init()
         candidateCache = Collections.synchronizedMap(new LRUCache<>(properties.getCacheSize()));
         semanticSignatureCache = Collections
             .synchronizedMap(new LRUCache<>(properties.getCacheSize()));
+
+        MENTION_CONTEXT_SIZE = properties.getMentionContextSize();
```
reckart: Capitalized variable names are reserved for constants. Better fetch the values always directly from the properties instead of storing them in variables.

Issue #101, opened 2018-04-24
Initialised Layer for Fact 


matchlesswei:
create a initialised layer for fact in project setting

PR #104, opened 2018-04-24
#101 initialised fact layer

matchlesswei:
- create a fact layer initializer
- update fact linking constants

### DIFF
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
@@ -19,12 +19,14 @@ package de.tudarmstadt.ukp.inception.ui.kb.feature;
 
 public final class FactLinkingConstants
 {
-    public static final String SUBJECT_LINK = "webanno.custom.FactSubjectLink";
-    public static final String OBJECT_LINK = "webanno.custom.FactObjectLink";
-    public static final String QUALIFIER_LINK = "webanno.custom.FactQualifierLink";
+    public static final String SUBJECT_LINK = "de.tudarmstadt.ukp.inception.FactSubjectLink";
+    public static final String OBJECT_LINK = "de.tudarmstadt.ukp.inceptionFactObjectLink";
+    public static final String QUALIFIER_LINK = "de.tudarmstadt.ukp.inception.FactQualifierLink";
 
     public static final String SUBJECT_ROLE = "subject";
     public static final String OBJECT_ROLE = "object";
 
-    public static final String FACT_LAYER = "webanno.custom.Fact";
+    public static final String FACT_LAYER = "de.tudarmstadt.ukp.inception.Fact";
+
+    public static final String LINKED_LAYER_FEATURE = "identifier";
 }
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
@@ -19,14 +19,15 @@ package de.tudarmstadt.ukp.inception.ui.kb.feature;
 
 public final class FactLinkingConstants
 {
-    public static final String SUBJECT_LINK = "de.tudarmstadt.ukp.inception.FactSubjectLink";
-    public static final String OBJECT_LINK = "de.tudarmstadt.ukp.inceptionFactObjectLink";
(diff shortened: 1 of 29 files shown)


inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
```diff
@@ -19,12 +19,14 @@
 
 public final class FactLinkingConstants
 {
-    public static final String SUBJECT_LINK = "webanno.custom.FactSubjectLink";
-    public static final String OBJECT_LINK = "webanno.custom.FactObjectLink";
-    public static final String QUALIFIER_LINK = "webanno.custom.FactQualifierLink";
+    public static final String SUBJECT_LINK = "de.tudarmstadt.ukp.inception.FactSubjectLink";
+    public static final String OBJECT_LINK = "de.tudarmstadt.ukp.inceptionFactObjectLink";
```
reckart: `de.tudarmstadt.ukp.inception.api.kb.type.FactObjectLink`
`de.tudarmstadt.ukp.inception.api.kb.type.FactSubjectLink`
`de.tudarmstadt.ukp.inception.api.kb.type.FactQualifierLink`

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
```diff
@@ -19,12 +19,14 @@
...
+    public static final String QUALIFIER_LINK = "de.tudarmstadt.ukp.inception.FactQualifierLink";
 
     public static final String SUBJECT_ROLE = "subject";
     public static final String OBJECT_ROLE = "object";
 
-    public static final String FACT_LAYER = "webanno.custom.Fact";
+    public static final String FACT_LAYER = "de.tudarmstadt.ukp.inception.Fact";
```
reckart: `de.tudarmstadt.ukp.inception.api.kb.type.Fact`

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/FactLinkingConstants.java
```diff
@@ -19,12 +19,14 @@
...
+
+    public static final String LINKED_LAYER_FEATURE = "identifier";
```
reckart: Please add a comment that `identifier` here is a feature of the `NamedEntity` layer.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/initializers/FactLayerInitializer.java
```diff
@@ -0,0 +1,111 @@
...
+    {
+        AnnotationLayer factLayer = new AnnotationLayer(FactLinkingConstants.FACT_LAYER, "Fact",
+            SPAN_TYPE, aProject, false);
+        factLayer.setAllowStacking(true);
+        factLayer.setCrossSentence(false);
+        factLayer.setLockToTokenOffset(false);
+        factLayer.setMultipleTokens(true);
+
+        annotationSchemaService.createFeature(
+            new AnnotationFeature(aProject, factLayer, "KBPredicate", "a-Predicate",
```
reckart: Feature names should start with a small letter.

Feature names: `predicate`, `subject`, `object', `qualifiers`.

If you really need to prefix the UI names, I would suggest to prefix them with numbers, e.g. `1) Predicate`.

matchlesswei: I cannot use numbers as prefix because "Feature names must start with a letter and consist only of letters, digits, or underscores".

reckart: That limitation exists if you create a feature in the UI and it only exists when you initially created the feature because at that point, the "internal name" is derived from the feature name. After the feature has been created, you can rename it arbitrarily, because then it only affects the "uiName". When you use the API directly, you set the "name" and "uiName" independently anyway, so you are not bound by this limitation. The "name" should start with a lower-case letter and must conform to the Java naming rules for class fields. The "uiName" can essentially be anything.

matchlesswei: Thanks, I have changed them. BTW, if I rename the name manually in projects setting, I still get this message and cannot rename it arbitrarily.

matchlesswei:
No particular reason, just not sure whether it can be merged before Friday. I have changed it to 0.3.0.

Issue #103, opened 2018-04-24
"FileUtils - Could not read file properties_with_labels.txt" when starting INCEpTION


Rentier:
When I start INCEpTION (current master), then I get the following error:

```
2018-04-24 17:04:45 ERROR [SYSTEM] FileUtils - Could not read file properties_with_labels.txt
java.io.FileNotFoundException: /home/klie/.inception/repository/resources/properties_with_labels.txt (No such file or directory)
	at java.io.FileInputStream.open0(Native Method) ~[?:1.8.0_162]
	at java.io.FileInputStream.open(FileInputStream.java:195) ~[?:1.8.0_162]
	at java.io.FileInputStream.<init>(FileInputStream.java:138) ~[?:1.8.0_162]
	at de.tudarmstadt.ukp.inception.conceptlinking.util.FileUtils.readLines(FileUtils.java:52) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.conceptlinking.util.FileUtils.loadPropertyLabels(FileUtils.java:73) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.conceptlinking.service.ConceptLinkingService.init(ConceptLinkingService.java:114) ~[classes/:?]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:1.8.0_162]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:1.8.0_162]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_162]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_162]
	at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleElement.invoke(InitDestroyAnnotationBeanPostProcessor.java:369) ~[spring-beans-5.0.4.RELEASE.jar:5.0.4.RELEASE]
	at org.springframework.beans.factory.annotation.InitDestroyAnnotationBeanPostProcessor$LifecycleMetadata.invokeInitMethods(InitDestroyAnnotationBeanPostProcessor.java:312) ~[spring-beans-5.0.4.RELEASE.jar:5.0.4.RELEASE]
[...]

PR #107, opened 2018-04-25
#103 - "FileUtils - Could not read file properties_with_labels.txt" w…

rcffc:
…hen starting INCEpTION

Just log that a resource file for EntityLinking has not been found instead of printing the stack trace

### DIFF
--- a/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/FileUtils.java
+++ b/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/FileUtils.java
@@ -56,7 +56,7 @@ public class FileUtils
             }
             br.close();
         } catch (IOException e) {
-            log.error("Could not read file " + r.getName(), e);
+            log.error("Resource file [{}] not present.", r.getName());
         }
         return lines;
     }


Rentier:
Why dont you add the file?

Rentier:
Then I would add in the error message how to obtain it.

reckart:
@Rentier the current implementation for entity linking resources is essentially a hack required because we need to do some testing. For a proper implementation, we would need to allow the user to upload the resources through the UI.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/FileUtils.java
```diff
@@ -56,7 +56,7 @@
             }
             br.close();
         } catch (IOException e) {
-            log.error("Could not read file " + r.getName(), e);
+            log.error("Resource file [{}] not present.", r.getName());
```
reckart: Since the code works even without the resources, I would make it a warning. Please also mention which functionality is not available when the resources is missing. E.g.

```
Resource file [{}] is missing - {}, r.getName(), reason
```

to print something lile

```
WARN - Resource file [/some/path/stopwords.txt] is missing - using entity linking support without stopwords may have a negative impact on the suggestion ranking.
```

rcffc: Addressed in new commit.

Issue #59, opened 2018-04-17
Indexing blocks UI when changing layer settings


betoboullosa:
The indexing process currently runs immediately and synchronously when the layer configuration changes. This is seriously annoying because the UI blocks until the process is complete which can take quite a while even for a small set of documents.

When the layer configuration is changed, the index should be invalidated and a background job should be scheduled for re-indexing. We currently have a job queue in the recommendation subsystem. Optimally, this queue should be generalized into a separate module and be used also by the indexer.

betoboullosa:
Done.

PR #109, opened 2018-04-25
#59 - Feature/search scheduling

betoboullosa:
Addition of search scheduler


betoboullosa:

reckart:
@betoboullosa could you please have a look at the conflicts?

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/index/PhysicalIndex.java
```diff
@@ -28,21 +28,21 @@
 import de.tudarmstadt.ukp.inception.search.ExecutionException;
 import de.tudarmstadt.ukp.inception.search.SearchResult;
 
-public interface Index
+public interface PhysicalIndex
```
Rentier: What is a physical index?

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/IndexScheduler.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private @Autowired ApplicationContext applicationContext;
+
+    private Thread consumer;
+    private BlockingQueue<Task> queue = new ArrayBlockingQueue<Task>(100);
+
+    @PostConstruct
+    private void startSchedulerThread()
+    {
+        consumer = new Thread(new TaskConsumer(applicationContext, queue), "Index task consumer");
```
Rentier: Is there a reason why you not use Java's `ExecutorService`? It looks like it also allows task scheduling, in addition to that, it has a thread pool and you can get access to the queue behind it.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/TaskConsumer.java
```diff
@@ -0,0 +1,77 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
+import org.springframework.context.ApplicationContext;
+
+import de.tudarmstadt.ukp.inception.search.scheduling.tasks.Task;
+
+public class TaskConsumer
+    implements Runnable
+{
+    private Logger log = LoggerFactory.getLogger(getClass());
```
Rentier: I would make logs static if there is no good reason not to.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/TaskConsumer.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private ApplicationContext applicationContext;
```
Rentier: You could do these final, can't you?

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/index/PhysicalIndex.java
betoboullosa: Indexes have two different levels of abstraction, the logical and the physical. This is the physical level of abstraction.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/IndexScheduler.java
```diff
@@ -0,0 +1,102 @@
...
+ */
+@Component
+public class IndexScheduler
+{
+    private Logger log = LoggerFactory.getLogger(getClass());
+
+    private @Autowired ApplicationContext applicationContext;
+
+    private Thread consumer;
+    private BlockingQueue<Task> queue = new ArrayBlockingQueue<Task>(100);
```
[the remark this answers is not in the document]
betoboullosa: The index scheduler is currently based on the recommendation scheduler, so I don't want to make it very different from its counterpart. Optimizations should be done when both of them are integrated in a general scheduler.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/IndexScheduler.java
betoboullosa: Cf. the comment on the recommender scheduler.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/IndexScheduler.java
```diff
@@ -0,0 +1,102 @@
...
+        // If there is no indexing in the queue on for this project, enqueue it
+        if (!isIndexing(aRunnable.getProject())) {
+            queue.offer(aRunnable);
+            log.info("Enqueued new indexing task: {}", aRunnable);
+        }
+    }
+
+    public void stopAllTasksForUser(String username)
+    {
+        Iterator<Task> it = queue.iterator();
```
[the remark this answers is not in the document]
betoboullosa: Cf. the comment on the recommender scheduler.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/TaskConsumer.java
betoboullosa: OK. But it should be seen in other points of INCEpTION code, for most of them declare the log variable as non-static.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/TaskConsumer.java
betoboullosa: OK.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/scheduling/TaskConsumer.java
reckart: Currently, log variables are only static when required in static methods or when used in Wicket components. Otherwise we keep them non-static. This SO post includes a few sensible answers explaining why one might always want to use static (https://stackoverflow.com/questions/6653520/why-do-we-declare-loggers-static-final). I usually prefer using non-static to avoid copy-paste problems. `log = LogFactory.getLog(getClass())` works when copy/pasted, but `log = LogFactory.getLog(MyClass.class)` is usually wrong when copy/pasted.

Issue #110, opened 2018-04-25
Set up an LRU Cache for renderValue


rcffc:
renderFeatureValue() in ConceptFeatureSupport is called very often during rendering, making the whole AnnotationPage very slow. It might be worth to set up an LRU cache instead of relying on the performance of the underlying KB store.

rcffc:
Rendering time dropped from 3.5s to <30ms.

PR #113, opened 2018-04-25
#110 - Set up an LRU Cache for renderValue

rcffc:


inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureSupport.java
```diff
@@ -120,6 +125,10 @@ public boolean accepts(AnnotationFeature aFeature)
     @Override
     public String renderFeatureValue(AnnotationFeature aFeature, String aLabel)
     {
+        FeatureValuePair pair = new FeatureValuePair(aFeature, aLabel);
+        if (renderValueCache.containsKey(pair)) {
```
reckart: It is even faster to call "get" and then to check if the returned value is null. That way only one lookup in the cache is necessary instead of two (containsKey/get).

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureSupport.java
```diff
@@ -280,4 +289,33 @@ public void generateFeature(TypeSystemDescription aTSD, TypeDescription aTD,
     {
         aTD.addFeature(aFeature.getName(), "", CAS.TYPE_NAME_STRING);
     }
+
+    private class FeatureValuePair {
```
reckart: How about using org.apache.commons.lang3.tuple.ImmutablePair<AnnotationFeature, Label>?

PR #116, opened 2018-04-26
Feature/73 external recommender

Rentier:
TODO:

- I added documentation, but the document module does not find it:

`asciidoctor: WARNING: user-guide.adoc: line 114: include file not found: /home/klie/git/inception-github/inception-app-webapp/target/inception-app-webapp-0.3.0-SNAPSHOT-documentation/META-INF/asciidoc/user-guide/external-recommender.adoc`

- Make the parameters (Layer name, feature name, URL configurable)
- Maybe use different HTTP client
- Mabye not use the JSON from Wicket, it is a strange dependency to have on wicket-core


reckart:
@Rentier the documentation in the external recommender module is pointless. No actual user can make any sense of it. I would suggest to either remove it or replace it with a description of the communication protocol which external classifiers need to implement in order to be used with the module.

Rentier:
I agree that it does not make sense for users. I have written it for us and ClS' Hiwi; I thought that is the best place for it.

reckart:
@Rentier I'd merge this branch after https://github.com/inception-project/inception/pull/121 because the latter contains a restructuring of the recommender modules and is fairly large.

Issue #168, opened 2018-05-08
Specifiable URI prefix in Knowledge Base


MdeBoer95:
Currently, all statements created in a local KB are created with prefix http://www.ukp.informatik.tu-darmstadt.de/inception/1.0#. Users importing, extending and exporting an existing KB will not be pleased.
- [ ] add base prefix option to knowledge base creation wizard (only visible for the "custom"/"expert" option, see #132.
- [ ] Add base prefix option to the local knowledge base edit page.
- [ ] Add JUnit tests for the additional knowledge base property.

PR #169, opened 2018-05-08
#168 - Specifiable URI prefix in Knowledge Base

MdeBoer95:
- added basePrefix property for KnowledgeBase
- added TextField for basePrefix in Wizard
- added Textfield for basePrefix in KB edit mode
- changed generateIdentifier-method in KnowledgeBaseService so that it uses the basePrefix now


reckart:
@MdeBoer95 can you please check the conflicts and resolve them?

reckart:
@MdeBoer95 could you please have a look at the conflicts.

MdeBoer95:
see comment on #184

naveen2507:
@MdeBoer95 Internal Error on KB Edit Mode

org.apache.wicket.WicketRuntimeException: Property 'kb.basePrefix' not found in property files. Markup: [markup = jar:file:/home/naveen/Inception/Inception-workspace/.metadata/.plugins/org.eclipse.wst.server.core/tmp0/wtpwebapps/inception-app-webapp/WEB-INF/lib/inception-ui-kb-0.4.0-SNAPSHOT.jar!/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseDetailsPanel.html
<wicket:message key="kb.basePrefix"/>, index = 1, current =  '<wicket:message key="kb.basePrefix"/>' (line 0, column 0)]

reckart:
The strings for components on the KB page seem all to be aggregated in `ProjectKnowledgeBasePanel.properties`. So I guess the new string should be added there and be specified in the `wicket:message` with the full prefix, e.g. `kb.iri.basePrefix`.

MdeBoer95:
@reckart I have tried to do this before but if I just call new Label("basePrefixLabel") (using same markup to display the label as for the other fields) without the optional second parameter I get:

org.apache.wicket.WicketRuntimeException: Property could not be resolved for class: class de.tudarmstadt.ukp.inception.ui.kb.project.KnowledgeBaseWrapper expression: basePrefixLabel

A new .properties file and using get("basePrefixLabel") was the easiest solution for that problem.

Do you know how I can solve it in another way?

MdeBoer95:
@naveen2507 error should not happen anymore. I also made some changes to when the field is visible.

reckart:
@MdeBoer95 Labels should only be used for non-static text. I.e. text that you need to modify programmatically. If you want just to add a static string to the UI, use e.g. `<wicket:message>` (https://ci.apache.org/projects/wicket/guide/7.x/single.html#_reading_bundles_with_tag_wicket_message).

reckart:
@MdeBoer95 if you want to group a static string with some input element so that the static string is not visible when the input element is hidden, you can use `wicket:enclosure`(https://ci.apache.org/projects/wicket/guide/7.x/single.html#_hiding_decorating_elements_with_the_wicket_enclosure_tag).

MdeBoer95:
@reckart Thanks!

naveen2507:
image

@MdeBoer95  Can you please see to it, as the Textfield for base prefix is not visible for me . 
In Page Source I see an empty div element.

MdeBoer95:
@naveen2507 it is only visible in custom mode as kind of "advanced setting" (see description #168 )or if it is not the default value anymore so that the user is aware of the changed prefix.

reckart:
@MdeBoer95 which prefix is used in the "pre-configured" modes?

MdeBoer95:
@reckart the mode does not affect the value of the prefix. So it is  **http://www.ukp.informatik.tu-darmstadt.de/inception/1.0#** for all, unless the user changes it.

reckart:
@MdeBoer95 then I think the prefix should always be offered for configuration, not only in "custom" mode.

MdeBoer95:
@reckart I think the idea behind this was to not have too many configurations, so that an unexperienced user does not get confused/overwhelmed.
If the user actually cares about the prefix (I don't know how frequent that might happen) he can change it in custom mode and then still select one of the pre-configured modes to set the IRIs.
Alternatively, adding a **"show advanced settings"** button for all modes would also be a possibility.

naveen2507:
@MdeBoer95 we need to have a general text field above the mode setup for basePrefix. Let's discuss the UI in tomorrow's meeting in case you are not clear about that.

naveen2507:
image

reckart:
What's happening in this screenshot? Seems there is a "scrolling" CSS class missing somewhere. What browser?

naveen2507:
@reckart Chrome

naveen2507:
@MdeBoer95 Can you look at the conflicts here

Issue #149, opened 2018-05-02
Feature editor directly in AL sidebar


matchlesswei:
Feature editor directly in AL sidebar

matchlesswei:
I want to create a FeatureEditor in the ALsidebar, I use 
```
FeatureEditor editor = featureSupport.createEditor("editor", mainContainer, this
            .getActionHandler(), this.getModel(), aFeatureStateModel);
```
So I need a FeatureStateModel `IModel<FeatureState> aFeatureStateModel = new FeatureStateModel(annotatorStateModel, annotationFeature);` 
The problem is:
This annotationFeature should be from the ALsidebar Layerselection(for example: POS layer), but the annotationFeature in annotatorStateModel is from the **right** sidebar Layerselection(for example: SurfaceForm)

I want to get POS feature editor, but the method `createEditor` in FeatureSupport cannot get POS feature, because it gets feature from AnnotatorState(feature in right sidebar). @reckart how can I solve this problem?

reckart:
Can you use this?

```
IModel<FeatureState> aFeatureStateModel = Model.of(new FeatureState(annotationFeature, recommendedValue));
```

matchlesswei:
Yes, thanks!

matchlesswei:
<img width="1280" alt="screen shot 2018-05-08 at 15 52 43" src="https://user-images.githubusercontent.com/23156705/39761344-23b3a3e8-52d8-11e8-8e1e-21e45b7fc6df.png">

@reckart @betoboullosa ,2 more questions:
1. Do we need to add a 'correct' button in the al sidebar? If the user sets a value in the feature editor and press 'correct', then this annotation is saved, right?
2. If the user choose a token in the middle and select a same layer in right sidebar, the feature editor in the right sidebar should also be a `KendoComboboxTextFeatureEditor` according to the debug result of `PrimitiveUimaFeatureSupport`, but it is not. Do you have any idea why?

reckart:
1) I think adding a "correct" button is necessary, yes. In the right sidebar, the value entered by the user is saved immediately when the focus leaves the input field. I feel that this would not be the proper approach in the AL sidebar though. Pressing "correct" should create an annotation with the selected value, save it to the AL history, and then directly move on to the next recommendation.

2) For string features, the `PrimitiveUimaFeatureSupport` has three modes: 
* no tagset: simple text field
* tagset with <75 tags: combobox
* tagset with >=75 tags: auto-complete field

[...]

PR #173, opened 2018-05-10
#149 - Feature editor directly in AL sidebar

reckart:
- Factor out code to set markup-id. This is handled by directly by the code of the right sidebar now.


reckart:
Already merged via #176.

Issue #176, opened 2018-05-12
Move ID-setting code from feature editors into FeatureEditorPanelContent


rcffc:
Adjust to changes in Webanno: 
https://github.com/webanno/webanno/issues/934

PR #177, opened 2018-05-12
#176 - Move ID-setting code from feature editors into FeatureEditorPanelContent

rcffc:
- Adjust to changes in Webanno


matchlesswei:
It's actually a duplicate of PR #173 , but it's really critical.

PR #179, opened 2018-05-12
Bugfix/156 recommendations disappear

### LINKED ISSUE #156: Recommondations disappear after resizing/reloading
When resizing, recommondations disappear. They are actually better than the average recommondations. This happens after deleting an annotation.

The issue is maybe one of the following two:
1) The annotation is deleted but should not be used for recommondation, but actually is 
OR
2) The recommondation disappears and should stay

recommondations_disappear

rcffc:
Fixes #156 .


Rentier:
@rcffc It looks like there is a lot of stuff unrelated to the bugfix, isn't it?

Rentier:
@rcffc Can you please state what the bug actually was?

rcffc:
Added link to description.

Rentier:
I know the related issue, but you never state what the underlying cause is, .e.g ` && sentenceIndex < sentence.size() - 2` is magic.

rcffc:
The first bug is that during the merging neighboring tokens an index outside the bounds of the sentenceList is accessed, resulting in an IndexOutOfBoundsError.
The second bug is that during rendering the predictions for each token, the system just examines the first prediction and tests whether there is already existing annotation for the given feature and ignores all predictions for this token if true, even predictions for other features.

Issue #185, opened 2018-05-14
Sort Knowledge Bases in Project Settings


MdeBoer95:
The "knowledge bases" list in the project settings should always appear sorted from A to Z.

PR #186, opened 2018-05-14
#185 - Sort Knowledge Bases in Project Settings

MdeBoer95:
- kbs are sorted from A-Z now, ignoring the case

### DIFF
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseListPanel.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseListPanel.java
@@ -17,6 +17,8 @@
  */
 package de.tudarmstadt.ukp.inception.ui.kb.project;
 
+import java.util.stream.Collectors;
+
 import org.apache.wicket.ajax.AjaxRequestTarget;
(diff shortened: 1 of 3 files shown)


naveen2507:
@reckart Do we need to count in the case? If not, then, for now, this can be approved and closed.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseListPanel.java
```diff
@@ -57,7 +59,9 @@ public KnowledgeBaseListPanel(String id, IModel<Project> aProjectModel,
         overviewList = new OverviewListChoice<>("knowledgebases");
         overviewList.setChoiceRenderer(new ChoiceRenderer<>("name"));
         overviewList.setChoices(
-                LambdaModel.of(() -> kbService.getKnowledgeBases(projectModel.getObject())));
+                LambdaModel.of(() -> kbService.getKnowledgeBases(projectModel.getObject()).stream()
```
Rentier: Is there a reason why you chose `stream` over `Collections::sort`?

MdeBoer95: less code and looks nicer

PR #189, opened 2018-05-14
Bugfix/161 liquibase changes

betoboullosa:


reckart:
Just for the record: it seems that liquibase includes the filename as part of the changelog identity. So when we move changelogs from one file to another, liquibase seems to try to re-execute them. So it is really always a good idea to have preconditions.

betoboullosa:
@reckart , I've corrected the author information. You can merge it.

PR #190, opened 2018-05-14
feature/6 refactor annotation object

Rentier:
I removed the inheritance from `AnnotationObject` and removed some fields. Next step will be the refactoring of the recommender API.


inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/AnnotationObject.java
```diff
@@ -18,239 +18,202 @@
...
 
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("AnnotationObject{");
+        sb.append("token=").append(token);
+        sb.append(", id=").append(id);
+        sb.append(", label='").append(label).append('\'');
+        sb.append(", uiLabel='").append(uiLabel).append('\'');
+        sb.append(", feature='").append(feature).append('\'');
+        sb.append(", source='").append(source).append('\'');
```
reckart: source and recommenderId are kind of redundant - but we can address this later.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/util/CasUtil.java
```diff
@@ -193,20 +194,23 @@ private CasUtil()
             TokenObject tObj = sentence.get(i);
 
             if (indexAnnotations >= annotations.size()) {
-                result.add(new AnnotationObject(null, tObj, sentence, id, feature));
+                result.add(new AnnotationObject(tObj, id, feature, SOURCE));
```
reckart: Why do we need to set Source to "CasUtil" here? Wouldn't it make more sense to pass in the classifier name as a parameter... or just leaving the source empty?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/util/CasUtil.java
```diff
@@ -238,7 +242,7 @@ private CasUtil()
                 TokenObject tObj = new TokenObject(offset, token.getCoveredText(),
                     documentURI, documentName, id);
                 result.add(
-                    new AnnotationObject(annotationLabel, tObj, sentence, id, feature.getName()));
+                    new AnnotationObject(tObj, annotationLabel, "", id, feature.getName(), SOURCE));
```
reckart: Passing an empty string looks a bit "magic". Happens more times below.

pom.xml
```diff
@@ -92,7 +92,7 @@
     <!-- module>inception-imls-mira</module -->
     <module>inception-imls-opennlp</module>
     <module>inception-imls-stringmatch</module>
-    <module>inception-imls-dl4j</module>
+    <!-- module>inception-imls-dl4j</module -->
```
reckart: Why? If you remove it here, you also need to remove it as a dependency from the webapp module.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/AnnotationObject.java
Rentier: Isn't the source used for the UI?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/util/CasUtil.java
reckart: Then how about leaving the source null?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/AnnotationObject.java
reckart: There are some AnnotationObject constructors which set the source (classifier) to null. A good measure might be to leave it at null and to add throw an exception in `getSource()` if the source has not been set. Then we get an idea where this might be needed but is not provided properly.

pom.xml
reckart: Ok :) will reduce our build times by a few minutes as well.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -114,14 +114,14 @@ private static void removeRecommendationsWithNullAnnotation(
     private static boolean isAlreadyInCleanList(List<AnnotationObject> cleanRecommendationList,
             AnnotationObject recommendationItem)
     {
-        String classifier = recommendationItem.getClassifier();
-        String annotation = recommendationItem.getAnnotation();
+        String classifier = recommendationItem.getSource();
```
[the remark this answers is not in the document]
Rentier: Done

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/util/CasUtil.java
Rentier: Done

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/util/CasUtil.java
Rentier: Done

pom.xml
Rentier: Done

Rentier:
@reckart Can you again have a look at this?

Issue #182, opened 2018-05-12
Recommendations in different documents have same VID


matchlesswei:
Using the method `public AnnotationObject getPredictionByVID(VID aVID) `, I got AnnotationObject from not desired document. By debugging, there are AnnotationObjects from different documents but having same VID.
<img width="1276" alt="screen shot 2018-05-13 at 00 34 23" src="https://user-images.githubusercontent.com/23156705/39962157-2ce33a88-5647-11e8-842f-6ef4f0494605.png">
This leads to that the learning record is different from user click in BratAnnotationEditor.

Rentier:
@matchlesswei Where do you use the function?

matchlesswei:
If you click to reject or accept the recommendations in BratEditor, then you may notice the learning record in ALsidebar history may be wrong.

PR #193, opened 2018-05-14
#182  - Recommendations in different documents have same VID

Rentier:
I just filter by document name to restrict results to a document.


inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -634,9 +635,10 @@ public void onRecommendationRejectEvent(AjaxRecommendationRejectedEvent aEvent)
         
         if (sessionActive && eventState.getUser().equals(annotatorState.getUser())
                 && eventState.getProject().equals(annotatorState.getProject())) {
-            if (eventState.getDocument().equals(annotatorState.getDocument())
+            SourceDocument document = eventState.getDocument();
+            if (document.equals(annotatorState.getDocument())
                     && aEvent.getVid().getLayerId() == selectedLayer.getObject().getId()
-                    && predictionModel.getPredictionByVID(aEvent.getVid())
```
reckart: The `getPredictionByVID` method seems to be fundamentally flawed, so IMHO it should be removed and replaced by the `getPredictionInDocumentByVID` method. It seems you only add the `getPredictionInDocumentByVID` and the old one still sticks around and could cause further trouble.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Predictions.java
```diff
@@ -207,6 +208,19 @@ public AnnotationObject getPredictionByVID(VID aVID)
...
+     * Returns the first prediction that matches recommendationId and recommenderId
+     * in the given document
+     */
+    public AnnotationObject getPredictionInDocumentByVID(SourceDocument document, VID aVID)
+    {
+        return predictions.values().stream()
+                .filter(f -> f.getDocumentName().equals(document.getName()))
+                .filter(f -> f.getId() == aVID.getSubId())
+                .filter(f -> f.getRecommenderId() == aVID.getId())
+                .collect(Collectors.toList()).get(0);
```
reckart: We do not need to collect the full list then to only return the first element, right? That seems inefficient. Also, we get a nasty exception in case no annotation with the given document/VID is found. Either return null or maybe an optional.

matchlesswei:
It looks fine!

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
```diff
@@ -102,11 +104,18 @@ private void actionAcceptRecommendation(AnnotationActionHandler aActionHandler,
...
     {
-        Predictions model = 
+
+        SourceDocument document = aState.getDocument();
+        Predictions model =
                 recommendationService.getPredictions(aState.getUser(), aState.getProject());
-        AnnotationObject prediction = model.getPredictionByVID(aVID);
+        Optional<AnnotationObject> prediction = model.getPredictionByVID(document, aVID);
+        if (!prediction.isPresent()) {
+            log.error("Could not find annotation in [{}] with id [{}]", document, aVID);
```
reckart: Such a problem should also be logged to the user. We are usually in an interactive cycle when hitting this method.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
```diff
@@ -168,14 +177,21 @@ private void actionRejectRecommendation(AnnotationActionHandler aActionHandler,
         Recommender recommender = recommendationService.getRecommender(aVID.getId());
         AnnotationLayer layer = annotationService.getLayer(aVID.getLayerId());
         AnnotationFeature feature = annotationService.getFeature(recommender.getFeature(), layer);
-        
-        AnnotationObject prediction = model.getPredictionByVID(aVID);
+
+        SourceDocument document = aState.getDocument();
+        Optional<AnnotationObject> oPrediction = model.getPredictionByVID(document, aVID);
+        if (!oPrediction.isPresent()) {
+            log.error("Could not find annotation in [{}] with id [{}]", document, aVID);
```
reckart: Such a problem should also be logged to the user. We are usually in an interactive cycle when hitting this method.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -649,12 +653,19 @@ public void onRecommendationRejectEvent(AjaxRecommendationRejectedEvent aEvent)
...
             annotatorState.getProject());
         AnnotatorState eventState = aEvent.getAnnotatorState();
-        AnnotationObject acceptedRecommendation = predictionModel
-                .getPredictionByVID(aEvent.getVid());
+        SourceDocument document = annotatorState.getDocument();
+        VID vid = aEvent.getVid();
+        Optional<AnnotationObject> oRecommendation = model.getPredictionByVID(document, vid);
 
+        if (!oRecommendation.isPresent()) {
+            LOG.error("Could not find annotation in [{}] with id [{}]", document, vid);
```
reckart: Such a problem should also be logged to the user. We are usually in an interactive cycle when hitting this method.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -661,7 +667,8 @@ public void onRecommendationAcceptEvent(AjaxRecommendationAcceptedEvent aEvent)
         Optional<AnnotationObject> oRecommendation = model.getPredictionByVID(document, vid);
 
         if (!oRecommendation.isPresent()) {
-            LOG.error("Could not find annotation in [{}] with id [{}]", document, vid);
+            LOG.error("Could not find prediction in [{}] with id [{}]", document, vid);
+            error("Could not find prediction");
```
reckart: When logging an error to the UI within an AJAX request, it must be ensured that any FeedbackPanel present on the page is also re-rendered. Applies to other `error("...")` calls as well.

Rentier: How would I achieve that?

reckart:
```
aEvent.getTarget().addChildren(getPage(), IFeedback.class);
```

Rentier:
@reckart I just saw instances where there is `error` used but no `log.error`

Issue #19, opened 2018-04-04
Upgrade to RDF4J 2.3.2


reckart:
Upgrade to RDF4J 2.3.0

https://github.com/eclipse/rdf4j/releases/tag/2.3.0

reckart:
Mind I also tried upgrading to 2.4.0-M3, but with that version, I cannot add new repos. It seems to be a bug in the milestone... need to check.

PR #195, opened 2018-05-15
Feature/157 Full-text-search for local knowledge base with LuceneSail

rcffc:
- [x] db changeset that adds the `fullTextSearchIri` needs to provide a default value for existing KBs, otherwise it cannot be applied. It should be changed such that the column is first created nullable, then we run SQL commands to set the the value to the Virtuoso IRI if `supportConceptLinking` is true and to the value that marks that FTS is not supported if `conceptLinking` is false. After this, the column can be set to non-nullable. Thinking about it, we might also just leave the value nullable and take a null value as FTS not being supported.


reckart:
Lots of conflicts. This branch needs attention before it gets completely stale (if it is not already).

reckart:
For #19, I tried upgrading to 2.4.0-M3 - but with that version, I could not add new KBs in INCEpTION anymore. Seems to me to be a bug in the RDF4J milestone - need to double-check. I expect this will bite us as well when we try using our lucene storage fork which is also based on the 2.4.x version.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1140,47 +1194,95 @@ else if (!stmts.hasNext()) {
...
+            log.info("LuceneIndex updated by creating Concept with id [{}] and label [{}].", aId,
+                aLabel);
+        }
+        catch (IOException e) {
+            log.error("Could not add concept with id [{}] and label [{}] to LuceneIndex.", aId,
+                aLabel);
+        }
+    }
+
+    private void indexUpdatedConcept(String aId, String aLabel, IndexWriter aIndexWriter)
```
reckart: We shouldn't have to directly interact with the index at all. It should be entirely sufficient to wrap the native `SailRepositoryConfig` with a `LuceneSailConfig` and then using the additional properties which this sail makes accessible in SPARQL queries: http://docs.rdf4j.org/programming/#_full_text_indexing_with_the_lucene_sail

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
```diff
@@ -54,8 +54,10 @@
 
     @JsonProperty("property-description")
     private IRI propertyDescriptionIri;
-    
-    
+
+    @JsonProperty("full-text-search")
+    private IRI ftsIri;
```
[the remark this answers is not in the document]
rcffc: I had planned to inject a different full-text matching part into the query, depending on kb.getFtsIri()

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -485,6 +546,14 @@ public void updateProperty(KnowledgeBase kb, KBProperty aProperty)
     public void deleteProperty(KnowledgeBase kb, KBProperty aType)
     {
         getReificationStrategy(kb).deleteProperty(kb, aType);
+        if (kb.getType().equals(RepositoryType.LOCAL)) {
+            try {
+                indexDeletedConcept(aType.getIdentifier(), aType.getName(), getIndexWriter(kb));
```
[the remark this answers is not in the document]
rcffc: What do you mean with transparently?
I have tested it without manual indexing, and the label of the KBObject is still selectable after deleting it from the KB.

rcffc:
I have a problem with getting full-text search to work, maybe you can have a look @reckart 
 @Rentier 
https://stackoverflow.com/questions/52368361/lucene-full-text-search-only-works-for-labels-that-match-search-string-exactly

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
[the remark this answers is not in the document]
reckart: What would you do if the user enters an IRI that you don't support? Or would the UI only allow to choose between `bif:contains` and `search:matches`?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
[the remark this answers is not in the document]
reckart: It seems to me that you are creating a second Lucene index in addition to the one that the LuceneSail creates and manages. Why?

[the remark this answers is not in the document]
rcffc: From what I have experienced the Sail is not aware of new KBObjects added/updated/deleted, so I have to notify it manually.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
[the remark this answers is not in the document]
rcffc: It seems like other RDF stores like Jena and YARS also support the Virtuoso-like query. (http://www.l3s.de/~minack/rdf-fulltext-benchmark/) So the Virtuoso part could be the default, and if a special IRI like Lucene is used, a different query part can be injected.

rcffc:
@reckart I have added a test.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -100,15 +106,28 @@
...
+    {
+
+    }
+
+    public ConceptLinkingService(KnowledgeBaseService aKbService,
+        EntityLinkingProperties aProperties)
+    {
+        kbService = aKbService;
+        properties = aProperties;
+        stopwordsFile = new File("${repository.path}/resources/stopwords-en.txt");
```
reckart: The repository path is not substituted here. `${...}` does not work when used as a parameter to "File"

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -215,58 +213,52 @@ public static TupleQuery generateCandidateFullTextQuery(RepositoryConnection con
...
-                + (aKb.getDefaultLanguage() != null ? aKb.getDefaultLanguage() : "en") + "\" )",
+            "  ?e2 ?labelIri ?label.",
             "}");
 
         ValueFactory vf = SimpleValueFactory.getInstance();
 
         TupleQuery tupleQuery = conn.prepareTupleQuery(QueryLanguage.SPARQL, query);
-        tupleQuery.setBinding("string", vf.createLiteral(aString));
+        // add wildcard '*' to perform wildcard search
+        tupleQuery.setBinding("string", vf.createLiteral(aString + "*"));
```
reckart: I guess for Virtuoso, the `*` should not be added, right?

inception-kb/pom.xml
```diff
@@ -58,8 +63,17 @@
       <artifactId>commons-compress</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.rdf4j</groupId>
+      <groupId>org.apache.lucene</groupId>
+      <artifactId>lucene-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.lucene</groupId>
+      <artifactId>lucene-spatial-extras</artifactId>
```
reckart: Why do we need the `spatial` dependency?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -114,6 +124,9 @@
     @org.springframework.beans.factory.annotation.Value(value = "${data.path}/kb")
     private File dataDir;
 
+    @org.springframework.beans.factory.annotation.Value(value = "${repository.path}/luceneIndex")
```
reckart: The KB lucene indexes should be stored next to the KB stores, i.e. under `${repository.path}/kb`.

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
```diff
@@ -331,4 +331,35 @@
...
+              defaultValue="http://www.w3.org/2000/01/rdf-schema#comment">
+        <constraints nullable="false" />
+      </column>
+    </addColumn>
+  </changeSet>
+
+  <changeSet author="INCEpTION Team" id="20180511">
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <columnExists tableName="knowledgebase" columnName="ftsIri"/>
```
reckart: Mind that the project importer needs to be changed as well when a new non-null IRI is added because previously exported projects do not contain the IRI and a default IRI needs to be set on import.

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
```diff
@@ -331,4 +331,35 @@
...
+    <addColumn tableName="knowledgebase">
+      <column name="descriptionIri"
+              type="VARCHAR(255)"
+              defaultValue="http://www.w3.org/2000/01/rdf-schema#comment">
+        <constraints nullable="false" />
+      </column>
+    </addColumn>
+  </changeSet>
+
+  <changeSet author="INCEpTION Team" id="20180511">
```
reckart: `20180511` -> 11th May 2018. That's long past!

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
```diff
@@ -331,4 +331,35 @@
...
+    <preConditions onFail="MARK_RAN">
+      <not>
+        <columnExists tableName="knowledgebase" columnName="ftsIri"/>
+      </not>
+    </preConditions>
+
+    <addColumn tableName="knowledgebase">
+      <column name="ftsIri"
+              type="VARCHAR(255)">
```
reckart: That should be non-null, right? Or can it be null in which case full-text search (i.e. concept linking) would not be supported and we could remove the "supports concept linking" checkbox?

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
```diff
@@ -331,4 +331,35 @@
       </column>
     </addColumn>
   </changeSet>
+
+  <changeSet author="INCEpTION Team" id="20180503">
```
reckart: Why did you add a duplicate of changeset `20180503` to the changelog?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseDetailsPanel.html
```diff
@@ -208,26 +215,27 @@ <h3 wicket:id="name" class="panel-title"></h3>
...
-              <input wicket:id="enabled" type="checkbox" /> <wicket:label key="enabled" />
-            </label>
-          </div>
-          <div class="checkbox">
-            <label wicket:for="supportConceptLinking">
-              <input wicket:id="supportConceptLinking" type="checkbox" />
-                <wicket:label key="kb.supportConceptLinking" />
-            </label>
-          </div>
+        <label class="col-sm-2 control-label">
```
reckart: What is the purpose of the changes in this block?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1184,4 +1197,57 @@ public File readKbFileFromClassPathResource(String aLocation) throws IOException
...
+    private IndexWriter getIndexWriter(KnowledgeBase aKb) throws IOException
+    {
+        Analyzer analyzer = new StandardAnalyzer();
+        Directory directory = FSDirectory
+            .open(new File(luceneIndexDir, aKb.getRepositoryId()).toPath());
+        return new IndexWriter(directory, new IndexWriterConfig(analyzer));
+    }
+
+    @Override
+    public void indexLocalKb(KnowledgeBase aKb)
```
reckart: There is already `LuceneSail.reindex()`. I expect it should be possible to do something like `((LuceneSail) ((SailRepository) repoManager.getRepository(kb.getRepositoryId())).getSail()).reindex()` - obviously not like that but rather with a could of "if / instanceof" checks along the way.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1184,4 +1197,57 @@ public File readKbFileFromClassPathResource(String aLocation) throws IOException
...
+                }
+                conn.commit();
+            }
+        }
+        catch (IOException e) {
+            log.error("Could not index local KB.", e);
+        }
+    }
+
+    private void indexCreatedConcept(String aId, String aLabel, IndexWriter aIndexWriter)
```
reckart: Did you debug if `LuceneSailConnection.commit()` properly updates the index with the stored triples?

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
rcffc: Where do I find the class that needs to be changed?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseDetailsPanel.html
rcffc: Moving the "supportsConceptLinking"  checkbox to the "Type" step.

rcffc:
I have addressed the points that you mentioned.

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/model/db-changelog.xml
reckart: `de.tudarmstadt.ukp.inception.kb.exporter.KnowledgeBaseExporter` and `de.tudarmstadt.ukp.inception.kb.exporter.KnowledgeBaseExporterTest`

reckart:
@rcffc please fix the JavaDoc (pointy brackets need to be written as HTML entities in JavaDoc):

```
[ERROR] /home/svc_jenkins/workspace/INCEpTION (GitHub) (PR builder)/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java:117: error: malformed HTML
[ERROR] * The IRI used for full text search, e.g. bif:contains or <http://www.openrdf.org/contrib/lucenesail#>
[ERROR] ^
[ERROR] /home/svc_jenkins/workspace/INCEpTION (GitHub) (PR builder)/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java:117: error: bad use of '>'
[ERROR] * The IRI used for full text search, e.g. bif:contains or <http://www.openrdf.org/contrib/lucenesail#>
[ERROR] ^
```

Please also see if checkstyle is ok with the code.

rcffc:
@reckart Can you please have another look?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -455,15 +496,15 @@ private String tokensToString(List<Token> aSentence)
         for (Token t : aSentence) {
             joiner.add(t.getCoveredText());
         }
-        return joiner.toString().substring(0, joiner.length() - 1);
+        return joiner.toString().substring(0, (joiner.length() != 0) ? joiner.length() - 1 : 0);
```
Rentier: Can you add a comment why you need to do this?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -254,8 +270,10 @@ public RepositoryImplConfig getNativeConfig()
         // See #221 - Disabled because it is too slow during import
         // return new SailRepositoryConfig(
         //   new ForwardChainingRDFSInferencerConfig(new NativeStoreConfig()));
-        
-        return new SailRepositoryConfig(new NativeStoreConfig());
+
+        LuceneSailConfig config = new LuceneSailConfig(new NativeStoreConfig());
+        config.setIndexDir("indexDir");
```
Rentier: Why dont you set it to `luceneIndexDir`?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
Rentier: Where is this function used?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/ExportedKnowledgeBase.java
```diff
@@ -195,6 +198,16 @@ public void setPropertyDescriptionIri(String aPropertyDescriptionIri)
...
     }
 
+    public String getFullTextSearchIri()
+    {
+        return fullTextSearchIri;
+    }
+
+    public void setFullTextSearchIri(String fullTextSearchIri)
+    {
+        this.fullTextSearchIri = fullTextSearchIri;
```
Rentier: aFullTextSearchIri

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -255,14 +281,17 @@ public void init()
...
 
                 CandidateEntity newEntity = new CandidateEntity(
                     (e2 != null) ? e2.stringValue() : "",
                     (label != null) ? label.stringValue() : "",
-                    (altLabel != null) ? altLabel.stringValue()
+                    (altLabel != null) ? altLabel.stringValue() :
                         // Exact matching does not use altLabel
-                        : (label != null) ? label.stringValue() : "",
-                    (description != null) ? description.stringValue() : "");
+                        (label != null) ? label.stringValue() : "",
```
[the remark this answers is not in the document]
rcffc: It is indented.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -344,31 +373,40 @@ private synchronized Sentence getMentionSentence(JCas aJcas, int aBegin)
...
         Set<String> sentenceContentTokens = new HashSet<>();
-        for (Token t : JCasUtil.selectCovered(Token.class, mentionSentence)) {
-            boolean isNotPartOfMention = !splitMention.contains(t.getCoveredText());
-            boolean isNotStopword = (stopwords == null) || (stopwords != null && !stopwords
-                .contains(t.getCoveredText().toLowerCase(Locale.ENGLISH)));
-            if (isNotPartOfMention && isNotStopword) {
-                sentenceContentTokens.add(t.getCoveredText().toLowerCase(Locale.ENGLISH));
+        List<Token> mentionContext = new ArrayList<>();
+
+        if (aJCas != null) {
```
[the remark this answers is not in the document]
rcffc: For testing, I just pass a "null" for the CAS argument.

rcffc:
For some reason, the Dashboard Page doesn't show the KB page after creation. There is no error, the last messages after creation of a local KB are;

```
2018-10-02 14:04:44 DEBUG [admin] FileUploadDownloadHelper - Creating temporary file for [more_pets.ttl] in [/tmp/inception_file8971659223205485340more_pets.ttl]
2018-10-02 14:04:49 DEBUG [admin] KnowledgeBaseServiceImpl - Stream is not compressed, continue as is.

```
I do not encounter this problem on master @ aea8c297d32cf1e4bdbde03a158749ca91ef2835.

reckart:
@rcffc if you have a directory called "SYSTEM" in your `INCEPTION_HOME/kb/repositories` folder, try renaming it to `SYSTEM.off`. Might be the RDF4J bug.

reckart:
@rcffc actually, it's a bug in the menu item - I'm looking into it

reckart:
@rcffc I believe the menu item bug should be fixed. If not, please let me know. Could you please update the branch, resolve the conflicts, and check that it builds properly on Jenkins?

rcffc:
When I create a new KB, I now get:
```
2018-10-09 15:06:35 ERROR [admin] DefaultExceptionMapper - Unexpected error occurred
org.apache.wicket.WicketRuntimeException: Error attaching this container for rendering: [ListItem [Component id = 0]]
	at org.apache.wicket.MarkupContainer.onBeforeRenderChildren(MarkupContainer.java:1765) ~[wicket-core-8.1.0.jar:8.1.0]
	at org.apache.wicket.Component.onBeforeRender(Component.java:3788) ~[wicket-core-8.1.0.jar:8.1.0]
	at org.apache.wicket.Component.beforeRender(Component.java:937) ~[wicket-core-8.1.0.jar:8.1.0]
...
Caused by: java.lang.IllegalStateException: onConfigure() in the hierarchy of de.tudarmstadt.ukp.inception.ui.kb.project.wizard.KnowledgeBaseCreationWizard$LocalRepositoryStep$1$1 has not called super.onConfigure() at org.apache.wicket.Component.verifySuperCall(Component.java:1013) ~[wicket-core-8.1.0.jar:8.1.0]
...
```
Furthermore, the Title Square looks a bit strange:
selection_098
Experienced on Inception @ d976ec7678ab9d5a036f77defd2faeb425a1702e, Webanno @ 1d54510b25a72c0086e5c598c1898357b317431

reckart:
@rcffc looks like I missed some places when doing the Wicket 8 upgrade. I hope I have covered them all in now in https://github.com/inception-project/inception/pull/544

reckart:
@rcffc The title square issue will be fixed along with https://github.com/inception-project/inception/issues/547

I will run a release of our custom RDF4J lucene module so that we can depend here on a stable (non-SNAPSHOT) version. It would be nice if you could meanwhile resolve the conflicts and make sure the build works so that we can wrap this up.

reckart:
I'm having a look at the conflicts...

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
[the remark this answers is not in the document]
reckart: It's a nested tertiary...  although I don't understand why altLabel is preferred over label.

reckart:
It seems to be **mostly** working now (configuration etc.).

However, when I import the wine ontology into a local KB (OWL mapping) with FTS enabled and create a layer with a KB-linked feature that allows classes/instances from any KB/scope - I still only get auto-completions for "Wine Ontology" and "wine", but for nothing else from the KB.

rcffc:
In the Wine ontology, these two are the only items that have used `rdfs:label`. I am having a look at the build.

rcffc:
The changes in KnowledgeBaseServiceImpl cause the MtasDocumentIndexTests fail to load the ApplicationContext.

reckart:
@rcffc yep, I just noticed it too. I'm trying to figure out how to fix it. The problem seems to be that the KB service tries to access the DB before all the entities are available (or that some entities are not available).

reckart:
> In the Wine ontology, these two are the only items that have used rdfs:label. I am having a look at the build.

I see. I wonder if we can maybe hook into/configure the LuceneSail indexing process somehow so that it also allows us to index the IRI suffixes that we show in the UI if no label is available.

reckart:
@rcffc ok, the MTAS tests should work now

reckart:
> I see. I wonder if we can maybe hook into/configure the LuceneSail indexing process somehow so that it also allows us to index the IRI suffixes that we show in the UI if no label is available.

Looks like that is not easily possible. We'd have to make changes directly to the RDF4J Lucene indexing code. The code already stores an "URI" field in the lucene index, but that cannot be queried - at least not via SPARQL as the analyzer used to parse the query tries to remove special characters. We'd get something like this when trying

```
ERROR LuceneIndex - There was a problem evaluating query 'uri:"http://mbugert.de/pets#animal"' for property 'null!
java.lang.IllegalStateException: field "uri" was indexed without position data; cannot run PhraseQuery (phrase=uri:"http mbugert.de pets animal")
```

Issue #203, opened 2018-05-15
Make external classifier properties configurable


Rentier:
Make the URL of a remote classifier configurable.

- [x] Make the external classifier backend configurable
- [x] Add UI

PR #204, opened 2018-05-15
#203 - Make external classifier properties configurable

Rentier:
- Remove hard coded layer and type
- Start traits editor


Rentier:
@reckart  Right now, the traits editor does not accept the new remote url.

Issue #188, opened 2018-05-14
Accepting active learning recommendations with named entities throws


Rentier:
When I use active learning with a string recommender on named entitie value, then when I want to accept them, the following error is thrown:

`java.lang.String cannot be cast to de.tudarmstadt.ukp.inception.kb.graph.KBHandle`

Full stack trace:

```
2018-05-14 09:53:17 ERROR [klie] AnnotationPage - Error: java.lang.String cannot be cast to de.tudarmstadt.ukp.inception.kb.graph.KBHandle
java.lang.ClassCastException: java.lang.String cannot be cast to de.tudarmstadt.ukp.inception.kb.graph.KBHandle
	at de.tudarmstadt.ukp.inception.ui.kb.feature.ConceptFeatureSupport.setFeatureValue(ConceptFeatureSupport.java:169) ~[inception-ui-kb-0.4.0-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.adapter.TypeAdapter_ImplBase.setFeatureValue(TypeAdapter_ImplBase.java:100) ~[webanno-api-annotation-3.5.0-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.inception.active.learning.sidebar.ActiveLearningSidebar.actionAccept(ActiveLearningSidebar.java:451) ~[inception-active-learning-0.4.0-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.clarin.webanno.support.lambda.LambdaAjaxLink.onClick(LambdaAjaxLink.java:66) [webanno-support-3.5.0-SNAPSHOT.jar!/:?]
	at org.apache.wicket.ajax.markup.html.AjaxLink$1.onEvent(AjaxLink.java:85) [wicket-core-7.10.0.jar!/:7.10.0]
	at org.apache.wicket.ajax.AjaxEventBehavior.respond(AjaxEventBehavior.java:155) [wicket-core-7.10.0.jar!/:7.10.0]
	at org.apache.wicket.ajax.AbstractDefaultAjaxBehavior.onRequest(AbstractDefaultAjaxBehavior.java:601) [wicket-core-7.10.0.jar!/:7.10.0]
	at sun.reflect.GeneratedMethodAccessor437.invoke(Unknown Source) ~[?:?]
[...]

PR #210, opened 2018-05-16
Bug/188 accepting al with ne throws

rcffc:
https://github.com/inception-project/inception/issues/188


rcffc:
Sorry, I didn't see it.

Issue #217, opened 2018-05-16
Concept tree loses scrolling position on select


MdeBoer95:
The concept list/tree and the property list scroll up (loose their scrollbar position) when a concept/property is selected.

This had already been fixed: Change KnowledgeBasePanel so it avoids redrawing the concept tree / property list if a concept / property is selected as that would cause scrolling in the UI element the user has just interacted with... annoying.

However, the change caused new concepts/properties no longer to show up in the tree/list.

Unfortunately, the fix for that basically restores the original state and problem: The concept list/tree and the property list scroll up (loose their scrollbar position) when a concept/property is selected.

PR #218, opened 2018-05-17
#217 - Concept tree loses scrolling position on select

MdeBoer95:
- selection-changed-events contain information whether listpanels should be redrawn or not


reckart:
@MdeBoer95 is it necessary that the property list jumps to the top when a concept is selected and vice versa? I guess the lists only need to be re-drawn if their content changes, e.g. if a new concept/property is added, if a concept/property is deleted or if a concept/property is re-named.

reckart:
@naveen2507 I have updated the branch permissions.

Issue #221, opened 2018-05-17
Unable to import medium-sized local KBs


reckart:
It is not really possible to import medium-sized KBs as local KBs. We are using the `ForwardChainingRDFSInferencer` for local KBs which has a problem: every time a new statement is generated via inference, the whole KB is re-written to disk. Since this can happen very often when importing a KB, it takes ages and hits the disk intensively.

Actually, we should be doing quite fine without the inferencer now that we have the concept tree. So it is time to remove the `ForwardChainingRDFSInferencer`.

PR #222, opened 2018-05-17
#221 - Unable to import medium-sized local KBs

reckart:
- disable the inferencer
- fix test files which wrongly use the rdfs prefix instead of the rdf prefix in some cases


Rentier:
```
[WARNING]    org.eclipse.rdf4j:rdf4j-sail-inferencer:jar:2.2.4:compile
```

Rentier:
@reckart  It would be nice if you could remove the `aIncludeInferred` flag from methods in e.g. KBService, because it would do nothing now from what I think.

reckart:
@rentier it would still do something for remote KBs.

Rentier:
@reckart  You still have that dependecy issue

reckart:
@Rentier I'm presently busy elsewhere; coming back to this later; feel free to fix it.

PR #226, opened 2018-05-19
Feature/225 easier subclassing in kb

MdeBoer95:


naveen2507:
Stacktrace : 

Caused by: java.lang.IllegalArgumentException: Invalid lambda deserialization
	at de.tudarmstadt.ukp.inception.ui.kb.project.KnowledgeBaseIriPanel.$deserializeLambda$(KnowledgeBaseIriPanel.java:1) ~[classes/:?]

reckart:
> Invalid lambda deserialization

This is a common thing when switching between different branches. Wicket serializes the session/page state to disk. Lambdas are quite fragile when it comes to serialization. So when switching between branches, the state can often not be restored from disk. It is not really a problem - you just have to log in again. And in production, it usually happens only once when doing an upgrade.

naveen2507:
@reckart Ok I get that. It works. 
@MdeBoer95 the issues I see with the current functionality are : 
- Multiple time same subclass name is allowed
- Subclasses are not rendered in the concept list on the side panel
- I can't see the details of the concept anymore i.e. subClassOf/type. All I see now is `No statements defined on this detail level.`

MdeBoer95:
@naveen2507 
> Multiple time same subclass name is allowed.   

I agree that it's not a good idea to allow this. We could either just not create a new subclass if the name already exists or automatically rename the new subclass (like in a file system e.g. Cat(2) or something like this).
> Subclasses are not rendered in the concept list on the side panel

>I can't see the details of the concept anymore i.e. subClassOf/type. All I see now is No statements defined on this detail level.

I am aware of this problem. This has something to do with the configuration of the knowledge base. I think it happens because of the reification. Try to create a new one with **Reification: None**. It should display the statements and the new concept should show up in the concept panel.
I don't know yet what exactly is going on here.
Creating the **subclassof statement** causes the new concept to not show up in the panel anymore.(just for wikidata reification)
If I change the code so that just the new concept/subclass is created but not the corresponding statement, it works fine.
Any idea?

MdeBoer95:
I wrote the following test:
```java
    @Test
    public void createSubclassAndListChildConceptsOfParent_ShouldReturnListWithSubclass() {
        
        // parent
        sut.registerKnowledgeBase(kb, sut.getNativeConfig());
        KBConcept concept = buildConcept();
        KBHandle conceptHandle = sut.createConcept(kb, concept);
        
        // child
        KBConcept subConcept = buildConcept();
        subConcept.setName("SubConcept");
        KBHandle subConceptHandle = sut.createConcept(kb, subConcept);
        
        // subclassof property
        KBProperty prop = sut.readProperty(kb, kb.getSubclassIri().stringValue()).get();
        KBHandle propHandle = new KBHandle(prop.getIdentifier(), prop.getName(), prop.getDescription());
        
        assertEquals(propHandle.getIdentifier(), kb.getSubclassIri().stringValue());

        KBStatement stmt = buildStatement(kb, subConceptHandle, propHandle, conceptHandle.getIdentifier());
        
        sut.initStatement(kb, stmt);
        sut.upsertStatement(kb, stmt);
        
        assertThat(sut.listChildConcepts(kb, conceptHandle.getIdentifier(), true))
            .isNotEmpty()
            .contains(subConceptHandle);


    }
```
This works for Reification = NONE
but for Reification = WIKIDATA I get:
**java.lang.AssertionError:  Expecting actual not to be empty**.
Maybe the listChildConcepts method needs to consider the ReificationStrategy?

reckart:
@MdeBoer95 following discussions with @Rentier and @matchlesswei , we do presently not want to have a reified class hierarchy. That means when creating the `x subclassOf y` statement, the reification strategy would need to be bypassed and the statement would need to be created as a simple non-reified triple.

naveen2507:
@MdeBoer95  Error Message : 
`Subclass with name -NativeAmeicanWine- already exists for concept -null- . 
`
In `SubclassCreationDialog.java` , I am changing the parameter `parentConceptHandleModel.getObject().getName()` passed for error to : 
`parentConceptHandleModel.getObject().getUiLabel()` .

Updated Error Message : 
`Subclass with name -NativeAmeicanWine- already exists for concept -AmericanWine-`

MdeBoer95:
@naveen2507 thanks

naveen2507:
@MdeBoer95  Can't see the details of the concept anymore i.e. subClassOf/type. All I see now is No statements defined on this detail level. 

For Reification: None , it works. 

Are you working on it w.r.t  Reification = WIKIDATA ?

MdeBoer95:
@naveen2507 this also happens when I try it on the master branch. I think this changes here are not the cause.

naveen2507:
@MdeBoer95 ok then we can approve the merge request. @reckart  can we raise an issue for this?

naveen2507:
@MdeBoer95 Please raise an issue for the problem we discussed. In Reification = WIKIDATA, we cant see the description  subClassOf/type.

PR #258, opened 2018-06-05
#259 - Use external recommender URL from trait

Rentier:


reckart:
@Rentier if there is a PR, there is an issue. When merging the PR, use "squash" which allows you to change the commit message and to enter the number of the PR.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalClassifier.java
```diff
@@ -108,9 +108,11 @@ public void reconfigure()
         }
         
         //Contruct Http Request
-        // TODO: Use traits url
-        String remoteUrl = "http://localhost:12889/tag";
-        // String remoteUrl = traits.getRemoteUrl();
+        String remoteUrl = traits.getRemoteUrl();
+        if (remoteUrl == null || remoteUrl.isEmpty()) {
```
reckart: Use Apache Commons `StringUtils.isBlank()` ;)

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalClassifier.java
```diff
@@ -108,9 +108,11 @@ public void reconfigure()
...
+            remoteUrl = "http://localhost:12889/tag";
```
reckart: Can we not encode magic default values please?

Rentier:
@reckart I thought about it and think that we should not set a default value in the classifier.

Issue #197, opened 2018-05-15
Support import/export in Recommendation module


reckart:
Implement a `ProjectExporter` for the recommendation module:

- [x] recommender settings which are stored in the DB should be exported to/imported from the JSON structure

Rentier:
@reckart Do you have an example importer/exporter that we can follow?

Rentier:
What happens when I persist a recommender which has a project field inside it? Do I need to set the project again from the outside? What happens to the database generated IDs of the imported stuff?

reckart:
You could have a look e.g. at the `LayerExporter` and the `TagSetExporter`.

Exporters can have dependencies on other exporters (mind that exporters also import - don't worry about the name). So when your exporter depends e.g. on layers, then `getImportDependencies` needs to be overwritten and return a list containing the LayerExporter class. 

Then during import, you need to look up already imported objects by their "business" key instead of their ID. E.g. layers have a unique name within the project, so you look them up by name. The target project to import to is passed as an argument to `.ProjectExporter.importData(ProjectImportRequest, Project, ExportedProject, ZipFile)`.

reckart:
Oh, and on export, you do not persist IDs, but you write the business key (e.g. the layer name). The project ID can be skipped because exports are per-project.

For every DB entity, we have a JSON DTO which is used during export/import (cf. `LayerExporter` and `TagSetExporter`).

PR #264, opened 2018-06-07
#197 - Recommender exporter

Rentier:
- Add recommender exporter
- Add recommender exporter tests


Rentier:
I do not understand why we need `spring-web` all of a sudden.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Recommender.java
```diff
@@ -180,46 +186,35 @@ public void setTraits(String aTraits)
...
-        final int prime = 31;
-        int result = 1;
-        result = prime * result + ((id == null) ? 0 : id.hashCode());
-        result = prime * result + ((name == null) ? 0 : name.hashCode());
-        return result;
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Recommender that = (Recommender) o;
+        return Objects.equals(id, that.id) &&
```
reckart: IMHO hashcode and equals should use name/project ("business" primary key) - id should not be included.

inception-recommendation/pom.xml
```diff
@@ -156,5 +163,27 @@
...
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
```
reckart: assertj dependency should be "test", right?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporter.java
```diff
@@ -0,0 +1,120 @@
...
+            String layerName = exportedRecommender.getLayerName();
+            AnnotationLayer layer = annotationService.getLayer(layerName, aProject);
+            recommender.setLayer(layer);
+
+            recommender.setProject(aProject);
+            recommendationService.createOrUpdateRecommender(recommender);
+        }
+
+        int n = recommenders.length;
+        LOG.info("Imported [{}] recommender for project [{}]", n, aProject.getName());
```
reckart: "recommenders" instead of "recommender" in message?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporter.java
```diff
@@ -0,0 +1,120 @@
...
+            exportedRecommender.setName(recommender.getName());
+            exportedRecommender.setThreshold(recommender.getThreshold());
+            exportedRecommender.setTool(recommender.getTool());
+            exportedRecommender.setTraits(recommender.getTraits());
+            exportedRecommenders.add(exportedRecommender);
+        }
+
+        aExProject.setProperty(KEY, exportedRecommenders);
+        int n = exportedRecommenders.size();
+        LOG.info("Exported [{}] recommender for project [{}]", n, project.getName());
```
reckart: "recommenders" instead of "recommender" in message?

inception-recommendation/src/test/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporterTest.java
```diff
@@ -0,0 +1,134 @@
...
+    @Test
+    public void thatExportingWorks()
+    {
+        // Export the project
+        ProjectExportRequest exportRequest = new ProjectExportRequest();
+        exportRequest.setProject(project);
+        ExportedProject exportedProject = new ExportedProject();
+        File file = mock(File.class);
+
+        sut.exportData(exportRequest, exportedProject, file);
```
reckart: What's a sut? Why not call it "exporter" or something?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Recommender.java
Rentier: I just used what was already there. If you want it to change then I will change it.

inception-recommendation/pom.xml
Rentier: yes

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporter.java
Rentier: Yes

inception-recommendation/src/test/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporterTest.java
Rentier: System under test, to indicate which object we want to test. If you do not like it, then I can rename it for you. https://en.wikipedia.org/wiki/System_under_test

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Recommender.java
reckart: If you prefer, a separate issue/PR can be assigned, but I think it should be addressed.

inception-recommendation/src/test/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporterTest.java
reckart: Though so ;)

> "The term SUT means also a stage of maturity of the software because a system test is the successor of integration test in the testing cycle."

Not sure if this really fits here. But I kind of understand why the name is used here - seems you want to highlight which of the objects being juggled around is the one which is being tested. I also understand that a unit test should not test more than one thing at a time, so the "sut" should be unique. I don't have a strong opinion.

Rentier: > System under test (SUT) refers to a system that is being tested for correct operation.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Recommender.java
Rentier: I created an issue for that.

Rentier:
@reckart I did what you asked

Issue #119, opened 2018-04-27
Create unit tests for search


betoboullosa:

betoboullosa:
Please, review the changes.

PR #265, opened 2018-06-08
#119 - Create unit tests for search 

betoboullosa:


inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
...
+
+            FileSystemUtils.deleteRecursively(new File("target/MtasDocumentIndexTest"));
+        }
+    }
+
+    @Test
+    public void testSimpleQuery() throws Exception
+    {
+        Project project = new Project();
+
```
Rentier: No need for so many new lines

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
...
+        InputStream fileStream = new ByteArrayInputStream(
+                fileContent.getBytes(StandardCharsets.UTF_8));
+
+        documentService.uploadSourceDocument(fileStream, sourceDocument);
+
+        User user = userRepository.get("admin");
+        String query = "Galicia";
+
+        ArrayList<SearchResult> results = (ArrayList<SearchResult>) searchService.query(user,
```
Rentier: `List`

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
...
+        expectedResult.setTokenLength(1);
+
+        ArrayList<SearchResult> expectedResults = new ArrayList<SearchResult>();
+        expectedResults.add(expectedResult);
+
+        assertEquals(results.get(0), expectedResult);
+    }
+
+    @Configuration
+    public static class TestContext
```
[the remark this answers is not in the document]
betoboullosa: I tried to but it didn't work.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: ok

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
...
+    private @Autowired ProjectService projectService;
+    private @Autowired DocumentService documentService;
+    private @Autowired SearchService searchService;
+    private @Autowired AnnotationSchemaService annotationSchemaService;
+
+    // If this is not static, for some reason the value is re-set to false before a
+    // test method is invoked. However, the DB is not reset - and it should not be.
+    // So we need to make this static to ensure that we really only create the user
+    // in the DB and clean the test repository once!
+    private static boolean initialized = false;
```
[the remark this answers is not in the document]
betoboullosa: Doesn't work because then the setup method and the userRepository should be static. Doing so gives me an NPE.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
+/*
+ * Copyright 2017
```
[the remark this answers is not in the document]
betoboullosa: ok

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
```diff
@@ -126,4 +126,68 @@ public void setDocumentTitle(String aDocumentTitle)
     {
         documentTitle = aDocumentTitle;
     }
+
+    @Override
+    public int hashCode()
+    {
+        final int prime = 31;
```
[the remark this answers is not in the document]
betoboullosa: These methods are self-generated by Eclipse, I think they're fine though.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
```diff
@@ -66,11 +67,19 @@
...
     private String dir;
 
+    @Autowired
     public SearchServiceImpl()
     {
         indexes = new HashMap<>();
     }
 
+    public SearchServiceImpl(
+            @org.springframework.beans.factory.annotation.Value("${data.path}") File aDir)
```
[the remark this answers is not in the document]
betoboullosa: ok

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: I think it's more of a personal preference in this case.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,347 @@
...
+        expectedResult.setRightContext("is ");
+        expectedResult.setOffsetStart(15);
+        expectedResult.setOffsetEnd(22);
+        expectedResult.setTokenStart(3);
+        expectedResult.setTokenLength(1);
+
+        ArrayList<SearchResult> expectedResults = new ArrayList<SearchResult>();
+        expectedResults.add(expectedResult);
+
+        assertEquals(results.get(0), expectedResult);
```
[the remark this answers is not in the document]
betoboullosa: What's the benefit here?

betoboullosa:
Please, review the changes.

reckart:
The PR has conflicts with master, could you please resolve them before the review?

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
```diff
@@ -126,4 +126,65 @@ public void setDocumentTitle(String aDocumentTitle)
...
+        int result = 1;
+        result = prime * result + ((documentTitle == null) ? 0 : documentTitle.hashCode());
+        result = prime * result + ((leftContext == null) ? 0 : leftContext.hashCode());
+        result = prime * result + offsetEnd;
+        result = prime * result + offsetStart;
+        result = prime * result + ((rightContext == null) ? 0 : rightContext.hashCode());
+        result = prime * result + ((text == null) ? 0 : text.hashCode());
+        result = prime * result + tokenLength;
+        result = prime * result + tokenStart;
+        return result;
```
reckart: I should be enough to compare the begin/end character offsets and the document identity. Other informations such as left/right context, text, length, etc. is depending on that.

Same in the equals method.

Although I do not have a particularly strong opinion on that over leaving it is as is.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
```diff
@@ -66,11 +67,18 @@
     @Value(value = "${repository.path}")
     private String dir;
 
+    @Autowired
     public SearchServiceImpl()
     {
         indexes = new HashMap<>();
     }
 
+    public SearchServiceImpl(@Value("${data.path}") File aDir)
```
reckart: The value annotation here is redundant with the Value annotation in line 67. The dir could be final. Anyway, we now have a class `RepositoryProperties` which can be injected as a dependency and which should be used instead.

inception-search-mtas/pom.xml
```diff
@@ -113,10 +114,130 @@
...
+
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <scope>test</scope>
```
reckart: Duplicate dependency (junit).

inception-search-mtas/pom.xml
```diff
@@ -113,10 +114,130 @@
...
+    </dependency>
+<!--     <dependency> -->
+<!--     	<groupId>de.tudarmstadt.ukp.clarin.webanno</groupId> -->
+<!--     	<artifactId>webanno-remote</artifactId> -->
+<!--     	<scope>test</scope> -->
+<!--     </dependency> -->
+    <dependency>
+    	<groupId>de.tudarmstadt.ukp.dkpro.core</groupId>
+    	<artifactId>
+    		de.tudarmstadt.ukp.dkpro.core.io.text-asl
```
reckart: Formatting.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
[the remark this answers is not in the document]
reckart: wrt aseertj: nothing in particular. 

However, there should also be an assert about the length of results.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
reckart: Use List instead of ArrayList and remove the type cast.

reckart:
Reviewed. Please address the comments and resolve the conflicts so that we can merge the changes. Thanks!

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
betoboullosa: The contexts could be different, and nothing stops one erroneous method of putting incorrect information in the other fields in a such a way that the dependency that you refer to is broken. In that case, all the fields would be important to technically differentiate between two instances. I would let it as is.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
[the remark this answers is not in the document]
Rentier: It is much easier to read and add new fields if you use `Objects.hash`.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
```diff
@@ -126,4 +126,65 @@ public void setDocumentTitle(String aDocumentTitle)
...
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
```
Rentier: Please use `Objects.equals`

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
```diff
@@ -36,5 +36,6 @@
     public Index getIndex(Project aProject);
 
     public void updateIndex(Index aIndexObject);
-
+    
+    public boolean isIndexValid(Project aProject);
```
Rentier: I think `indexIsValid` or `hasValidIndex` sounds better

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
```diff
@@ -79,17 +79,28 @@
...
+    @Autowired
     public SearchServiceImpl()
     {
         indexes = new HashMap<>();
     }
 
+//<<<<<<< HEAD
+//    public SearchServiceImpl(@Value("${data.path}") File aDir)
+//    {
+//        dir = aDir.getAbsolutePath();
```
Rentier: I would remove these

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
```diff
@@ -79,17 +79,28 @@
...
+//
+//    private Index getIndexByProject(Project aProject)
+//=======
     /** 
      * Get an index entry from the memory map
      * @param aProject The project
      * @return The index
      */
     private Index getIndexFromMemory(Project aProject)
+//>>>>>>> branch '0.4.x' of https://github.com/inception-project/inception.git
```
Rentier: I would remove these

inception-search-mtas/pom.xml
```diff
@@ -29,7 +29,8 @@
   <dependencies>
     <dependency>
       <groupId>de.tudarmstadt.ukp.clarin.webanno</groupId>
-      <artifactId>webanno-api</artifactId>
+      <artifactId>webanno-project</artifactId>
+      <scope>test</scope>
```
Rentier: Please move this down to the other test dependencies

inception-search-mtas/pom.xml
```diff
@@ -118,25 +119,151 @@
...
+    <dependency>
+    	<groupId>net.jodah</groupId>
+    	<artifactId>concurrentunit</artifactId>
+    	<version>0.4.3</version>
+    	<scope>test</scope>
+    </dependency>
+    <dependency>
+    	<groupId>de.tudarmstadt.ukp.dkpro.core</groupId>
+    	<artifactId>
+    		de.tudarmstadt.ukp.dkpro.core.io.conll-asl
```
Rentier: This formatting looks strange

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -477,11 +477,12 @@ public void indexDocument(AnnotationDocument aDocument, JCas aJCas) throws IOExc
     private void deindexDocument(long aSourceDocumentId, long aAnnotationDocumentId, String aUser)
         throws IOException
     {
-        indexWriter.deleteDocuments(new Term(FIELD_ID,
-                String.valueOf(aSourceDocumentId) + "/" + String.valueOf(aAnnotationDocumentId)));
-
-        indexWriter.commit();
+        if (indexWriter != null) {
+            indexWriter.deleteDocuments(new Term(FIELD_ID, String.valueOf(aSourceDocumentId) + "/"
+                    + String.valueOf(aAnnotationDocumentId)));
```
Rentier: Why do you need `String.valueOf`? If you add a long and a String, then it should result in a String.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -595,15 +596,18 @@ public void openPhysicalIndex()
         if (!isOpen) {
             // Only open if it is not already open
             try {
-                log.info("Opening index for project " + project.getName());
+                log.info("indexWriter was not open. Opening it for project " + project.getName());
```
Rentier: You can use

 `log.info("indexWriter was not open. Opening it for project [{}]", project.getName());`

like we do in other places.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -595,15 +596,18 @@ public void openPhysicalIndex()
...
-                log.info("Index has been opened for project " + project.getName());
+
+                log.info("indexWriter has been opened for project " + project.getName());
             }
             catch (Exception e) {
-                log.error("Unable to open index", e);
+                log.error("Unable to open indexWriter", e);
             }
+        } else {
+            log.info("indexWriter is already open for project " + project.getName());
```
Rentier: I would again use string formatting when logging.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -621,12 +625,17 @@ public void createPhysicalIndex()
             FileUtils.forceMkdir(indexDir);
 
             // Open the index
+            log.info("Opening index for project " + project.getName());
```
Rentier: I would again use string formatting when logging.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -671,11 +680,17 @@ private void indexAllDocuments()
     {
         log.info("Indexing all annotation documents of project {}", project.getName());
 
+        int u = 0;
```
Rentier: This is really unreadable, especially with these variable names.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -687,11 +702,16 @@ private void indexAllDocuments()
         for (SourceDocument document : documentService.listSourceDocuments(project)) {
             try {
                 indexDocument(document, documentService.createOrReadInitialCas(document));
+                sd++;
             }
             catch (IOException e) {
                 log.error("Unable to index source document", e);
             }
         }
+
```
Rentier: I would catch the `IOException` in `indexDocument` and log it there once instead of catching it it a lot of different places.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,557 @@
...
+public class MtasDocumentIndexTest
+{
+    private final Logger log = LoggerFactory.getLogger(getClass());
+
+    // Number of miliseconds to wait for the indexing to finish. This time must be enough
+    // to allow the index be built before the query is made. Otherwise, it could affect the
+    // test results. If this happens, a largest value could allow the test to pass.
+    private final int WAIT_TIME = 3000;
+
+    private final String temporaryFolderPath = "target/MtasDocumentIndexTest";
```
Rentier: You can use 

```
  @Rule
  public TemporaryFolder folder= new TemporaryFolder();
```

instead of managing a temporary folder by yourself.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,557 @@
...
+        expectedResult.setOffsetEnd(22);
+        expectedResult.setTokenStart(3);
+        expectedResult.setTokenLength(1);
+
+        ArrayList<SearchResult> expectedResults = new ArrayList<SearchResult>();
+        expectedResults.add(expectedResult);
+
+        assertNotNull(results);
+        if (results != null) {
+            assertEquals(1, results.size());
```
Rentier: Please use `AssertJ`

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,557 @@
...
+        expectedResult.setDocumentTitle("test");
+        expectedResult.setText("Galicia ");
+        expectedResult.setLeftContext("capital of ");
+        expectedResult.setRightContext("is ");
+        expectedResult.setOffsetStart(15);
+        expectedResult.setOffsetEnd(22);
+        expectedResult.setTokenStart(3);
+        expectedResult.setTokenLength(1);
+
+        ArrayList<SearchResult> expectedResults = new ArrayList<SearchResult>();
```
Rentier: `List = ArrayList`

Rentier: You do not need `expectedResults `, do you?

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,557 @@
...
+        new Thread(() -> {
+            try {
+                documentService.uploadSourceDocument(fileStream, sourceDocument);
+            }
+            catch (UIMAException | IOException e) {
+                e.printStackTrace();
+            }
+
+            try {
+                // Wait some time so that the document can be indexed
```
Rentier: Why dont you move this into the try/catch above?

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -0,0 +1,557 @@
...
+    @Before
+    public void setUp()
+    {
+        userRepository.create(new User("admin", Role.ROLE_ADMIN));
+
+        FileSystemUtils.deleteRecursively(new File(temporaryFolderPath));
+    }
+
+    @Test
+    public void testSimpleQuery() throws Exception
```
Rentier: I would name this test so it describes what you actually expect, e.g. `thatSimpleDocumentCanBeIndexedAndRetreived`

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
reckart: I would discern between authoritative information and auxiliary information. I would count the offsets and IDs as authoritative and the text/context strings as auxiliary - reason being that the text/context can be reconstructed from the CAS when IDs/offsets are known. The hash/equals should IMHO rely only on authoritative information.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
reckart: I usually prefer using `DkproTestContext` (which is also a rule) and `DkproTestContext.getTestOutputFolder()` because that places the test output folder under `test-output` in the `target` folder where the data can actually be inspected after the tests. The `TemporaryFolder` is removed after the tests are complete which I find terribly annoying when writing/debugging tests.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
reckart: With SLF4J, concatenation should definitely be avoided. It can have similar undesired effects as code injection, e.g. if the concatenated string contains `{}` (not the case here, but still).

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchResult.java
reckart: Eclipse's auto-generated hashcode/equals do not make use of `Objects.equals`. It is quite annoying to manually write hashcode/equals, even with these convenience functions. 

There is an open issue for Eclipse to support `Objects`: https://bugs.eclipse.org/bugs/show_bug.cgi?id=424214

There also seems to be a recent plugin which generates the more compact versions - we could try that: https://marketplace.eclipse.org/content/jenerate

But IMHO not using that is not a blocker as long as the hashcode/equals are auto-generated from the correct set of fields.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
betoboullosa: I think both your suggestions lead to a misinterpretation of the method's goal. 1) indexIsValid does not sound as a question. 2) As for hasValidIndex, the method must answer the question about the state of the index, not about its existence.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
betoboullosa: OK.

inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchServiceImpl.java
betoboullosa: OK.

inception-search-mtas/pom.xml
betoboullosa: OK.

inception-search-mtas/pom.xml
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK.

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: It's important to know and log the different points in which it may occur.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: Why? Do we have a standard for this?

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: It was a leftover, I've removed it.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: OK.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
Rentier: I think it reads better and you can describe what you are asserting. You also added it in the pom so I assumed you want to use it.
`assertThat(results).as("Results could be retrieved").hasSize(1)''

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Rentier: You know the points because of the stack trace in the exception you log.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: I haven't been able to use a temporary folder rule, since I need the temporary folder to be initialized in the text context. Cf. https://stackoverflow.com/questions/24922153/spring-config-with-junit-temporary-folder

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
betoboullosa: The assert was in the pom probably due some copy-paste. Anyway, I like the idea of using the hasSize method, but I haven't been able to use it. The command that you've suggested gives me this syntax error: The method assertThat(T, Matcher<? super T>) in the type Assert is not applicable for the arguments (List<SearchResult>)

inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
betoboullosa: OK

betoboullosa:
Tests have passed, I think all pending comments have been addressed.

Rentier:
Can you briefly document in the test why we need the threading?

PR #267, opened 2018-06-09
Feature/155 unexpected exception

reckart:
We cannot use feedback messages in LoadableDetachableModel.load(), so using a "error" object as a workaround


inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/SubjectObjectFeatureEditor.java
```diff
@@ -293,43 +294,61 @@ protected IJQueryTemplate newTemplate()
...
+            }
+            
+            try {
                 if (selectedKBItemIdentifier != null) {
                     ConceptFeatureTraits traits = factService.getFeatureTraits(project);
                     selectedKBHandleItem = factService.getKBInstancesByIdentifierAndTraits
                         (selectedKBItemIdentifier, project, traits);
                 }
+                
+                throw new RuntimeException("test");
```
Rentier: It might be good to remove this

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/SubjectObjectFeatureEditor.java
```diff
@@ -293,43 +294,61 @@ protected IJQueryTemplate newTemplate()
...
+                selectedKBItemIdentifier = WebAnnoCasUtil
                     .getFeature(selectedFS, linkedAnnotationFeature.getName());
-
+            }
+            catch (Exception e) {
+                LOG.error("Error: " + e.getMessage(), e);
+                // We cannot use feedback messages in code that is called from the load() method
+                // of a LoadableDetachableModel, so this is an alternative way of passing the
+                // error on to the user.
+                selectedKBHandleItem = KBHandle.errorHandle("Unable to load CAS", e);
```
Rentier: Why no early return here?

reckart:
Indeed :)

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/SubjectObjectFeatureEditor.java
reckart: Sometimes I tend to avoid returns in the middle of a method because it's like a goto.

Rentier: From what I see in the error case, you will check whether `selectedKBHandleItem ``is not null, which it is not in the error case, then you will try to `getKBInstancesByIdentifierAndTraits`, which should error again. I would just error once.

reckart: I might be missing something, but the second case shouldn't be triggered since the selectedKBItemIdentifier will remain null. 

I'm rearranging the code in https://github.com/inception-project/inception/pull/268 - please have another look there. Thanks!

Rentier: You are right, these are two selected thingies, sorry for bothering you!

reckart: No worries. Better double check than having bad code... as with the extra throws I forgot to remove ;)

PR #268, opened 2018-06-09
#155 - Unexpected RuntimeException when linking facts

reckart:
- Removing exception used during debugging.


Rentier:
What do you think about creating a class `KBErrorHandle extends KBHandle` and add the methods relating to errors into the KBHandle (e.g. `isError`, with default implementation to `false` for `KBHandle` and true for `KBErrorHandle `? Yoi could also add the error methods to `KBObject`. I like it more, because magic strings for identifiers are a bit strange. also. the error code would be in one or two places.

reckart:
If I would create a KBErrorHandle subclass, why would I introduce an isError() method on KBHandle instead of using instanceof?

Issue #272, opened 2018-06-11
Refactoring of feature supports


reckart:
The de-coupling of getting/setting feature values from converting between feature editor and CAS representations will require some changes.

betoboullosa:
Done.

PR #273, opened 2018-06-11
#272 - Refactoring of feature supports

reckart:
- Adjust to refactorings for https://github.com/webanno/webanno/issues/985
- ConceptFeatureSupport and PropertyFeatureSupport re-use renderFeatureValue code when constructing the KBHandles instead of implementing the code to get the handle twice


reckart:

reckart:

reckart:
@matchlesswei this PR (hopefully) contains the changes necessary for #149 as we discussed in the meeting with @betoboullosa. Use the `unwrapFeatureValue` to turn the value you get from the feature editor into the value that should be stored in the LearningRecord. Use something like `value != null ? String.valueOf(value) : null` to convert the object value to a string value.

reckart:
>  @reckart why do we need CAS aCAS as an input param for the wrapFeatureValue() and unwrapFeatureValue()? Any specific reason w.r.t future work?

unwrap converts the feature editor value into a CAS value. We would already be making use of this parameter if the handling of slot features wasn't baked in so deeply in the code. When this slot support is refactored, we will need this parameter.

PR #280, opened 2018-06-12
Feature/136 concepts class hierarchy

naveen2507:
Allow concepts that are not explicitly class instances

- Improve listing of root concepts 
- List child concepts even if they do not have an explicit instance-of statement


Rentier:
Please add unit tests for this in the KBServiceImplIntegration test

naveen2507:

PR #305, opened 2018-06-20
Feature/196 support import export in kb module

MdeBoer95:


inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/KnowledgeBaseExporter.java
```diff
@@ -0,0 +1,193 @@
...
+            exportedKB.setTypeIri(kb.getTypeIri().stringValue());
+            exportedKB.setDescriptionIri(kb.getDescriptionIri().stringValue());
+            exportedKB.setLabelIri(kb.getLabelIri().stringValue());
+            exportedKB.setPropertyTypeIri(kb.getPropertyTypeIri().stringValue());
+            exportedKB.setReadOnly(kb.isReadOnly());
+            exportedKB.setEnabled(kb.isEnabled());
+            exportedKB.setReification(kb.getReification().toString());
+            exportedKB.setSupportConceptLinking(kb.isSupportConceptLinking());
+            exportedKB.setBasePrefix(kb.getBasePrefix());
+            exportedKnowledgeBases.add(exportedKB);
```
naveen2507: @reckart Do we need to export anything else apart from these? Are we looking for the changes made in KB as well?

Rentier:
I would like to see a unit test like the `RecommenderExporterTest`. You can use some random RDF for the input and the `TemporaryFolder` rule for your target file. I would just check whether the import was called with Mockitos `verifify`, not start a Spring environment.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/KnowledgeBaseExporter.java
reckart: Basically, we need all the properties of the `KnowledgeBase` class so that we can re-create it from the JSON DTO. What do you mean by "the changes made in the KB"?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/KnowledgeBaseExporter.java
```diff
@@ -0,0 +1,193 @@
...
+import de.tudarmstadt.ukp.inception.kb.model.KnowledgeBase;
+import de.tudarmstadt.ukp.inception.kb.reification.Reification;
+
+@Component
+public class KnowledgeBaseExporter implements ProjectExporter
+{
+    private static final String KEY = "knowledgeBases";
+    private static final Logger LOG = LoggerFactory.getLogger(KnowledgeBaseExporter.class);
+    
+    private static final String SOURCE = "knowledgeBase_source";
```
reckart: Just store it under "kb"? Please rename SOURCE to KB and SOURCE_FOLDER to KB_FOLDER.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/KnowledgeBaseExporter.java
```diff
@@ -0,0 +1,193 @@
...
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.inception.kb.KnowledgeBaseService;
+import de.tudarmstadt.ukp.inception.kb.RepositoryType;
+import de.tudarmstadt.ukp.inception.kb.model.KnowledgeBase;
+import de.tudarmstadt.ukp.inception.kb.reification.Reification;
+
+@Component
+public class KnowledgeBaseExporter implements ProjectExporter
+{
+    private static final String KEY = "knowledgeBases";
```
reckart: Please use `knowledge_bases`.

reckart:
Also check the build status on Jenkins please.

naveen2507:
@MdeBoer95  can u check the dependencies as for Jenkins it gives an error 

inception-kb: Dependency problems found for used undeclared dependencies.

MdeBoer95:
@Rentier 
> I would just check whether the import was called with Mockitos verifify, not start a Spring environment.

Do you mean something like this?
```java
 verify(sut, times(1)).importData(importRequest, project, exportedProject, zipFile);
```
And not check what the imported KB looks like ?

Rentier:
I would veriffy on the knowledge base service that import was called.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseExporterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        sut.importData(importRequest, project, exportedProject, zipFile);
+
+        // Check how many localKBs have been exported
+        List<KnowledgeBase> exportedKbs = exportKbCaptor.getAllValues();
+        int numOfLocalKBs = exportedKbs.stream()
+                .filter(kb -> kb.getType().equals(RepositoryType.LOCAL))
+                .collect(Collectors.toList()).size();
+
+        // Verify that importData is called as many times as there are localKBs
+        ArgumentCaptor<String> fileNameCaptor = ArgumentCaptor.forClass(String.class);
```
Rentier: You do not need argument captors for that, you can just use matchers like any (https://static.javadoc.io/org.mockito/mockito-core/1.9.5/org/mockito/Matchers.html). You use captors if you want to later assert on the parameters

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/ExportedKnowledgeBase.java
```diff
@@ -0,0 +1,208 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+
+@JsonPropertyOrder(alphabetic = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ExportedKnowledgeBase
+{    
+    @JsonProperty("name")
```
reckart: We use `snake_case` instead of `camelCase` for the JSON properties, e.g.

```
    @JsonProperty("project_name")
    private String projectName;

    @JsonProperty("linked_list_behavior")
    private boolean linkedListBehavior;

    @JsonProperty("on_click_javascript_action")
    private String onClickJavascriptAction;
```

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/KnowledgeBaseExporter.java
```diff
@@ -0,0 +1,193 @@
...
+import de.tudarmstadt.ukp.inception.kb.model.KnowledgeBase;
+import de.tudarmstadt.ukp.inception.kb.reification.Reification;
+
+@Component
+public class KnowledgeBaseExporter implements ProjectExporter
+{
+    private static final String KEY = "knowledge_bases";
+    private static final Logger LOG = LoggerFactory.getLogger(KnowledgeBaseExporter.class);
+    
+    private static final String KB = "kb";
```
Rentier: I would either use the `File(String parent, String child)` constructor or Java 8 `Paths.get`. If you concatenate by hand, then there can be weird errors with Windows using a different file path seperator than `/`.

reckart: Paths with slashes work nice on Windows. I always write my paths using slashes. Using `File(a, b)` is useful when you want to avoid problems with duplicate slashes when concatenating paths. But usually duplicate slashes are also unproblematic.

Issue #316, opened 2018-06-20
Accepting AL recommendation while in other document fails


reckart:
Consider this case:

* activate AL and ensure it provides a recommendation in some document A
* move to another document B in the project
* now press accept

Instead of creating the annotation in document A, the AL sidebar will try to create the annotation at the same position in document B which is very likely to fail horribly.

PR #317, opened 2018-06-20
#316 - Accepting AL recommendation while in other document fails

reckart:
- Store document in which recommendation exists instead of current document in the learning record
- Avoid exceptions when reacting to annotation update event
- Create annotation in the right document
- Made trainning code more robust

### DIFF
--- a/inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
+++ b/inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
@@ -26,6 +26,7 @@ import org.apache.uima.cas.Type;
 import org.apache.uima.cas.text.AnnotationFS;
 import org.apache.uima.fit.util.CasUtil;
 import org.apache.uima.jcas.JCas;
+import org.apache.wicket.Component;
 import org.apache.wicket.ajax.AjaxRequestTarget;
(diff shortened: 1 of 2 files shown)


inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -589,15 +602,20 @@ public void afterAnnotationUpdateEvent(AjaxAfterAnnotationUpdateEvent aEvent)
...
+                eventState.getUser().equals(annotatorState.getUser()) && 
+                eventState.getDocument().equals(annotatorState.getDocument()) && 
+                annotatorState.getSelectedAnnotationLayer().equals(selectedLayer.getObject())
+        ) {
             //check same document and same token
             if (annotatorState.getSelection().getBegin() == currentRecommendation.getOffset()
                 .getBeginCharacter()
                 && annotatorState.getSelection().getEnd() == currentRecommendation.getOffset()
-                .getEndCharacter() && aEvent.getValue() != null) {
+                .getEndCharacter()
```
matchlesswei: I think  `aEvent.getValue() != null` should be kept. 
If `aEvent.getValue()` of `AjaxAfterAnnotationUpdateEvent` is null, it means that a user has selected the token in the mainEditor, but hasn't made any annotation. So at this time, AL sidebar doesn't need to update the suggestion.

PR #319, opened 2018-06-21
Feature/297 show IRIs on mouse over

naveen2507:


naveen2507:
Tooltip for statement properties works 

tooltipscreenshot

naveen2507:
But for copying the content Mootip Settings provide an option to fix it. But its not what looks good on UI. 

mootipscreenshot

reckart:
Looks like the Kendo tooltip should stay visible as long as the mouse is either on the target element or in the tooltip area: https://www.telerik.com/forums/is-there-a-way-to-set-a-close-delay-for-a-tooltip-

The post above suggests to adjust the CSS to make it easier to move the mouse cursor from the target element into the tooltip area - if we can do that successfully, I guess we should also be able to copy the tooltip contents.

naveen2507:
image

The IRI can be copied from the tooltip and the tool tip can be closed thereafter.

Issue #150, opened 2018-05-02
Logging results of recommender evaluation runs


matchlesswei:
Results of recommender evaluation should be logged

PR #320, opened 2018-06-21
#150 - Logging results of recommender evaluation runs

reckart:
- Improve the logging of results
- Improve robustness of the training and selection tasks by catching Throwable instead of Exception at relevant points


reckart:
I think the current calculation of TN/FN/TP/FP and possibly of precision/recall/f-score in the `ExtendedResult` may not be good.

E.g. considering every token on which there is no annotation in the gold standard and also no annotation in the classifier results as a TN yields "too good" results. This definitely needs to be checked, best with unit tests.

Any comments/suggestions?

https://github.com/inception-project/inception/blob/ccf039e94bc92fbdb40e219c7ae025db94b1bff0/inception-imls-core/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/core/dataobjects/ExtendedResult.java#L74-L150

reckart:

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/log/RecommenderEvaluationResultEventAdapter.java
```diff
@@ -0,0 +1,125 @@
...
+        return aEvent.getUser();
+    }
+    
+    @Override
+    public String getDetails(RecommenderEvaluationResultEvent aEvent)
+    {
+        try {
+            Details details = new Details();
+            
+            details.score = aEvent.getScore();
```
Rentier: Can we remove this commented out stuff? I do not like commented out stuff.

reckart: Will do before merge and then put to the follow up PR ;)

Issue #324, opened 2018-06-22
Suddenly lost all annotations


reckart:
While trying out the recommenders and annotating, at some moment all my annotation disappeared - I only saw some suggestions but all the named entity annotations that I had created were gone.

I don't know exactly what I did to make this happen. 

Maybe some sanity checking mechanism could be introduced which alarms if the number of annotations in a CAS suddenly drops - this should at least help tracking down where this problem occurs.

reckart:
Got a stack trace where this happens:

```
Daemon Thread [ForkJoinPool.commonPool-worker-1] (Suspended (breakpoint at line 216 in CasStorageServiceImpl))	
	owns: Object  (id=183)	
	CasStorageServiceImpl.realWriteCas(Project, String, long, JCas, String, File, File) line: 216	
	CasStorageServiceImpl.writeCas(SourceDocument, JCas, String) line: 153	
	DocumentServiceImpl.readAnnotationCas(AnnotationDocument, boolean) line: 685	
	DocumentServiceImpl.readAnnotationCas(AnnotationDocument) line: 654	
	DocumentServiceImpl$$FastClassBySpringCGLIB$$7e8e7e07.invoke(int, Object, Object[]) line: not available	
	MethodProxy.invoke(Object, Object[]) line: 204	
	CglibAopProxy$CglibMethodInvocation.invokeJoinpoint() line: 747	
	CglibAopProxy$CglibMethodInvocation(ReflectiveMethodInvocation).proceed() line: 163	
	851370092.proceedWithInvocation() line: not available	
	TransactionInterceptor(TransactionAspectSupport).invokeWithinTransaction(Method, Class<?>, InvocationCallback) line: 294	
	TransactionInterceptor.invoke(MethodInvocation) line: 98	
	CglibAopProxy$CglibMethodInvocation(ReflectiveMethodInvocation).proceed() line: 185	
	CglibAopProxy$DynamicAdvisedInterceptor.intercept(Object, Method, Object[], MethodProxy) line: 689	
	DocumentServiceImpl$$EnhancerBySpringCGLIB$$e226d662.readAnnotationCas(AnnotationDocument) line: not available	
	NamedEntityLinker.lambda$4(TokenObject, KnowledgeBase, RepositoryConnection) line: 218	
	874775620.accept(RepositoryConnection) line: not available	
	KnowledgeBaseServiceImpl.read(KnowledgeBase, ReadAction<T>) line: 614	
[...]

PR #334, opened 2018-06-25
#324 - Suddenly lost all annotations

reckart:
- Adjusting to changes made for https://github.com/webanno/webanno/issues/1005


naveen2507:
@reckart Build fails for `inception-ui-kb`

Issue #323, opened 2018-06-22
Changing IRIs has delayed effect


reckart:
Due to the caching of KB information, changing IRIs in the project settings has no immediate effect.

E.g. when I notice that the IRI for descriptions is set wrongly and I got to the KB settings to correct it and go back to the annotation screen, recommendations and auto-completes in the concept feature editors that I have used previously still show no descriptions. Only when I search for other concepts, I get descriptions.

Similar effects could occur when changes are made in the KB, e.g.  when a user update a description.

Flush the cache on a ChangeIriEvent.

PR #341, opened 2018-06-26
#323 - Changing IRIs has delayed effect

rcffc:
Created IriChangeEvent.
Listen on IriChangeEvent when value of IRI ComboBox has changed and clear caches.


reckart:
I would go even further and use a `KnowledgeBaseConfigurationChangedEvent` (similar to `LayerConfigurationChangedEvent`) that triggers always when a KB is saved no matter whether something actually changed or not and no matter which setting was changed.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/KnowledgeBaseIriPanel.java
```diff
@@ -155,7 +155,7 @@ public void onSelectionChanged(AjaxRequestTarget target, SchemaProfile bean)
         comboBox.setRequired(true);
         comboBox.add(Validators.IRI_VALIDATOR);
         comboBox.add(new LambdaAjaxFormComponentUpdatingBehavior("change", t -> {
-            // Do nothing just update the model values
+
```
reckart: Please retain the comment. Empty code blocks should always be commented, otherwise it looks like a bug/missing code.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/event/KnowledgeBaseConfigurationChangedEvent.java
```diff
@@ -0,0 +1,32 @@
...
+package de.tudarmstadt.ukp.inception.kb.event;
+
+import org.springframework.context.ApplicationEvent;
+
+public class KnowledgeBaseConfigurationChangedEvent
+    extends ApplicationEvent
+{
+    private static final long serialVersionUID = -2741267700429534514L;
+
+    public KnowledgeBaseConfigurationChangedEvent(Object aSource)
```
reckart: Please include the project.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/service/ConceptLinkingService.java
```diff
@@ -468,4 +470,12 @@ private SemanticSignature getSemanticSignature(KnowledgeBase aKB, String aWikida
             .collect(Collectors.toList());
     }
 
+    @EventListener
+    public void onKnowledgeBaseConfigurationChangedEvent(
+        KnowledgeBaseConfigurationChangedEvent aEvent)
+    {
+        candidateCache.clear();
```
reckart: If the caches are cross-project and the projects have different KBs using same IRIs but different values for them, then the cache would leak across projects. 

I believe the caches must be per-project and when receiving the event, only the cache for the respective project must be clear. In that case, there must also be some time-based cache-expunge strategy since we can have arbitrary numbers of project and if there is no automatic cache clearing, then the caches could easily consume lots of memory and starve the system.

An alternative would be to include the project in the cache key and when clearing the caches on a KB change, drop only those cache entries with a matching project. In this case, all projects share a common cache which is memory-bounded and wouldn't cause the system to starve even if not regularly cleared up.

The second alternative is probable the better one.

reckart:
@rcffc checkstyle issues

PR #351, opened 2018-06-28
Feature/296 move definition of knowledge base profiles into a file

MdeBoer95:


MdeBoer95:
It works with this constructor.
https://github.com/inception-project/inception/blob/aec258d5192ad2a845423c278f0353134b852271/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java#L35-L50

 So why not use it like that?

Rentier:
You could also implement Jackson Serializer and Deserializer for this, it is not so difficult. E.g. http://www.baeldung.com/jackson-custom-serialization

reckart:
I'm a bit worried if it is by design or accidental that the values are serialized correctly from IRI to string. But ok. In fact, we use this only for deserialization, not for serialization. So if we have a unit test for deserialization that works, let's leave it at the solution with the constructor that you propose. However, please also include a no-args constructor. Serializable classes should always have one.

MdeBoer95:
The custom Serializer/Deserializer is nice but I think a it's a bit overkill for this. I will just implement a unit test for now.

reckart:
@MdeBoer95 the build fails, but I assume the PR is still a work-in-progress, right? Could you please add the tag "WIP"?

MdeBoer95:
It is basically done. I don't know why the tests fail, they are all successful on my local machine.

reckart:
Why do `KnowledgeBaseProfile` and `KnowledgeBaseMapping` require a hashcode/equals implementation?

Issue #352, opened 2018-06-29
No properties shown in KB UI 


naveen2507:
**Describe the bug**
No properties for ontologies for new project created. (Please check if somebody can reproduce)

**To Reproduce**
1. Create a new project
2. Add a local ontology - I took wine ontology
3. Add a document to the project
4. Go to KB UI and check if the properties are visible for you 

**Expected behavior**
List of properties on KB UI 

**Screenshots**
![image](https://user-images.githubusercontent.com/10420950/42077747-22eec878-7b7a-11e8-9f46-6fa93599a11e.png)

**Please complete the following information:**
 -  Technische Universität Darmstadt -- Computer Science Department -- INCEpTION 𝜶 -- 0.4.0-SNAPSHOT (2018-06-28 14:45:31, build 492d8c20bf4e6f0e2b12546cb56020ca1c90a71b) - master branch
 - Linux
 - chrome

naveen2507:
#322 Duplicate Issue

naveen2507:
Issue covered by PR #394

reckart:
Since the PR is not yet merged, the issue should not be closed yet, right?

naveen2507:
I have closed the PR, for the KB labelled issues there are lot of issue dependent on each other and overlap the fixes. Like Statement validation is dependent on the issues related to property. WDYS?

reckart:
Ok, I understood from your comment that PR #394 covers (i.e. includes) the changes for this issue, not that #322 (the issue governing #394) supersedes this issue.

reckart:
But actually... anyway. As long as the bug still exists, it should not be closed. Only once the properties are shown, it should be closed. Otherwise there is a risk that the issue will be opened again. As far as I can see PR #394 is not yet merged and does not yet fix the issue. So reopening it.

naveen2507:
To be closed after closing PR #394

naveen2507:
> To be closed after closing PR #394

Closing issue now.

PR #354, opened 2018-06-29
#352 No properties shown in KB UI

naveen2507:
- Properties in KB UI
- Pending: Number of issues after removing the inferencer. Inception
code doesn't identify any obvious relation for SPARQL queries. Need
debugging and changes for most queries.


naveen2507:
@reckart  Can't have generic queries for in `getStatementsSparql()` anymore.

reckart:
> @reckart Can't have generic queries for in getStatementsSparql() anymore.

Can you be more specific?

naveen2507:
@reckart Now as we don't have the Inferencer anymore, the code doesn't recognize that ObjectProperty, DatatypeProperty, etc are subProperty of Property.  The method getStatementsSparql()  is pretty generic for now and is called for reading concept, instance, and property. But what I see now is we have to cover every aspect in SPARQL queries for reading property as we don't have the support from inferencer anymore.

reckart:
@naveen2507 We could re-enable the inferencer but it would mean that the internal KB can really only pretty small.

naveen2507:
@reckart Let's discuss this. We can get rid of inferencer but it comes with its own pros and cons.

naveen2507:
@reckart I think if we have serious issues with inferencer then we can get rid of that, but in case

naveen2507:
@reckart I think if we have serious issues with working on internal KB's we should get rid of inferencer or we can also explore for some lightweight inferencer for some basics,  or we can ger rid of inferencers and design separate and in depth SPARQL queries. We can have a list of semantic relations,  which we can use as part of our SPARQL query.

reckart:
@naveen2507 I had a look some time back on how the inferencer and native storage interplay in RDF4J and this is what I remember:

1. whenever a triple is added to the KB, the inferencer kicks in
2. every time the inferencer kicks in, it iterates over all triples in the KB and tries to apply its inference roles
3. (on this one I am not entirely sure...) either a) when one or more triple is has been added after applying all inference rule to a given triple, the entire KB is written to disk or b) once all inference rules have passed over all triples the KB is written to disk - I think it is a)
4. if any triple has been added by the inferencer, then go back to step 2

You can easily imagine that this is prohibitive for any but small KBs. RDF4J documentation also warns about using the inferencer. The strategy that they use works ok for KBs that are kept in-memory, but not for on-disk KBs.

Mind this is from memory... I might have missed/forgotten something.

So the options I see are:

* we could choose to use in-memory KBs with the inferencer and only store them to disk from time-to-time. No idea how much main memory use that would incur.
* we could choose to use the disk KB without the inferencer but then we have to write rather complicated queries (the status we have now)
* we could choose to use the disk KB with the inferencer but then we can only have very small internal KBs and for anything non-trivial an external SPARQL service would be required. That would increase pressure for us to implement write support for external KBs (currently they are read-only)
* ... any other suggestions?

naveen2507:
@reckart I dont see any other option for now.  
RDF4J documentation states that : `However, a word of warning: the RDF4J inferencers add a significant performance overhead when adding and removing data to a repository, an overhead that gets progressively worse as the total size of the repository increases. For small to medium-sized datasets it peforms fine, but for larger datasets you are advised not to use it and to switch to alternatives.` 

Have we explored anything on the line of this : 
http://docs.rdf4j.org/programming/#_reasoning_and_validation_suport_with_spin

reckart:
> Have we explored anything on the line of this : http://docs.rdf4j.org/programming/#_reasoning_and_validation_suport_with_spin

Well, we have used the `ForwardChainingRDFSInferencer` before... not sure what you mean?

What do you think about the option of using a memory store and persisting that to disk from time-to-time?

reckart:
@naveen2507 Btw, this issue here confirms my analysis: https://github.com/eclipse/rdf4j/issues/462

naveen2507:
> we could choose to use in-memory KBs with the inferencer and only store them to disk from time-to-time. No idea how much main memory use that would incur.

Yeah i think this would be the most feasible option for us rt now.

reckart:
>> we could choose to use in-memory KBs with the inferencer and only store them to disk from time-to-time. No idea how much main memory use that would incur.
> Yeah i think this would be the most feasible option for us rt now.
I think it would be useful to create a benchmark in the form of a unit test which tries to import WordNet RDF into a native store as well as into a memory store with the inferencer enabled. This should be written without any dependency on INCEpTION in general but can be placed in the KB module. We'd normally mark these tests using `@Ignore` but keep them around in case we need to benchmark different setups.

Rentier:
One can also change the configuration before importing to not use the inferencer, then import, then use the previous configuration. One might need to trigger the inferencer then afterwards once.

reckart:
> One can also change the configuration before importing to not use the inferencer, then import, then use the previous configuration. One might need to trigger the inferencer then afterwards once.

There might be a problem with that when trying to import a **second** RDF file into an existing KB. I'm also not sure how this will work with the repo manager. Can we actually easily change a repo-manager configuration after it has been created? Any idea how this would impact concurrent connections to the KB?

Rentier:
I tried it and it did not complain. The doc (http://docs.rdf4j.org/javadoc/1.0/org/eclipse/rdf4j/repository/manager/RepositoryManager.html#addRepositoryConfig-org.eclipse.rdf4j.repository.config.RepositoryConfig-) says: 

> Adds or updates the configuration of a repository to the manager's system repository. The system repository may already contain a configuration for a repository with the same ID as specified by config, in which case all previous configuration data for that repository will be cleared before the new configuration is added.

The problem with concurrency is already there: what do you do if someone imports a KB and another user edits it at the same time?

reckart:
> The problem with concurrency is already there: what do you do if someone imports a KB and another user edits it at the same time?

The repo supports transactions. So in principle concurrent update should not cause significant problems.

naveen2507:
Issue covered by #394

Issue #306, opened 2018-06-20
Unable to link "Indonesia"


reckart:
The auto-complete only offers very strange choices

![2018-06-20_11-29-21](https://user-images.githubusercontent.com/1410238/41650185-be55a4c0-747d-11e8-99a4-5167d0c20742.png)

A SPARQL query on the KB reveals though that the entity we are looking for is indeed contained in the KB

![2018-06-20_11-33-14](https://user-images.githubusercontent.com/1410238/41650239-d8645708-747d-11e8-85dc-b0a882e0506a.png)

rcffc:
![screenshot from 2018-06-22 23-19-20](https://user-images.githubusercontent.com/13070150/41799724-d5fee3d8-7672-11e8-9293-68ae21ee7168.png)
Cannot reproduce this error. I suspect it is because of the EntityLinking properties currently active on blinky-experimental. Try with these settings that I use:
```
inception.entity-linking.cacheSize = 1024
inception.entity-linking.mentionContextSize = 5
inception.entity-linking.candidateQueryLimit = 10000
inception.entity-linking.candidateDisplayLimit = 20
inception.entity-linking.candidateFrequencyThreshold = 25
inception.entity-linking.signatureQueryLimit = 2147483647
```

reckart:
Try writing "Indonesia" instead of "indo".

reckart:
btw. it seems that if I write a prefix of the mention string into the field, the search always is for the mention string. E.g. if I write "indo" on a mention of "indonesia", the log would still says that it searched for "indonesia" instead of "indo".

You can also try something else. Search e.g. for "2017". I got a lot of stuff but not the year. Searching for "year 2017" also didn't help, but e.g. searching for "MMXVII" returned the year.

rcffc:
> btw. it seems that if I write a prefix of the mention string into the field, the search always is for the mention string

That was done intentionally to improve the performance, see https://github.com/inception-project/inception/issues/87

rcffc:
> Search e.g. for "2017"

![screenshot from 2018-06-22 23-35-22](https://user-images.githubusercontent.com/13070150/41800268-1936ec0c-7675-11e8-8ec3-b933b1cfc9fc.png)

Typing "Indonesia" also lead to the correct result.

reckart:
> That was done intentionally to improve the performance, see #87

But it kind of prevents matching against an entity which in the KB has a prefix of the mention as its label, doesn't it?

[...]

PR #358, opened 2018-07-01
#306 - Unable to link "Indonesia"

rcffc:
Split query for candidate concept retrieval in two parts - one that matches the label exactly and one using full-text-search.
Apply limit only for the full-text-search part.
Cannot test it out at the moment, since the server is still down.


reckart:
Still WIP?

reckart:
@rcffc the max seems to be set to 400 *seconds* - that seems the pretty generous?!

reckart:
Ok, commented out and server is re-starting. This may take a few hours.

reckart:
Since the PR is still WIP, I expect you don't need a review immediately. Please request one when you're done. I might still comment on the code from time to time. If something needs to be discussed, just add a comment.

rcffc:
I disabled caching for testing, now the only thing left to do is to reactivate it. How should it be done? A CandidateRetrieval query is now both dependent on the typed string and the marked mention. Probably by including the typed string as a key.

reckart:
Hm... if we did two separate queries for exact matching and smart matching it would be easier to do the caching. I suppose we'd only need the caching for the smart part, not for the exact part.

rcffc:
Right.

reckart:
@rcffc What's the status on this branch?

rcffc:
It's done.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/model/CandidateEntity.java
```diff
@@ -172,6 +177,19 @@ public int getLevContext()
         return levContext;
     }
 
+    /**
+     * @param levTypedString edit distance between typed string and candidate entity label
+     */
+    public void setLevTypedString(int levTypedString)
```
Rentier: `aLevTypedString`

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,24 +66,106 @@
...
+
+        String exactMatchingMentionCapitalized = getExactMatchingQueryPart("exactMentionCapitalized");
+
         String query = String.join("\n",
             "DEFINE input:inference 'instances'",
             SPARQL_PREFIX,
             "SELECT DISTINCT ?e2 ?altLabel ?label ?description WHERE",
             "{",
             "  {",
+            exactMatchingTypedString,
```
Rentier: This looks susceptible to SPARQL injection.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,24 +66,106 @@
...
+            "    VALUES ?topic {" + String.join(" ", WIKIMEDIA_INTERNAL,
+                WIKIMEDIA_PROJECT_PAGE, WIKIMEDIA_CATEGORY, WIKIMEDIA_DISAMBIGUATION_PAGE,
+                WIKIMEDIA_LIST_ARTICLE, WIKIMEDIA_TEMPLATE, WIKIMEDIA_NEWS_ARTICLE,
+                WIKIMEDIA_NAVIGATIONAL_TEMPLATE) +
+                "}",
+            "    ?e2 rdf:type ?topic",
+            "  }",
+            "  ?e2 rdfs:label ?label.",
+            "  FILTER ( lang(?label) = \"en\" )",
+            "}");
```
Rentier: Is there no limit here?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,24 +66,106 @@
...
+     *
+     * This query retrieves candidates via exact matching of their labels and full-text-search
+     *
+     * @param aTypedString the unprocessed raw string the user searched for or
+     * @param aMention the words spanned by the mention, pre-processed
+     * @param aLimit maximum number of results
+     * @param aDescriptionIri KB-specific IRI that indicates a description
+     * @return a query to retrieve candidate entities
+     */
+    public static TupleQuery generateCandidateFullTextQuery(RepositoryConnection conn,
```
Rentier: Why is it public?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I am assuming that there are not many objects matching exactly.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I call it from a class in a different package.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I will escape it.

Rentier:
Does not work for me:

image

image

reckart:
These [screenshots](url
https://github.com/inception-project/inception/pull/358#issuecomment-405227253) are funny - it seems that `mention` is actually passed as string to SPARQL instead of taking whatever value the user has entered into the field ;)

rcffc:
Sorry, I did not test it after the last commit. Can you try again please?

Rentier:
Does not work for me. neither *Austrian empire* nor *United states*.

image

rcffc:
screenshot from 2018-07-19 21-55-59

screenshot from 2018-07-19 21-56-25

While I can link United States and at least the Cisleithania for Austrian Empire, Austrian Empire cannot be found. The cause of the problem is that the Wicket Kendo AutoCompleteTextField in ConceptFeatureEditor line 111 automatically gives me the input in lower case, and I don't know how to preserve the original capitalization. It is a problem because exact matching does not work if the capitalization is not correct.

reckart:
Try if you can use `lcase` in the comparison so that the lower-case string you get from the auto-complete is matched against a lower-case value from the KB. User's normally won't want to pay attention to capitalization when searching.

Do we have a unit test for this query that you can test against multiple resources? (cf. https://github.com/inception-project/inception/pull/401)

rcffc:
Like this?

```
SELECT DISTINCT ?e2 ?description WHERE
    {
     VALUES ?labelpredicate {rdfs:label skos:altLabel}
      {
        ?e2 ?labelpredicate ?label .
        FILTER (LCASE(STR(?label)) = "united states of america" @en) .
      }
    }
```

Unfortunately it times out with our Virtuoso.

reckart:
What about the "official" Wikidata endpoint?

Whenever a KB supports full-text search (such as our virtuoso) we should be using that.

naveen2507:
It doesn't work on the "official" Wikidata endpoint as well.

naveen2507:
May be you can try with UNION in SPARQL.  Take the name and put it using in the query through binding. For me this query works: 

SELECT * WHERE {
  {?s ?label "United States of America"@en .}
  UNION
  {?s ?label "united states of america"@en .}
  VALUES ?label {rdfs:label skos:altLabel}
}

reckart:
If you need the original case from the auto-complete, construct it with the `ignoreCase = false` option: https://docs.telerik.com/kendo-ui/api/javascript/ui/autocomplete/configuration/ignorecase

@naveen2507 can you point @rcffc to an example where you passed options to Kendo components?

naveen2507:
@rfcc

https://github.com/inception-project/inception/blob/c958a3df5850a5245c054f9d7dd0a101d0dea22a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/AbstractInfoPanel.java#L168

rcffc:
Thanks @naveen2507.
screenshot from 2018-07-21 12-42-47
screenshot from 2018-07-21 12-42-28

I have split exact matching and full-text matching in two separate methods to add candidates from exact matching to ranking _after_ frequency cutoff, to make sure they don't get lost. Furthermore, I have removed mention pre-processing for full-text matching, I don't think it is necessary for our case, since e.g. a number could also be part of a NE: AK-47.

reckart:
@rcffc If you have not done so already, please add comments to the code explaining why the query was designed the way it was and which other alternatives were tried and why they did not work.

Rentier:
When I try it, it works better than before. There are still some issues:

1) The returned items do not have description
2) Sometimes it seems slow: `It took [10127] ms to rank candidates for mention [United States] and typed string [U]`
3)

```
2018-07-23 09:50:18 ERROR [SYSTEM] ConceptLinkingService - could not get semantic signature
org.eclipse.rdf4j.query.QueryEvaluationException: org.eclipse.rdf4j.query.QueryEvaluationException: com.sun.org.apache.xerces.internal.impl.io.MalformedByteSequenceException: Invalid byte 2 of 3-byte UTF-8 sequence.
	at org.eclipse.rdf4j.http.client.QueueCursor.convert(QueueCursor.java:47) ~[rdf4j-http-client-2.2.4.jar:?]
	at org.eclipse.rdf4j.http.client.QueueCursor.convert(QueueCursor.java:19) ~[rdf4j-http-client-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.QueueIteration.checkException(QueueIteration.java:185) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.QueueIteration.getNextElement(QueueIteration.java:146) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.LookAheadIteration.lookAhead(LookAheadIteration.java:90) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.LookAheadIteration.hasNext(LookAheadIteration.java:52) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.IterationWrapper.hasNext(IterationWrapper.java:67) ~[rdf4j-util-2.2.4.jar:?]
	at de.tudarmstadt.ukp.inception.conceptlinking.service.ConceptLinkingService.getSemanticSignature(ConceptLinkingService.java:436) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.conceptlinking.service.ConceptLinkingService.lambda$rankCandidates$1(ConceptLinkingService.java:351) ~[classes/:?]
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184) [?:1.8.0_171]
	at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382) [?:1.8.0_171]
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481) [?:1.8.0_171]
	at java.util.stream.ForEachOps$ForEachTask.compute(ForEachOps.java:291) [?:1.8.0_171]
	at java.util.concurrent.CountedCompleter.exec(CountedCompleter.java:731) [?:1.8.0_171]
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289) [?:1.8.0_171]
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056) [?:1.8.0_171]
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692) [?:1.8.0_171]
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157) [?:1.8.0_171]
Caused by: org.eclipse.rdf4j.query.QueryEvaluationException: com.sun.org.apache.xerces.internal.impl.io.MalformedByteSequenceException: Invalid byte 2 of 3-byte UTF-8 sequence.
	at org.eclipse.rdf4j.http.client.QueueCursor.convert(QueueCursor.java:47) ~[rdf4j-http-client-2.2.4.jar:?]
	at org.eclipse.rdf4j.http.client.QueueCursor.convert(QueueCursor.java:19) ~[rdf4j-http-client-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.QueueIteration.checkException(QueueIteration.java:185) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.QueueIteration.handleClose(QueueIteration.java:172) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.AbstractCloseableIteration.close(AbstractCloseableIteration.java:51) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.iteration.QueueIteration.checkException(QueueIteration.java:181) ~[rdf4j-util-2.2.4.jar:?]
	... 15 more
Caused by: com.sun.org.apache.xerces.internal.impl.io.MalformedByteSequenceException: Invalid byte 2 of 3-byte UTF-8 sequence.
	at com.sun.org.apache.xerces.internal.impl.io.UTF8Reader.invalidByte(UTF8Reader.java:701) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.io.UTF8Reader.read(UTF8Reader.java:408) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLEntityScanner.load(XMLEntityScanner.java:1895) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLEntityScanner.scanContent(XMLEntityScanner.java:1032) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:2817) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:602) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.next(XMLNSDocumentScannerImpl.java:112) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl.scanDocument(XMLDocumentFragmentScannerImpl.java:505) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:842) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.parsers.XML11Configuration.parse(XML11Configuration.java:771) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.parsers.XMLParser.parse(XMLParser.java:141) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.parse(AbstractSAXParser.java:1213) ~[?:1.8.0_171]
	at com.sun.org.apache.xerces.internal.jaxp.SAXParserImpl$JAXPSAXParser.parse(SAXParserImpl.java:643) ~[?:1.8.0_171]
	at org.eclipse.rdf4j.common.xml.SimpleSAXParser.parse(SimpleSAXParser.java:221) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.common.xml.SimpleSAXParser.parse(SimpleSAXParser.java:196) ~[rdf4j-util-2.2.4.jar:?]
	at org.eclipse.rdf4j.query.resultio.sparqlxml.AbstractSPARQLXMLParser.parseQueryResultInternal(AbstractSPARQLXMLParser.java:211) ~[rdf4j-queryresultio-sparqlxml-2.2.4.jar:?]
	at org.eclipse.rdf4j.query.resultio.sparqlxml.SPARQLResultsXMLParser.parseQueryResult(SPARQLResultsXMLParser.java:74) ~[rdf4j-queryresultio-sparqlxml-2.2.4.jar:?]
	at org.eclipse.rdf4j.query.resultio.helpers.BackgroundTupleResult.run(BackgroundTupleResult.java:98) ~[rdf4j-queryresultio-api-2.2.4.jar:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) ~[?:1.8.0_171]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) ~[?:1.8.0_171]
	at java.lang.Thread.run(Thread.java:748) ~[?:1.8.0_171]```

**Comment by rcffc** [comment]
Some labels could not be retrieved by SPARQL, e.g. 

```
PREFIX e:<http://www.wikidata.org/entity/>
PREFIX rdfs:<http://www.w3.org/2000/01/rdf-schema#>
PREFIX skos:<http://www.w3.org/2004/02/skos/core#>
PREFIX base:<http://www.wikidata.org/ontology#>
SELECT DISTINCT ?label ?p ?e1 WHERE 
  {
    {
      {
        GRAPH <http://wikidata.org/statements>
          { ?e1  ?rd ?m . ?m ?p <http://www.wikidata.org/entity/Q726738> . }
      }
      UNION
      {
        GRAPH <http://wikidata.org/statements>
          { <http://www.wikidata.org/entity/Q726738> ?p ?m . ?m ?rr ?e1 . }
      }
    }
    {
      GRAPH <http://wikidata.org/terms> { ?e1 rdfs:label ?label. }
      FILTER ( lang(?label) = "en" )
    }
  }
 LIMIT 2147483647
```

**Comment by rcffc** [comment]
> http://docs.rdf4j.org/server-workbench-console/  points to https://wiki.apache.org/tomcat/FAQ/CharacterEncoding#Q3:
> 
> How do I change how POST parameters are interpreted?
> 
> POST requests should specify the encoding of the parameters and values they send. Since many clients fail to set an explicit encoding, the default is used (ISO-8859-1). In many cases this is not the preferred interpretation so one can employ a javax.servlet.Filter to set request encodings. Writing such a filter is trivial.
> 
> 6.x, 7.x
> Tomcat already comes with such an example filter. Please take a look at:
> 
> webapps/examples/WEB-INF/classes/filters/SetCharacterEncodingFilter.java

**Comment by reckart** [comment]
> org.apache.catalina.filters.SetCharacterEncodingFilter

That would be a measure the remote side (the server) has to take, not our side (we are the client in this case). If we wanted to influence encoding, we'd probably need to do so somehow through RDF4J.

I found http://openrdf.callimachus.net/sesame/2.7/docs/articles/serverinstall.docbook?view#workbench-utf-8 but I am not sure if this is related to the issue we are observing.

**Comment by reckart** [comment]
When I run the query you posted above on Virtuoso, I see some sensible Unicode results, e.g.

```
"B♯"@en
```

and others like

```
"B������"@en  (which should probably be "B𝄫")
```

It is difficult to tell if these are encoding problems coming from the transferring of data or if they exist already in the KB itself (i.e. if they were caused by importing badly-encoded data into the KB).

My guess would be that the data is already broken in the KB since when I try to force different other encodings in the browser (e.g. Latin 1), it looks even worse. And actually, some of the characters seem to be good Unicode characters (e.g. "♯" while others are broken).

But as I said - it is only a guess.

**Comment by reckart** [comment]
Potentially relevant: https://github.com/openlink/virtuoso-opensource/issues/17

**Comment by reckart** [comment]
@rcffc @Rentier so, I think server-side problems like a badly configured encoding in Virtuoso shouldn't be handled as part of this issue. 

Wouldn't it makes sense to merge this PR and open new issues for remaining problems? After all, you say that it is working better now.

**Comment by Rentier** [comment]
@reckart I would propose that you try it and say when it is ready to merge, as it was you who reported the issue.

**Comment by rcffc** [comment]
Jenkins says 

> Build timed out (after 20 minutes). Marking the build as failed.

Why?

**Comment by rcffc** [comment]
@reckart Can you try it out please?

File: inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java [hunk 5]
```diff
@@ -67,58 +68,146 @@
...
     private static final String WIKIMEDIA_NAVIGATIONAL_TEMPLATE = "e:Q11753321";
+
+    private static String getExactMatchingQueryPart(String aString)
+    {
+        return String.join("\n",
+            "    SELECT DISTINCT ?e2 ?description WHERE",
+            "    {",
+            "     VALUES ?labelpredicate {rdfs:label skos:altLabel}",
+            "      {",
+            "        ?e2 ?labelpredicate ?" + aString + " @en .",
```
reckart: This opens the way to query injection problems. If the query string includes a double quote, the SPARQL query breaks. I see that some escaping is done in other locations - but it would be better to do the escaping where it is needed, namely when constructing the query.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,58 +68,146 @@
...
+    private static String getFullTextMatchingQueryPart(String aString, int aLimit) {
+        return  String.join("\n",
+            "    SELECT DISTINCT ?e2 ?altLabel ?description WHERE",
             "    {",
             "      VALUES ?labelpredicate {rdfs:label skos:altLabel}",
-            "      GRAPH " + TERMS,
             "      {",
             "        ?e2 ?labelpredicate ?altLabel.",
-            "        ?altLabel bif:contains '?entityLabel'. ",
+            "        ?altLabel bif:contains '?" + aString + "'. ",
```
reckart: This opens the way to query injection problems. If the query string includes a double quote, the SPARQL query breaks. I see that some escaping is done in other locations - but it would be better to do the escaping where it is needed, namely when constructing the query.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,58 +68,146 @@
...
+     * @param aLimit maximum number of results
+     * @param aDescriptionIri KB-specific IRI that indicates a description
+     * @return a query to retrieve candidate entities
+     */
+    public static TupleQuery generateCandidateFullTextQuery(RepositoryConnection conn,
+        String aString, int aLimit, IRI aDescriptionIri)
+    {
+        aString = RenderUtils.escape(aString).toLowerCase(Locale.ENGLISH);
+
+        String fullTextMatchingString = getFullTextMatchingQueryPart("string", aLimit);
```
reckart: It seems to me that `"string"` here should instead be the text that is being searched for (i.e. `aString`)? Or am I misunderstanding something?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -67,58 +68,146 @@
...
+     *
+     * @param aString String for which to perform full text search
+     * @param aLimit maximum number of results
+     * @param aDescriptionIri KB-specific IRI that indicates a description
+     * @return a query to retrieve candidate entities
+     */
+    public static TupleQuery generateCandidateFullTextQuery(RepositoryConnection conn,
+        String aString, int aLimit, IRI aDescriptionIri)
+    {
+        aString = RenderUtils.escape(aString).toLowerCase(Locale.ENGLISH);
```
reckart: Create a local variable with a new name instead of assigning a value to the parameter variable. Parameters should not be reassigned.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
reckart: Is it actually necessary to inject the query string in this way? Isn't it possible to create a string literal with a language and bind that to a proper query variable here?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: No, in line 210 "string" is bound to the text that is being searched for.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I inject it here so I don't have to write the query part twice. But you are right about creating the literal with a language.

rcffc: > but it would be better to do the escaping where it is needed, namely when constructing the query.

I have double quotes in the query (`FILTER ( lang(?label) = "en")`). If I escape when constructing the query (`TupleQuery tupleQuery = conn
            .prepareTupleQuery(QueryLanguage.SPARQL, RenderUtils.escape(query)); `), the backslashes are not accepted by the SPARQL processor.

reckart:
@rcffc seems to work reasonably. I believe that it would be nice if the "exact match" could be case-insensitive. E.g. one of the texts I have the Wired magazine is written as `WIRED` but that is not found in the ontology - I have to search for `Wired` instead. Furthermore, if I search for `Wired magazine`, then I don't find it at all even though the description is `monthly American magazine`.

So I think the PR makes improvements, but it feels like we're not there yet. I would suggest to merge it and then to open new issues for case-insensitive search and also for unit tests that check if specific concepts are found when certain search strings are being used.

reckart:
@rcffc btw, is the caching still disabled? Somehow doing entity linking feels very slow.

rcffc:
For me it does not even work anymore, the page lock times out, same as in #475. (0.5.0, commit b48f64941a30700f200628ec6412a2bc0ce83130 and 0.4.3 commit 9ac30ad626e5e4246aa1650d835b5c63ec191af9)

rcffc:
Caching has been reactivated.

reckart:
@rcffc are you sure that you can reach the KB server?

rcffc:
I believe that it would be nice if the "exact match" could be case-insensitive. 

> I believe that it would be expensive because we would need to use FILTER, comparing over all concepts in a knowledge base.

rcffc:
@rcffc are you sure that you can reach the KB server?

> Yes

reckart:
@rcffc Maybe case sensitivity is something that needs to be configured on the KB server (e.g. Virtuoso) such that the data is simply all indexed internally in lower case?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I am calling this query twice, once with the surface form and once with the typed string. I made that part variable so I don't have to write duplicate code.

reckart: Ok, but that doesn't require that you change the query. I'd just use a hard-coded query variable like "?value" and in one case bind the surface form to it and in the other the typed string.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
rcffc: I first had _getFullTextMatchingQueryPart_ twice for typed string and surface form like in exact matching, but now _generateCandidateFullTextQuery_ is just called twice with typed string or surface time respectively.

reckart:
@rcffc Does entity linking work for you on master?

reckart:
@rcffc My trusty entity linking project on blinky seems to work (build b48f64941a30700f200628ec6412a2bc0ce83130). Can you test on blinky?

rcffc:
Do you mean _REC Entity Linking Playground_? Yes it works, but it does not use UKP_WIKIDATA. If I use my own project with UKP_WIKIDATA on blinky it does not load the annotations: 
peek 2018-08-23 13-49

reckart:
@rcffc ok, I see the problem. I have opened a separate issue for it: https://github.com/inception-project/inception/issues/486

Since it doesn't seem to be related to the present PR, I'm merging it now.

Issue #297, opened 2018-06-17
Show IRIs on-mouse-over


reckart:
While the human-readable labels in the UI are much nicer than the IRIs, it is sometimes important for the user to see what the IRI behind a label is. A use-case is if the `label` IRI in the KB settings is not correctly set and the user wants to change it to another property. 

The IRI (identifier) of a property or linked concept/class should be shown in a tooltip on-mouse-over.

This already happens when moving the mouse over the title of a concept/property/instance, but not on the statements.

![2018-06-17_11-10-38](https://user-images.githubusercontent.com/1410238/41506495-2e6c0526-721f-11e8-9c6d-ac2642f6de90.png)

![2018-06-17_10-55-15](https://user-images.githubusercontent.com/1410238/41506486-cfc699b4-721e-11e8-9452-1ec7e9bd70d4.png)

Note that just showing the identifier is not enough. It must also be possible to somehow **select and copy it**, e.g. to then paste it into the KB settings.

reckart:
@naveen2507 the tooltips need a delay before they appear. Right now, if I just swipe the mouse over the screen, many tooltips appear and I need to close all of them manually. This is quite annoying.

naveen2507:
@reckart explored options but didn't find anything to delay the tooltip, i will keep looking on this if I get something reasonable which matches our req. @reckart WDYS?

reckart:
https://docs.telerik.com/kendo-ui/api/javascript/ui/tooltip/configuration/showafter

naveen2507:
@reckart  I already tried that, its for delay before showing the tooltip initially. It takes a parameter for delay in milliseconds before the Tooltip is shown

reckart:
I guess that means that the mouse has to hover for X milliseconds over the element before the tooltip is shown - wouldn't that solve our problem? I mean there is a minimal time required for the mouse to hover before the TT is shown, then a simple swipe shouldn't open many tooltips, no?

naveen2507:
@reckart But that wouldn't enable the user to copy the text from the TT.

reckart:
@naveen2507 Why not? The user could over with the mouse over the element for a while, then the TT appears, then the user can copy the text and then close the TT again using the close button. Does it not work in this way?

naveen2507:
@reckart When we use `autoHide` in combination with `showafter`, we need to wait for the tooltip to show itself. I will commit in the code and you can have a look if its okay.  PR #362

naveen2507:
[...]

PR #362, opened 2018-07-02
#297 Show IRIs on-mouse-over

naveen2507:
- Putting in combination of autoHide and showAfter config.


reckart:
Tested it, but it doesn't work as desired. The tooltip appears with a delay: yes - however, it does not matter whether the mouse actually remains hovering on the element during that time. So if one just quickly moves the mouse over the element, the TT still appears after a moment.

Another idea: how about adding some font-awesome symbol after the property name, e.g. "fa-info-circle" and add the TT to that element with `showOn: "click"`?

naveen2507:
@reckart that can be a better option. I will work on this approach .

Issue #346, opened 2018-06-27
Error when making annotations (com.mysql.jdbc.MysqlDataTruncation)


Rentier:
```2018-06-26 19:57:08 WARN [annotator.redacted] SqlExceptionHelper - SQL Error: 1406, SQLState: 22001
2018-06-26 19:57:08 ERROR [annotator.redacted] SqlExceptionHelper - Data truncation: Data too long for column 'tokenText' at row 1
2018-06-26 19:57:08 ERROR [annotator.redacted] BratAnnotationEditor - Error: org.hibernate.exception.DataException: could not execute statement
javax.persistence.PersistenceException: org.hibernate.exception.DataException: could not execute statement
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:149) ~[hibernate-core-5.2.13.Final.jar!/:5.2.13.Final]
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:157) ~[hibernate-core-5.2.13.Final.jar!/:5.2.13.Final]
	at org.hibernate.internal.ExceptionConverterImpl.convert(ExceptionConverterImpl.java:164) ~[hibernate-core-5.2.13.Final.jar!/:5.2.13.Final]
	at org.hibernate.internal.SessionImpl.firePersist(SessionImpl.java:789) ~[hibernate-core-5.2.13.Final.jar!/:5.2.13.Final]
	at org.hibernate.internal.SessionImpl.persist(SessionImpl.java:767) ~[hibernate-core-5.2.13.Final.jar!/:5.2.13.Final]
	at sun.reflect.GeneratedMethodAccessor434.invoke(Unknown Source) ~[?:?]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:1.8.0_131]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[?:1.8.0_131]
	at org.springframework.orm.jpa.ExtendedEntityManagerCreator$ExtendedEntityManagerInvocationHandler.invoke(ExtendedEntityManagerCreator.java:350) ~[spring-orm-5.0.4.RELEASE.jar!/:5.0.4.RELEASE]
	at com.sun.proxy.$Proxy131.persist(Unknown Source) ~[?:?]
	at sun.reflect.GeneratedMethodAccessor434.invoke(Unknown Source) ~[?:?]
[...]

PR #364, opened 2018-07-02
#346 - Truncate tokenText in LearningRecord

Rentier:
I truncate the tokenText to 255 characters when setting it.


reckart:
@Rentier checkstyle violation reported by Jenkins

Issue #377, opened 2018-07-03
Disable "read only" checkbox in KB settings instead of hiding it


reckart:
**Is your feature request related to a problem? Please describe.**
The "read only" checkbox in the KB settings is hidden for remote KBs because the must be read-only. However, this can lead to hard-to-track problems such as https://github.com/inception-project/inception/issues/367.

**Describe the solution you'd like**
Instead of hiding the checkbox, just disable it so that the user can still see the state of the flag.

PR #379, opened 2018-07-04
#377 - disable "read-only"-checkbox in kb settings instead of hiding it

MdeBoer95:
- add disabled "read-only"-checkbox for remote KBs in KB settings


naveen2507:
image

Issue #353, opened 2018-06-29
Create Instance throws Null Pointer Exception


naveen2507:
**Describe the bug**
Create instance in KB UI throws NPE for `Objects.requireNonNull(language);` in SimpleLiteral.java

```2018-06-29 15:15:55 ERROR [admin] KnowledgeBasePage - Error: null
java.lang.NullPointerException: null
	at java.util.Objects.requireNonNull(Objects.java:203) ~[?:1.8.0_161]
	at org.eclipse.rdf4j.model.impl.SimpleLiteral.setLanguage(SimpleLiteral.java:121) ~[rdf4j-model-2.2.4.jar:?]
	at org.eclipse.rdf4j.model.impl.SimpleLiteral.<init>(SimpleLiteral.java:85) ~[rdf4j-model-2.2.4.jar:?]
	at org.eclipse.rdf4j.sail.nativerdf.model.NativeLiteral.<init>(NativeLiteral.java:53) ~[rdf4j-sail-nativerdf-2.2.4.jar:?]
	at org.eclipse.rdf4j.sail.nativerdf.model.NativeLiteral.<init>(NativeLiteral.java:49) ~[rdf4j-sail-nativerdf-2.2.4.jar:?]
	at org.eclipse.rdf4j.sail.nativerdf.ValueStore.createLiteral(ValueStore.java:720) ~[rdf4j-sail-nativerdf-2.2.4.jar:?]
	at org.eclipse.rdf4j.sail.nativerdf.ValueStore.createLiteral(ValueStore.java:42) ~[rdf4j-sail-nativerdf-2.2.4.jar:?]
	at de.tudarmstadt.ukp.inception.kb.graph.KBInstance.write(KBInstance.java:135) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceImpl.lambda$6(KnowledgeBaseServiceImpl.java:443) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceImpl.update(KnowledgeBaseServiceImpl.java:597) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceImpl.createInstance(KnowledgeBaseServiceImpl.java:440) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceImpl$$FastClassBySpringCGLIB$$77259834.invoke(<generated>) ~[classes/:?]
[...]

PR #384, opened 2018-07-05
#353 - Create Instance throws Null Pointer Exception

matchlesswei:
If there is no language field, just create literal using name.


reckart:
@naveen2507 @rcffc @MdeBoer95 Why do you approve the change if Jenkins fails with many tests?

matchlesswei:
My fault... Now it should pass all the tests.

PR #388, opened 2018-07-06
Bugfix/369 al suggests invisiable annotations

rcffc:
**How to test**
* Create a project and add documents and recommenders to it (e.g. POS or NER)
* Go to annotation screen
* Open the AL sidebar and start an AL session
* Start annotating stuff
* When the AL sidebar starts detecting suggestions work with it
* Mix interactions with the AL sidebar with interactions in the main editor panel, e.g. 
  * manually create an annotation in the main panel if the recommender gets the boundaries wrong, 
  * accept/reject also via the main panel, 
  * correct suggestions via the feature editor in the sidebar, 
  * skip some suggestions as well, try removing history items
  * whatever else you come up with

**To do**
- [x] don't log info about suggestion having been offered to learning history, log it to the event log


reckart:
Since the PR is still WIP, I expect you don't need a review immediately. Please request one when you're done. I might still comment on the code from time to time. If something needs to be discussed, just add a comment.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/AnnotationObject.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Objects;
 
 public class AnnotationObject
-    implements Serializable
+    implements Serializable, Comparable<AnnotationObject>
```
Rentier: I would not implement Comparable, but create a `Comparator `where you need it, as this `compareTo` does one of many things you might want to compare.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningRecommender.java
```diff
@@ -77,7 +77,7 @@ public RecommendationDifference generateRecommendationWithLowestDifference(
         // remove rejected recommendations
         removeRejectedOrSkippedAnnotations(aRecordService, true, learnSkippedRecommendationTime);
 
-        return calculateDifferencesAndReturnLowestDifference(listOfRecommendationsForEachToken);
+        return calculateDifferencesAndReturnLowestVisibleDifference(listOfRecommendationsForEachToken);
```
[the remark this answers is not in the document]
reckart: Sounds like the method has a side-effect and a return value. I'd recommend to separate it into two methods in this case: one with a side-effect and no return value and one with a return value.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -44,6 +46,12 @@
 {
     private final DocumentService documentService;
     private final RecommendationService recommendationService;
+    private List<AnnotationObject> recommendations;
+    private List<List<AnnotationObject>> listOfRecommendationsForEachToken;
+    private AnnotatorState annotatorState;
+    private AnnotationLayer selectedLayer;
```
reckart: This state must not be in the service since the service is shared between all users. It must be in the user's state.
```
     private List<AnnotationObject> recommendations;
     private List<List<AnnotationObject>> listOfRecommendationsForEachToken;
     private AnnotatorState annotatorState;
     private AnnotationLayer selectedLayer;
```

matchlesswei: These fields are not used. I forgot to remove them. Now, these fields are not there.

rcffc:
No progress.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/AnnotationObject.java
reckart: Removed the Comparable interface from the class.

reckart:
@mbugert Would you like to break something? ;)

reckart:
The refactorings in this PR break the "Entity linking example":

```
java.lang.IllegalArgumentException: All suggestions in a group must be for the same feature: expected [value] but got [identifier]
     at org.apache.commons.lang3.Validate.isTrue(Validate.java:158)
     at de.tudarmstadt.ukp.inception.recommendation.api.model.SuggestionGroup.add(SuggestionGroup.java:258)
     at java.util.stream.Collectors.lambda$groupingBy$45(Collectors.java:909)
     at java.util.stream.ReduceOps$3ReducingSink.accept(ReduceOps.java:169)
     at java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1382)
     at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
     at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
     at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
     at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
     at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:499)
     at de.tudarmstadt.ukp.inception.recommendation.api.model.SuggestionGroup.group(SuggestionGroup.java:302)
     at de.tudarmstadt.ukp.inception.recommendation.api.model.SuggestionDocumentGroup.<init>(SuggestionDocumentGroup.java:48)
     at de.tudarmstadt.ukp.inception.recommendation.api.model.Predictions.getPredictions(Predictions.java:108)
     at de.tudarmstadt.ukp.inception.recommendation.render.RecommendationSpanRenderer.render(RecommendationSpanRenderer.java:103)
     at de.tudarmstadt.ukp.inception.recommendation.render.RecommendationRenderer.render(RecommendationRenderer.java:77)
     at de.tudarmstadt.ukp.inception.recommendation.RecommendationEditorExtension.render(RecommendationEditorExtension.java:261)
     at de.tudarmstadt.ukp.clarin.webanno.api.annotation.AnnotationEditorExtensionRegistryImpl.fireRender(AnnotationEditorExtensionRegistryImpl.java:116)
```

reckart:
Ok. Latest push fixes that again :)

Rentier:
In this image, I do not see the suggested token in the annotation view:

image

reckart:
@Rentier Ok, but why? ;) What happens if you click on the text in the sidebar? Does it take you to the right location and highlight the text? Is the suggestion hidden/invisible or did the editor just not jump to the right location.

Rentier:
@reckart I reopened the project and AL session. It starts with the right annotation highlighted. I assume that the AL just did not jump.

PR #393, opened 2018-07-09
Bug/391 unable to index non string primitive features

naveen2507:
- Use feasureSupport.renderFeatureValue()


reckart:
@naveen2507 Check the commits in this PR: it is based on master, not on 0.4.x. We need a PR based on a branch that was created from 0.4.x so we can merge this as a bugfix into 0.4.x.

Issue #391, opened 2018-07-06
Unable to index non-string primitive features


reckart:
**Describe the bug**
None-string fields do not get indexed.

```
2018-07-06 18:39:03 ERROR [simon.eichler] MtasUimaParser - Unable to index document
java.lang.ClassCastException: java.lang.Boolean cannot be cast to java.lang.String
        at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.indexFeatureValue(MtasUimaParser.java:247) ~[inception-search-mtas-0.4.0.jar!/:?]
        at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.createTokenCollection(MtasUimaParser.java:220) ~[inception-search-mtas-0.4.0.jar!/:?]
        at mtas.analysis.MtasTokenizer.constructTokenCollection(MtasTokenizer.java:238) ~[mtas-7.1.0.2.jar!/:?]
        at mtas.analysis.MtasTokenizer.reset(MtasTokenizer.java:171) ~[mtas-7.1.0.2.jar!/:?]
```

**To Reproduce**
1. Create a layer with a boolean feature
2. Create an annotation from that layer

**Expected behavior**

**Screenshots**
Non-string values should be indexed properly, e.g. boolean values should be indexed as "true" or "false".

**Please complete the following information:**
 - Version and build ID: INCEpTION 𝜶 -- 0.4.0 (2018-07-06 17:11:43, build aca896f929f83bd8e6aff7339398315b1db67568)

naveen2507:
@reckart exception from `return (T) (Boolean) aFS.getBooleanValue(feature);` in function `WebAnnoCasUtil.getFeature(FeatureStructure aFS, String aFeatureName)`

PR #396, opened 2018-07-10
#391 Unable to index non-string primitive features

naveen2507:
- Use feasureSupport.renderFeatureValue()

### DIFF
--- a/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasUimaParser.java
+++ b/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasUimaParser.java
@@ -244,9 +244,7 @@ public class MtasUimaParser
     {
         int mtasId = aMtasId;
         FeatureSupport<?> featSup = featureSupportRegistry.getFeatureSupport(aFeature);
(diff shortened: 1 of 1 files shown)


inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasUimaParser.java
```diff
@@ -243,8 +243,9 @@ private int indexFeatureValue(MtasTokenCollection aTokenCollection, AnnotationFe
             String annotationUiName)
     {
         int mtasId = aMtasId;
-        
-        String featureValue = WebAnnoCasUtil.getFeature(aAnnotation, aFeature.getName());
+        FeatureSupport<?> featSup = featureSupportRegistry.getFeatureSupport(aFeature);
```
reckart: Mind that for concept features previously `WebAnnoCasUtil.getFeature(aAnnotation, aFeature.getName());` was indexing the URI at this point. If we switch to the UI value, then `indexConcept()` should probably be changed to index the URI, otherwise we do not index the URI anywhere meaning that URI-based search for concepts would no longer work. This would break the mention view on the KB page (which should be searching URI-based).

Issue #398, opened 2018-07-10
Refactor recommender to "CAS in - CAS out"


Rentier:
**Describe the refactoring action**
Refactor the recommender so they use `CAS` both as input and output.

**Expected benefit**
It will be easier to test, cleaner, looser coupled and will allow us to implement recommender that were not possible before, e.g. relation recommender.

**Things to do in PR #399**

- [x] Adjust Prediction and selection tasks (@Rentier PR #399)
- [x] Port and test DL4J (@reckart PR #399)
- [x] Rewrite, port and test String based recommender (@reckart PR #399)
- [x] Port and test OpenNLP POS/NER recommenders (@Rentier, @reckart PR #399)
- [x] Port and test External recommender (@Rentier,  PR #399) 
- [x] Port and test Named entity linker (@rcffc, PR #399) 
- [x] Reintroduce traits

**Things to do in follow-up PR**

- [ ] Refactor annotation object
- [ ] Port active learning
- [ ] Parameterized recommender smoke test
- [ ] Port renderer (with lots of unit tests)
- [ ] JavaDoc
- [ ] Developer documentation
- [ ] Tutorial to implement simple recommender
- [ ] Adhere to max suggestions

reckart:
I believe it is not necessary to use concurrent collections in the `RecommenderContext` because it once it has been created, it should only be shared read-only.

reckart:
I have added a JCas class called `PredictedSpan` to the recommender API module.

reckart:
I have reactivated the DL4J module again and ported it to the new API. It will still require some work to be really usable, but in general it seems to work ok with the new API.

Originally, I wanted to keep the data I keep in memory small by processing CASes one-by-one. However, I ended up extract all the training data from all CASes before starting the training process because otherwise, I couldn't properly split it into training and test sets.

The unit tests you did for the OpenNlpNer could be ported to the DL4J recommender almost copy/paste. That was also pretty cool. Maybe we can later set up some parametrized test suite which can be applied to multiple recommenders instead of maintaining per-recommender tests.

reckart:
@rcffc @Rentier I have organized the todo list here at the top. Has anybody already started working on re-introducing traits and/or porting the OpenNLP recommenders?

Rentier:
I thought that POS (you) and NER (me) is already done.

reckart:
@Rentier I thought that as well, but the checkbox wasn't ticked yet - ok, I ticked it and put our names there.

[...]

PR #399, opened 2018-07-10
#398 - Recommender use CAS for input and output

reckart:
- Define the first set of new interfaces


reckart:
@Rentier I have taken the liberty to create a PR.

Rentier:
I experienced some issues with our design when I started working on the OpenNlpNer implementation:
1) Evaluating and predicting needs a model. One could either split these into trainer and evaluator classes and pass it as a constructor argument or have one class like now and pass them in the call.
2) The model needs to be saved somehow
3) Configuration is needed, e.g. language parameter for NER. For the configuration, I would just use traits, not a different configuration class combined with traits

reckart:
Wrt. 1) How about we tell the trainer/predictor some context ID (e.g. a string consisting of the project and user IDs) under which it should internally store the model - similar to what we discussed today for the external recommenders. When predicting, the ID must be provided as well.

Wrt.2 ) Not necessarily. The model could also remain in memory. If the model needs to be saved, the recommender could handle that internally

Wrt. 3) I believe the language code is only informational and could be anything. If that is the case, we could just pass some dummy value.

reckart:
A string ID may not be sufficient. We could create some session class into which the recommender can store the model and/or other state if necessary. The external recommenders would not store anything there, but the internal ones could.

Rentier:
@reckart I finished the implementation of the first new recommender. Can you have a look at it whether we want to pursue this road?

reckart:
@Rentier what is the purpose of the "counts" in the PercentageBasedSplitter?

Rentier:
@reckart It counts how often an annotation was already added. Based on this count, we decide whether data is training or test data. I was not sure whether we want to support splitting datasets that have annotations of several kinds. Right now, `counts ` could also be just an int.

reckart:
> It counts how often an annotation was already added.

It counts how often an annotation of a particular class has been added. I cannot imagine a scenario where this would be relevant. I'd suggest to remove this and add it back if we really have a need for it.

reckart:
The present code looks reasonable to me. I wonder how to best transition to it...

Would it make sense to implement an adapter between the new API and the old "AnnotationObject" API which would then gradually be removed?

reckart:
The formatting in general looks a bit odd - not DKPro style.

Rentier:
I would change prediction and selection tasks to use the API. I do not know how much we have to rewrite, but I guess

1) Prediction and selection tasks
2) OpenNLP POS
3) String based recommender
4) External recommender
5) Named entity linker
6) Active learning
7) Renderer

I do not see too much issues with 2-4. 1 should be also doable. 5 I do not know, 6 and 7 need an updated annotation object. I will look how simple using the new API and converting internally to the old API would be.

dl4j and miralium is another issue, as they are deactivated anyways.

reckart:
Since the `evaluate` method internally trains and predicts, I think we could do without the `context` parameter.

Rentier:
We could give all annotations to the splitter and then split internally. Then there are two methods `getTrainingData` and `getTestData`.

reckart:
@Rentier We could - but that would enforce that we actually have all the annotations materialized. Right now, we only need to know their total number.

Rentier:
In which cases do we know how many annotations we have but do not have them materialized? Setter in an interface somehow does not feel right.

reckart:
A `next()` method in the IncrementalSplitter which simply returns `this` also doesn't feel right. How about turning the strategy into a factory with a `getInstance(total)` method or something like that? That way the strategy would also become stateless.

reckart:
Well, I guess the incremental strategy would still be stateful.

Rentier:
@reckart  You could implement `Iterable` instead of `Iterator` and create an inner class for the iterator itself. I personally find that also cleaner.

I really like the `getTrainingData` / `getTestData` approach the most, but if you say we sometimes do not have the annotations (when?) then I would go with the factory/builder.

reckart:
My first try in implementing the DL4J recommender, I implemented it such that I didn't have to keep the entire set of samples in memory. I iterated over the CASes twice: once to determine the number of sentences and then to extract batches of samples. So I never had more than one batch in memory. However, it was difficult to implement it, so eventually I switched to extracting all the samples and then keeping them in memory. The question is maybe: how much data do we expect to process. I'd personally tend to try making as few assumptions as necessary regarding the availability of the data and try to keep options for a streaming-learning approach open. But that's just my opinion. It's certainly easier to assume the data is all there.

reckart:
Actually, I think we can get away without knowing the total at all - we just need to implement a sampling approach in the splitter. E.g. we take the first *n* samples and assign them to the train set and the next *m* samples and assign them to the test set and then we repeat the procedure. In that way, we do not need to know the total beforehand. In order to allow evaluation to run even with almost no data, we could assign every second sample of the first 10 samples to the train/test sets and only then proceed with sampling.

Rentier:
We also could modulo it into two groups.

reckart:
I'm trying something out ;)

reckart:
@Rentier Ok, wir haben kein `setTotal`mehr.

reckart:
I think the `OpenNlpNerRecommenderTraits` should have explicit fields for the parameters that we want to be able to configure instead of embedding a `TrainingParameters`.

reckart:
@Rentier how about making the return value of evaluate a more complex object. Maybe a map with some well-known keys such as "precision", "recall", "accuracy", "f1", etc. depending on which measures the recommender supports?

reckart:
Trying to have a stab at porting the string matching recommender now.

Rentier:
I would be interested in supporting multiple evaluation metrics, but right now I just would use a double for our minimum viable refactoring.

Rentier:
I also would prefer a builder instead of a map, e.g. 

`Metrics.builder().withF1(f1).withAccuracy(accuracy).build();`

Rentier:
`Build timed out (after 20 minutes). Marking the build as failed.`

reckart:
The DL4J tests are running long. Also the incremental tests are running long.

The DL4J code internally has `limit` variables to restrict the training data taken into consideration but they can currently not be set from the test. I would propose that we introduce a way to limit recommenders to a specific number of samples and set that value to something reasonably low in the unit tests.

Rentier:
We could add a limit to the contexts and adhere to that, also we can add limiting splitters.

reckart:
The limit should be added in such a way that the recommenders can stop early processing the data. So e.g. adding it to the splitter which tends to be applied only after all the samples have been generated (at the moment) doesn't seem appropriate.

I find it a bit strange to add parameters to the contexts.

How about storing the limit either in the recommender configuration itself or in the traits - or just have a setter on the recommender since this setting is usually only required during testing (at least at the moment).

reckart:
Mind for evaluation, we might consider limiting in general to bound the time required to perform the evaluation - even in production use.

Rentier:
I would save the limit then in the `Recommender`, as it  will apply most likely to all recommender algorithms.

reckart:
Mind that settings in `Recommender` are usually user-specified settings and they are usually global for all instances of the corresponding recommendation engine (e.g. during training as well as during evaluation). However, we probably do not want to limit during training - only during evaluation.

Rentier:
I tried to run the dl4j tests on my workstation and it is extremely slow just for the pos training test (~15min). I do not know whether this recommender will be useful in the future. Where would you add the limit? We could also add it as an argument to the `evalaute` function.

Rentier:
I am currently working on the prediction task, therefore refactoring the `RecommenderService`. Do we need to make it backwards compatible? Problem is that the IDs are wrong now, e.g. 

   `public static final String ID =       "de.tudarmstadt.ukp.inception.recommendation.imls.opennlp.ner.OpenNlpNerClassificationTool";`

I also want to rename the `tool` field in `Recommender` to `engine` or so.

reckart:
Originally, I used the class names as IDs, but then introduced constants to allow refactoring. If we can for the moment stay backwards compatible and use the same IDs, then I think we should do that.

As for fields in the recommender: you could just rename them on the Java level and keep the DB column name - at least for the moment. We can always add a Liquibase changeset later.

Adding the limit as a parameter to the evaluate function seems reasonable.

reckart:
... but it means that we can only limit in evaluate then, and not also during training/prediction (to keep test runtimes small).

reckart:
@Rentier I have added a new StringMatchingRecommender - it uses KrippendorffAlphaUnitizingAgreement from DKPro Statistics to for evaluation. I'm not really sure if I am using it in the right way - probably need to ask @chmeyer. The measure doesn't seem to produce a score between 0 and 1 - rather some kind of distance which can grow pretty large and seems to reduce the more the annotators agree with each other. I didn't really look deep into it.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -62,66 +73,88 @@ public void run()
     {
         User user = getUser();
 
-        Predictions model = new Predictions(getProject(), getUser()); 
+        Project project = getProject();
+        Predictions model = new Predictions(project, getUser());
+        List<AnnotationDocument> documents = documentService.listAnnotationDocuments(project, user);
```
reckart: @Rentier `listAnnotationDocuments` is not adequate because it only returns annotation documents that were opened by the user at least once. Hence we are not guaranteed to get predictions for all documents here. It is necessary to iterate over the source documents and then obtain the annotation CAS for each source document for the given user. See https://github.com/webanno/webanno/pull/1025

The same probably goes for the evaluation task.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -62,66 +73,88 @@ public void run()
...
 
-        for (AnnotationLayer layer : annoService.listAnnotationLayer(getProject())) {
-            if (!layer.isEnabled()) {
+        for (AnnotationDocument document : documents) {
+            JCas jCas;
+            try {
+                jCas = documentService.readAnnotationCas(document);
```
reckart: The CAS we obtain at this point may not have the correct type system. E.g. if I open a legacy project that was created before the introduction of the `PredictedSpan`, then it does not contain the type definition for that type and the prediction task fails. Hence, it is necessary to "upgrade" the CAS here to the latest type system by calling `AnnotationSchemaServiceImpl.upgradeCas(CAS, SourceDocument, String)`.

The same probably goes for the evaluation task.

Rentier:
Do we delete the old two string based recommender? I would assume that we then break stuff. To circumvent this, we could implement two additional factories which use the id of POS and NER string based recommender respectively to make it compatible.

reckart:
I think we might be generally deprecating recommenders from time to time. So how about adding a flag to the factory which indicates whether the user can select it when creating a new recommender? That would allow us to keep old (deprecated) recommender engines around but preventing users from creating recommenders using them. 

For the String-matching case, we could deprecate e.g. the "POS string matching" recommender engine and rename the "NER string matching" recommender to simply to "String matching recommender" since it covers both cases now.

Rentier:
@reckart The string recommender does not work for me, it looks like it produces predicted spans that do not cover tokens for me (`PredictionTask::extractAnntations`) throws.

I added a deprecated flag to factories and deprecated the two older string matching recommender. We need to work on the visuals later, e.g. showing a text stating that it is deprecated and fixing the tool choices for these recommenders.

reckart:
@Rentier There is todo comment still in the code to ensure that the end boundaries of predicted spans correspond to token boundaries. Do you want to have a look at it?

Rentier:
@reckart It would be nice if you could tackle that, as you are more experienced with the code around. I am working right now on the external recommender rewrite and then want to rewrite the OpenLP pos.

For the external recommender, do we assume that we are given `PredictedSpan` or the target annotation (e.g. `webanno.custom.EDAktivitt`) and convert it internally?

reckart:
@Rentier What do you see as the pro's/con's towards either of the two approaches?

Rentier:
If we just use `PredictedSpan`, then we can load the remote annotated CAS in the annotated local CAS and do not need to copy. Problem is that the typesystem would need to include this annotation, we would need to change neuralweb and might lock us in as we leak this annotation outside.

reckart:
Our code doesn't strictly require that there is an annotation type called "predicted span". What we need is an annotation type with a label feature and score feature. We could look up that annotation instead of the default `PredictedSpan` when extracting the data we need. I.e. we could have the annotation type, name and label feature configurable in the traits. Would that help us in any way? E.g. would it remove the need to copy anything?

Rentier:
We can do that, then we would need to change the `PredictionTask` (maybe also `SelectionTask`) to not look for the `PredictedSpan` annotation, but for the one the recommender configuration tells us.

reckart:
@Rentier and would you think it is a good option? Do you see other alternatives that provide specific benefits?

Rentier:
It would be a bit difficult to implement, as we would store the feature name in the traits which are heterogeneous, e..g only the external recommender has this. I do not see exactly how the recommender tasks would have access to these without `instanceof` except when we add this to the factory/recommendation engine interface which has access to the traits.

reckart:
There settings could be there for all recommenders (i.e. they do not reside in the traits but in the recommender entity directly) but they could be made exclusively editable for the external recommender through the traits UI - I have done such a thing e.g. for the "tagset" of string features - a setting which historically exists in the feature entity, but is not only configurable through the feature traits UI.

So better make the `PredictedSpan` a part of the external recommender protocol then, i.e. expect that the external recommender produces exactly this annotation?
 
I believe that user creating the custom type on which the recommender is applied would normally only create the label feature, but not create a score feature since the latter would be considered a technical detail of the recommender system, not part of the annotation schema.

Thus, it we do neither make the `PredictedSpan` a part of the external recommender protocol nor make the type/label-feature/score-feature configurable, how would we then get the score back from the external recommender?

Rentier:
We can make it configurable by adding a method `RecommendationEngine::getPredictionType -> String` which defaults to `PredictedSpan`. For the external recommender, we can offer a `ExternalPredictedSpan` with label and score. The predicted annotation does not need to be the same as the layer + feature annotation the recommender predicts, so we can entangle that. Right now, e.g. neuralweb uses the same annotation for both INCEpTION and predicting.

reckart:
What is the difference between `PredictedSpan` and `ExternalPredictedSpan`?

reckart:
If we introduce a way of configuring the layer/features as mentioned in 
https://github.com/inception-project/inception/pull/399#issuecomment-409911218 then the neuralweb could continue to predict the same type it trains on - no modifications would be required. We also wouldn't need to introduce a new type.

Rentier:
Then neuralweb could not use active learning, because we are not given confidences per annotation. Is it possible to add features on the fly, e.g. using `MyNamedEntity` which in the INCEpTION feature editor just has a label  to have also a score, but just between the web service and the external recommender class in INCEpTION?

reckart:
@Rentier The UIMA framework doesn't permit to change the type system of a CAS object one it has been initialized. However, we obviously have to do that regularly, e.g. if a user adds new layers or features. That is what the `AnnotationService.upgradeCas()` does - internally, it serializes the CAS data to a specific format, then creates a new CAS object with the new type system and then deserializes the original data into that new CAS object. So in principle, we can add new types/features. So in theory, if we have a type `X` with a label `X.l`, we could auto-generate a `X.l_score`. I think it would be possible to confine such a feature to the recommendation service, i.e. it wouldn't show up in the project settings and wouldn't be stored along with the annotation CAS files either.

chmeyer:
>  I have added a new StringMatchingRecommender - it uses KrippendorffAlphaUnitizingAgreement from DKPro Statistics to for evaluation. I'm not really sure if I am using it in the right way - probably need to ask @chmeyer. The measure doesn't seem to produce a score between 0 and 1 - rather some kind of distance which can grow pretty large and seems to reduce the more the annotators agree with each other. I didn't really look deep into it.

@reckart: I looked at `StringMatchingRecommender`in commit 25de2e3b2ec5ec6dc5a690019edaf557d50eca7b. Problematic is your way of filling up the continuum: As I understand it, you create a near-unlimited continuum and then start each document's annotations at an point without filling up all spaces. AlphaU's underlying null model, however, assumes that also the gaps are of correct lengths. That is to say: offset 0 of doc2 should be length(doc1), and correspondingly the size of the continuum should be sum_i length(doc_i) instead of Long.MAX_VALUE. If you deviate from this by allowing larger gaps (e.g., offset 0 of doc2 is >> length(doc1)) then the expected agreement will be higher.

I experimented with exactly your setting and I guess that the method runs into some numeric instability if using such a large continuum size (i.e., Long.MAX_VALUE). This explains, why the methods returns such strange values! AlphaU may only be in the range [-1, +1], any other range is wrong. If I find some time and get some cold drinks, I can fix the instability. But this won't solve the whole problem.

reckart:
@chmeyer Thanks for the feedback! Is it necessary for the implementation to know the continuum size when the study is created or would be be possible to add a setter for the continuum size to the study?

chmeyer:
@reckart Currently there's no API function for that, but it is save to start with a large continuum and then make it smaller. The quick and dirty solution would be subclass the UnitizingStudy and add the setContinuumLength function to make it shorter when all units have been added. Ultimately, this can be included in DKPro Agreement.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -32,12 +37,18 @@
...
-import de.tudarmstadt.ukp.inception.recommendation.api.Classifier;
+import de.tudarmstadt.ukp.dkpro.core.api.metadata.type.DocumentMetaData;
+import de.tudarmstadt.ukp.dkpro.core.api.segmentation.type.Token;
 import de.tudarmstadt.ukp.inception.recommendation.api.RecommendationService;
 import de.tudarmstadt.ukp.inception.recommendation.api.model.AnnotationObject;
+import de.tudarmstadt.ukp.inception.recommendation.api.model.Offset;
 import de.tudarmstadt.ukp.inception.recommendation.api.model.Predictions;
 import de.tudarmstadt.ukp.inception.recommendation.api.model.Recommender;
+import de.tudarmstadt.ukp.inception.recommendation.api.model.TokenObject;
+import de.tudarmstadt.ukp.inception.recommendation.api.type.PredictedSpan;
```
rcffc: This file does not seem to exist for me.

Rentier:
You need to mark the generated folder (I think it is in recommender-api target and then jcas) as an Intellij generated source folder.

rcffc:
I tried, but even setting the whole target folder to "generated source root" did not help.

Rentier:
You have to set the right folder, else the class package is wrong. Then you also can try `mvn clean install` to generate it.

Rentier:
I rewrote the external recommender and added training to it. I am a bit unsure about the request, I assume we can just have a JSON object which contains the recommender and the type system and a list of casses. Right now, I reused the prediction request, therefore the project information is duplicated.

I also added a `CasAssert` to `inception-imls-external` to make nicer asserts like 

        assertThat(cas).as("Predictions are correct")
            .containsNamedEntity("Ecce homo", "OTH")
            .containsNamedEntity("The Lindsey School Lindsey School & Community Arts College", "ORG")
            .containsNamedEntity("Lido delle Nazioni", "LOC");

If we add a test module, then we can move it there.

rcffc:
> You have to set the right folder, else the class package is wrong. Then you also can try mvn clean install to generate it. 

Thanks, it worked.

reckart:
@chmeyer Should the continuum be defined in terms of characters of in terms of tokens? If the range is `[-1,1]`, what do 0, 1 and -1 respectively mean?

chmeyer:
The continuum should be on the same scale as the offsets and lengths. Typically, it ranges from 0 to max_{u \in units} (offset_u + length_u), but it could be longer if there is only non-annotated data at the end. For token annotations in multiple documents, the continuum would be 0 to sum_d num-tokens(d).

For the alpha score, values < 0 indicate systematic disagreement. If annotator 1 would mark all tokens with even index as category A and all tokens with odd index with category B and if annotator 2 would do the same with exchanged categories then this would be a case of systematic disagreement (most likely, one of the two just mixed up the labels). A score of near 0 indicates randomness (i.e., the annotators act similar to a random number generator with the same distribution as the raters.  A score near 1 indicates good agreement.

reckart:
@Rentier I wonder... how about merging this PR to master?

Rentier:
The tests are failing and running really slow, so I would fix that first.

reckart:
> The tests are failing and running really slow, so I would fix that first.

@Rentier Jenkins says that there are no failing tests. And I have limited the runtime of the really slow test (i.e. the incremental training tests).

So... what do you think now about merging?

Rentier:
There are still some points open, but the biggest thing lacking is the external recommender I think for two reasons: Traits are missing and when we merge this, the neuralweb part of the external recommender will be broken as we changed the request format (lower case names) and added basic training.

reckart:
> Traits are missing 

What do you mean regarding the traits? Where are they missing?

Rentier:
From what I remember, I did not add trait support to the new recommenders/factories yet.

reckart:
@Rentier ok, I see it.

rcffc:
For testing the NamedEntityLinker I need to create an instance of ConcepFeatureSupport. It would need to be moved from inception-ui-kb to inception-kb because otherwise there would be circular dependencies.

reckart:
@rcffc I believe we cannot move it because if we do, we get yet another circular dependency. Did you try moving it? How about mocking it?

rcffc:
@reckart  I am already mocking FeatureSupportRegistry, which is passed to NamedEntityLinker's constructor. At prediction time, the following code is executed:

```
        AnnotationFeature feat = annoService
            .getFeature(recommender.getFeature(), recommender.getLayer());
        FeatureSupport<ConceptFeatureTraits> fs = fsRegistry.getFeatureSupport(feat);
        ConceptFeatureTraits traits = fs.readTraits(feat);
```

 FeatureSupport<ConceptFeatureTraits> cannot be instantiated.

Rentier:
Cant you mock the `fsRegistry` to return a mocked `FeatureSupport` that returns your desired `ConceptFeatureTraits` when `readTraits` is called on it?

rcffc:
Thanks, that worked.

Rentier:
@reckart When the traits are fixed then we can merge imho.

reckart:
@Rentier ok, traits are back.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
```diff
@@ -18,213 +18,223 @@
...
     private KnowledgeBaseService kbService;
     private ConceptLinkingService clService;
-    private DocumentService documentService;
     private AnnotationSchemaService annoService;
     private FeatureSupportRegistry fsRegistry;
 
-    public NamedEntityLinker(ClassifierConfiguration<Object> aConf, KnowledgeBaseService aKbService,
-        ConceptLinkingService aClService, DocumentService aDocService,
-        AnnotationSchemaService aAnnoService, FeatureSupportRegistry aFsRegistry, String aFeature)
+    private int numPredictions = 3;
```
reckart: This should probably be configured in a trait.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
```diff
@@ -18,213 +18,223 @@
...
-            }
-            catch (IOException e) {
-                log.error("An error occurred while retrieving entity candidates.", e);
-                return Collections.emptyList();
-            }
-        });
+    @Override
+    public double evaluate(List<CAS> aCasses, DataSplitter aDataSplitter)
+    {
+        return -1;
```
reckart: Should throw an exception when called, e.g. `UnsupportedOperationException`.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -0,0 +1,219 @@
...
+            String msg = String.format("Request was not succesfull: [%d] - [%s]", code, status);
+            throw new RuntimeException(msg);
+        }
+
+        try {
+            XmiCasDeserializer.deserialize(response.body().byteStream(), aCas);
+        }
+        catch (SAXException | IOException e) {
+            LOG.error("Error while deserializing CAS!", e);
+            throw new RuntimeException("Error while deserializing CAS!", e);
```
reckart: Should throw RecommendationException.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -0,0 +1,219 @@
...
+
+        Response response = sendRequest(request);
+
+        // If the response indicates that the request was not successfull,
+        // then it does not make sense to go on and try to decode the XMI
+        if (!response.isSuccessful()) {
+            int code = response.code();
+            String status = response.message();
+            String msg = String.format("Request was not succesfull: [%d] - [%s]", code, status);
+            throw new RuntimeException(msg);
```
reckart: Should throw RecommendationException.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -0,0 +1,219 @@
...
```
reckart: Should throw RecommendationException.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/PredictionRequest.java
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.recommendation.imls.external;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class PredictionRequest {
```
reckart: Format class according to DKPro Core style.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/TrainingRequest.java
```diff
@@ -0,0 +1,76 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.recommendation.imls.external;
+
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class TrainingRequest {
```
reckart: Format class according to DKPro Core style.

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
```diff
@@ -0,0 +1,166 @@
...
+        recommendationEngine.train(context, casses);
+    }
+
+    private TrainingRequest deserializeTrainingRequest(String aRequestJson)
+    {
+        try {
+            ObjectMapper objectMapper = new ObjectMapper();
+            return objectMapper.readValue(aRequestJson, TrainingRequest.class);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
reckart: Should throw IOException.

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
```diff
@@ -0,0 +1,166 @@
...
+        return serializeCasToXmi(cas);
+    }
+
+    private PredictionRequest deserializePredictionRequest(String aPredictionRequestJson)
+    {
+        try {
+            ObjectMapper objectMapper = new ObjectMapper();
+            return objectMapper.readValue(aPredictionRequestJson, PredictionRequest.class);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
reckart: Should throw IOException.

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
```diff
@@ -0,0 +1,166 @@
...
+    // CAS handling
+
+    private CAS deserializeCas(String xmi, String typeSystem)
+    {
+        byte[] casBytes = Base64.getDecoder().decode(xmi);
+        CAS cas = buildCas(typeSystem);
+        try (InputStream bais = new ByteArrayInputStream(casBytes)) {
+            XmiCasDeserializer.deserialize(bais, cas);
+        } catch (IOException | SAXException e) {
+            throw new RuntimeException(e);
```
reckart: Should thrown IOException.

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        // We need to save the typeSystem XML to disk as the
+        // JCasFactory needs a file and not a string
+        try {
+            File typeSystemFile = File.createTempFile("typeSystem", ".xmi");
+            FileUtils.writeByteArrayToFile(typeSystemFile, Base64.getDecoder().decode(typeSystem));
+            JCas jCas = JCasFactory.createJCasFromPath(typeSystemFile.getAbsolutePath());
+            return jCas.getCas();
+        } catch (IOException | UIMAException e) {
+            throw new RuntimeException(e);
```
reckart: Should throw IOException.

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    private String serializeCasToXmi(CAS aCas)
+    {
+        try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
+            XmiCasSerializer.serialize(aCas, null, out, true, null);
+            return new String(out.toByteArray(), "utf-8");
+        }
+        catch (CASRuntimeException | SAXException | IOException e) {
```
reckart: Should throw IOException.

inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommenderTest.java
```diff
@@ -0,0 +1,176 @@
...
+{
+    private static File cache = DkproTestContext.getCacheFolder();
+    private static DatasetFactory loader = new DatasetFactory(cache);
+
+    private RecommenderContext context;
+    private Recommender recommender;
+    private OpenNlpNerRecommenderTraits traits;
+
+    @Before
+    public void setUp() {
```
reckart: Bracket on next line.

inception-recommendation-api/src/main/resources/desc/type/inception-recommendation-api-typesystem.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<typeSystemDescription xmlns="http://uima.apache.org/resourceSpecifier">
+    <name>inception-recommendation-api-typesystem</name>
+    <description/>
+    <version>1.0</version>
+    <vendor/>
+  <types>
+    <typeDescription>
+      <name>de.tudarmstadt.ukp.inception.recommendation.api.type.PredictedSpan</name>
```
reckart: @Rentier didn't you say we add the original type to the CAS instead of the `PredictedSpan`? Or does that only apply when talking to the DKPro TC external recommender?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -62,66 +77,95 @@ public void run()
...
+                            .getRecommenderFactory(recommender);
+                    RecommendationEngine recommendationEngine = factory.build(recommender);
+
                     try {
-                        JCas jcas = documentService.readAnnotationCas(doc);
-                        predictions.addAll(classifier.predict(jcas, layer));
-                    } catch (IOException e) {
-                        log.error("Cannot read annotation CAS.", e);
+                        recommendationEngine.predict(ctx, jCas.getCas());
+                    } catch (RecommendationException e) {
```
reckart: Should catch `Throwable` here.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -62,66 +77,95 @@ public void run()
...
+                        log.error("Error while predicting", e);
```
reckart: Error message should include information on which recommender/document/project the problem happened and whether that causes the entire prediction to fail or only the document to be skipped (or the remainder of the document to be skipped?).

inception-recommendation-api/src/main/resources/desc/type/inception-recommendation-api-typesystem.xml
reckart: ping?

inception-imls-external/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/RemoteStringMatchingNerRecommender.java
reckart: Never mind - test code.

Rentier:
It breaks when I run

    mvn -U clean  install -DskipTests

with

```Compilation failure
[ERROR] /home/klie/git/inception-github/inception-imls-dl4j/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/dl4j/pos/DL4JSequenceRecommenderFactory.java:[58,22] cannot find symbol
```

Rentier:
I pulled it, and the external recommender module disappeared somehow from the main pom.

Rentier:
I checked out the wrong repo, my fault.

inception-recommendation-api/src/main/resources/desc/type/inception-recommendation-api-typesystem.xml
Rentier: I do not understand your question.

reckart:
*phew* :)

inception-recommendation-api/src/main/resources/desc/type/inception-recommendation-api-typesystem.xml
Rentier: I made it flexible so that a recommender can decide what type he wants to predict. Right now, we let external recommender predict the target annotation, internal recommender predict `PredictedSpan`.

Rentier:
OpenNLP pos is still missing, I thought I added it.

reckart:
@Rentier I can see a `OpenNlpPosClassificationToolFactory` here which seems to be v2:

https://github.com/inception-project/inception/tree/feature/398-recommender-cas-in-out/inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/pos

Rentier:
```
    @Override
    public RecommendationEngine build(Recommender aRecommender) {
        return null;
    }
```

I will add it.

inception-recommendation-api/src/main/resources/desc/type/inception-recommendation-api-typesystem.xml
reckart: I guess that means that right now we don't get confidence scores from the external recommender meaning that it cannot be used with active learning, right?

Rentier: Yes, that is true. But you can specify the feature that the recommender predicts. If we make it configurable via the UI then you could implement an external recommender that uses `PredictedSpan`.

reckart:
@Rentier I assume you are asking because of the OpenNLP POS recommender. The recommender should be enabled only for layers with the anchoring mode `single token`. Then you can assume that every token has either one or no POS annotation. There might be select methods in `WebAnnoCasUtil` that can be used (`selectAt` or `selectSingleAt` returning null). That would be preferable to re-implementing these methods I guess - because then refactoring is easier if we find way of improving these selections.

Rentier:
@reckart `selectSingleFsAt` in Webanno uses `JCas`, but I use`CAS`, therefore I have to use `selectAt.get(0)`.

Rentier:
I added the OpenNLP pos and removed a lot of old code, especially the evaluation code. Thanks to the magic of source control, we can find that stuff when we reimplement it. I just want to get this pull request done, therefore I prefer deleting over fixing for now.

rcffc:
The OpenNlp Pos classifier does not display recommendations for all tokens. It did in the beginning, but then most of the recommendations disappeared.

screenshot from 2018-10-15 17-24-14

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/recommender/RecommenderContext.java
```diff
@@ -0,0 +1,64 @@
...
+        store = new ConcurrentHashMap<>();
+    }
+
+    @SuppressWarnings("unchecked")
+    @Nullable
+    public <T> T get(Key<T> aKey)
+    {
+        String name = aKey.name;
+        if (!store.containsKey(name)) {
+            LOG.warn("Value with key [{}] not found in context!", name);
```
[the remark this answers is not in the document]
Rentier: I annotated it with `Nullable`, you could also do `Optional`. I changed it because it can be that a recommender is not trained yet, but is asked for predictions.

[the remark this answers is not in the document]
reckart: Well, your code - you choose :) Just update the test and calling code accordingly. I'm recently moving more towards using Optional. I tried the JSR annotations once, but Eclipse wasn't very helpful with them.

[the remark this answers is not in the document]
reckart: I'm adding a `isReadyForPrediction()` flag to the context.

reckart:
> The OpenNlp Pos classifier does not display recommendations for all tokens. It did in the beginning, but then most of the recommendations disappeared.

@Rentier said that the POS recommender internally uses a "padding" (no label) tag for all tokens that do not have an annotation. That may lead to a strong class imbalance and to the POS recommender simply always predicting the padding tag (which means no prediction). At least that would be my present theory. I didn't do any debugging on this so far.

Rentier:
I only add sentences to the training set that have at least one annotation in it. OpenNLP POS throws an exception if there is only one tag type per sentence.

reckart:
@Rentier @rcffc the problem with the POS tag recommendations disappearing was due to class imbalance. The more you annotated, the more sentence with few POS and many PAD labels were generated, the more the PAD was overweight. I have mitigated this by allowing to the OpenNLP POS recommender to not only return the best sequence, but the n best sequences (configurable via "Max recommendations" in the project recommender settings). I recommend setting this at least to 3, better even to 5 (maybe even higher) to get reasonably stable POS recommendations.

reckart:

Rentier:
I will push the documentation for the external recommender on Monday, review the PR and then we can merge. Enjoy your weekend!

reckart:

reckart:

reckart:

reckart:

reckart:

PR #411, opened 2018-07-18
Bugfix/269 active learning deactivate when switching to other sidebar

matchlesswei:
Active learning is still active after a user switches to other sidebar and then switches back.


inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -102,4 +114,339 @@ public ActiveLearningServiceImpl(DocumentService aDocumentService,
...
+        boolean aHasUnseenRecommendation)
+    {
+        ActiveLearningUserState state = getState(aUser.getUsername(), aProject);
+        synchronized (state) {
+            state.setHasUnseenRecommendation(aProject, aHasUnseenRecommendation);
+        }
+    }
+
+    @Override
+    public boolean getHasUnseenRecommendation(User aUser, Project aProject)
```
reckart: Get methods with boolean return values start with `is`, not with `get`.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -102,4 +114,339 @@ public ActiveLearningServiceImpl(DocumentService aDocumentService,
...
+    {
+        ActiveLearningUserState state = getState(aUser.getUsername(), aProject);
+        boolean hasUnseenRecommendation;
```
reckart: I think there is no need for the variable, just `return` from inside the synchronized block (applies to other methods as well).

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -102,4 +114,339 @@ public ActiveLearningServiceImpl(DocumentService aDocumentService,
...
+        ActiveLearningUserState state = getState(aUser.getUsername(), aProject);
+        boolean hasSkippedRecommendation;
+        synchronized (state) {
+            hasSkippedRecommendation = state.isHasSkippedRecommendation(aProject);
+        }
+        return hasSkippedRecommendation;
+    }
+
+    @Override
+    public void putDoExistRecommender(User aUser, Project aProject, boolean aDoExistRecommenders)
```
reckart: "set" methods start with `set` - applies to other methods as well.

Why is there a `doExistRecommender` flag in the state? Shouldn't this information be obtained from the recommender service? In the state, this information could easily be outdated.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -102,4 +114,339 @@ public ActiveLearningServiceImpl(DocumentService aDocumentService,
...
+    private void clearState(String aUserName)
+    {
+        synchronized (states) {
+            states.remove(aUserName);
+        }
+    }
+
+    private static class ActiveLearningUserState
+    {
+        private Map<Long, Boolean> sessionActive = new HashMap<>();
```
reckart: Instead of having a map for every single variable, why not have a higher-level map with the project ID as the key and the `ActiveLearningUserState` as the value?

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -172,6 +175,21 @@ public ActiveLearningSidebar(String aId, IModel<AnnotatorState> aModel,
         super(aId, aModel, aActionHandler, aJCasProvider, aAnnotationPage);
 
         annotationPage = aAnnotationPage;
+        project = aModel.getObject().getProject();
```
reckart: Do not copy the values (have them as fields in the component) - always retrieve them from the model object when needed. You may keep them for a short time in local variables, but not in class fields.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -172,6 +175,21 @@ public ActiveLearningSidebar(String aId, IModel<AnnotatorState> aModel,
...
+        user = aModel.getObject().getUser();
+
+        //set the user states to the ActiveLearningSidebar fields
+        sessionActive = activeLearningService.getSessionActive(user, project);
```
reckart: The components should read/write directly to the state, not to local copies of the values. So all these variables should be removed.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -188,21 +206,24 @@ public ActiveLearningSidebar(String aId, IModel<AnnotatorState> aModel,
...
-        // If the currently selected layer has no recommenders, use the first one which has
-        else if (!layersWithRecommenders.isEmpty()) {
-            selectedLayer = Model.of(layersWithRecommenders.get(0));
-        }
-        // If there are no layers with recommenders, then choose nothing and show no recommenders
-        // message.
-        else {
-            selectedLayer = Model.of();
-            doExistRecommenders = false;
+        if (!sessionActive) {
```
reckart: `sessionActive` must be obtained from the active learning state here.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -931,4 +955,19 @@ else if (vMarkerType.equals(TEXT_MARKER)) {
             }
         }
     }
+
+    private void setUserState()
```
reckart: All components must directly read/write the AL state fields. So this method should be removed and proper models (e.g. LambdaModels)  should be used which connect components directly to the underlying state.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -485,20 +458,71 @@ private void jumpToRecommendationLocationAndHighlightRecommendation(AjaxRequestT
         highlightCurrentRecommendation(aTarget);
     }
 
-    private FeatureEditor createFeatureEditor()
+    private FeatureEditor initializeFeatureEditor()
+    {
+        try {
+            createFeatureEditor();
```
reckart: Why should createFeatureEditor switch from one document to another? It seems this method is doing too much.

reckart:
I also think that the synchronization on the state is too fine-granular and unnecessary most of the time. If makes the code very complex and prevents the use of a `CompoundPropertyModel` for the state. If there are complex changes to the state, it probably needs to be synchronized outside of the service.

reckart:
I think the main point that needs to be synchronized is the creation and deletion of the states to e.g. to avoid that two states are created for the same project. Although, the way we use AL this is highly unlikely anyway. I believe access to the fields of the states do not need to be synchronized.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
matchlesswei: @reckart Why do you think createFeatureEditor switch from one document to another?

This method gets SourceDocument and uses SourceDocument to get AnnotationDocument and uses AnnotationDocument to get Jcas and uses Jcas to get FeatureState using the code:
```
featureState = new FeatureState(annotationFeature, (Serializable) featureSupport
            .wrapFeatureValue(annotationFeature, jCas.getCas(),
                currentRecommendation.getLabel()));
```

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
matchlesswei: Every time the active learning sidebar is opened, all the methods here are loaded. And I think this is the only time that these fields get value from the saved state. That's why I let these components read here.

matchlesswei:
@reckart So for using compoundPropertyModel, shall I create a class `ActiveLearningUserState` and save the field User(or Username) also in this class to avoid using `Map<String, ActiveLearningUserState>` for getting the state?

reckart:
@matchlesswei I think I see two options:

1. a nested map, i.e. `Map<String, Map<Long, ActiveLearningUserState>>` where the String is the user ID and the Long is the project ID
2. a flat map, i.e. `Map<ActiveLearningUserStateKey, ActiveLearningUserState>` where `ActiveLearningUserStateKey` has two fields, namely the project ID and the user ID.

inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/ActiveLearningServiceImpl.java
```diff
@@ -102,4 +112,266 @@ public ActiveLearningServiceImpl(DocumentService aDocumentService,
         return model.getFlattenedPredictions(aState.getDocument().getName(), aSelectedLayer,
             windowBegin, windowEnd, aJcas, true);
     }
+
+    @Override
+    public void setSessionActive(User aUser, boolean aSesscionActive)
```
reckart: If you make `getState(user)` public and allow then you could wrap `ActiveLearningUserState` it in a CompoundPropertyModel and use it in the forms. All the getters and setters here would not be needed and also `setUserState` would go away.

matchlesswei: Thanks @reckart I'm still not sure whether I understand it correctly. 

For example, in ActiveLearningSidebar, I created a field `private CompoundPropertyModel<ActiveLearningUserState> userStateModel;` and assigned this field `userStateModel = new CompoundPropertyModel<>(activeLearningService.getState(user.getUsername()));` when active learning sidebar generated.

ActiveLearningSiebar has a `sessionControlForm`, it contains a DropDownChoice to select the `AnnotationLayer` and a LambdaAjaxButton to show terminate/start and change the field `sessionActive`. (AnnotationLayer selectedLayer and boolean sessionActive are fields in ActiveLearningUserState.)

For the DropDownChoice, I think I should write:
```
DropDownChoice<AnnotationLayer> layersDropdown = new DropDownChoice<AnnotationLayer>(CID_SELECT_LAYER, userStateModel.bind("selectedLayer"));
layersDropdown.setChoices(LambdaModel.of(this::listLayersWithRecommenders));
```
This component is disabled when `sessionActive` is true. How can I bind `sessionActive` also to the DropDownChoice? Using `userStateModel.getObject().isSessionActive`?

For the second component LambdaAjaxButton, I wrote like this:
 ```
LambdaAjaxButton<Void> startStopButton = new LambdaAjaxButton<>(
                CID_LAYER_SELECTION_BUTTON, this::actionStartStopTraining);
startStopButton.setModel(userStateModel.bind("sessionActive"));
```
How could I change the button value according to "sessionActive"? (Before:         `startStopButton.setModel(LambdaModel.of(() -> sessionActive ? "Terminate" : "Start"));`)
How could I set `sessionActive` in the method `actionStartStopTraining`?

reckart: > For example, in ActiveLearningSidebar, I created a field private CompoundPropertyModel<ActiveLearningUserState> userStateModel; and assigned this field userStateModel = new CompoundPropertyModel<>(activeLearningService.getState(user.getUsername())); when active learning sidebar generated.

Instead of fetching the state there, I would wrap the state fetching into a `LambdaModelAdapter`, i.e. something like `new CompoundPropertyModel<>(LambdaModelAdapter.of(() -> activeLearningService.getState(user.getUsername()), state -> activeLearningService.setState(user.getUsername(), state))`. That directly binds the model to the underlying data from the ActiveLearningService.

In order to activate/deactivate a session, I think you can fetch the active learning state from the model and update it directly, e.g. `userStateModel.getModelObject().setActive(true/false).

In order to change the label of the `startStopButton`, I would try using a `LambdaBehavior`, e.g. `startStopButton.add(LambdaBehavior.configure(_component -> setModelObject(userStateModel.getModelObject().iActive() ? "Terminate" : "Start"));`

Does that help?

matchlesswei:
@reckart Now AL Sidebar uses CompoundPropertyModel to save and get the values like selectedLayer etc. 

Then there is  a problem: 
The state will be cleared with regard to the `DocumentOpenedEvent`. So if the next recommendation is in another document, the current state(like` selectedLayer` and `currentRecommendation`) will be cleared after `actionShowSelectedDocument` is called. How can we solve this problem?

matchlesswei:
After merging the master, I cannot pass the `testRawTextQuery()` in `MtasDocumentIndexTest`:

> java.lang.AssertionError: 
> Expected :1
> Actual : 2
 
Shall I create a new issue for it?

reckart:
> Then there is a problem:
The state will be cleared with regard to the DocumentOpenedEvent. So if the next recommendation is in another document, the current state(likeselectedLayer and currentRecommendation) will be cleared after actionShowSelectedDocument is called. How can we solve this problem?

You could remember the project in the sidebar and not reset the settings if `DocumentOpenedEvent` is within the same project.

matchlesswei:
@reckart , if a user moves from AnnotationPage to the WelcomePage, and then goes back to the AnnotationPage, if the user doesn't change project, the state cannot be reset.

reckart:
Hm. So what is the desired behavior, i.e. in which situations do we want the AL state to be cleared? What would you say?

matchlesswei:
I think only if a user leaves the AnnotationPage, for example to the WelcomePage, all the states should be cleared.

reckart:
The problem is that we can't really detect when a user leaves a page. The user might close the browser, the computer might crash, etc. - it is something that cannot be reliably detected.

We can detect when a user visits a page and we can also detect when the user's session times out.

matchlesswei:
Can we always keep the state? If a user reset a document, then we clear the state.

Or we don't use compoundPropertyModel? Then we can avoid this problem and clear the state when the user opens a new document.

reckart:
> Or we don't use compoundPropertyModel? Then we can avoid this problem and clear the state when the user opens a new document.

Not sure how not using the CompoundPropertyModel would help. We do not want to clear the state when a user switches between documents. The active learning even switches between documents itself as it considers recommendations in all documents of a project. 

I think what we are implicitly thinking about is that the state of the active learning is somehow connected to the annotation page itself - i.e. that it is independent of whether the sidebar is visible or not but gets initialized when a user opens the page. 

Thinking about it, we could probably attach the AL state directly to the annotation page. That would ensure that the state exists as long as the user is on the page. We could do this by storing the state using something like `getPage().setMetaData(ActiveLearningMetaData.STATE, activeLearningUserState)`. This would actually remove the need for maintaining the state in the `ActiveLearningService` and for explicitly monitoring the user's session and clearing the state when the session times out. As long as the user remains on the page, we'd have access to the state. When the user opens the page again (or even if the user opens it in another browser window), we get a new state. When the user leaves a page, Wicket automatically cleans up the state after some time along with the rest of the page data.

Actually, that would probably solve many of our problems. 

WDYT?

reckart:
@matchlesswei mind that any information we add as metadata to the page would need to be serializable. That may not be the case for all of the fields in `ActiveLearningUserState`.

matchlesswei:
@reckart Component meta data is now used to set the ActiveLearningUserState. Even I didn't store and compare the project information, when a user switches to another project, the userState is automatically cleared. Changes can be reviewed.

PR #416, opened 2018-07-20
Release/413 afer release merge 0.4.x into master

Rentier:
Immediately after the release, again merge the 0.4.x branch into the master branch, but this time using the following command:

`git merge -s ours 0.4.x`

Doing this second merge does not actually change anything in the master. However, it ensures that the updates to the version number in the 0.4.x branch will not make it to the master branch.


reckart:
Btw. this is not related to the demo server - the demo server doesn't exist yet.

Rentier:
@reckart Yes, but it was the label that was closest for me.

Issue #170, opened 2018-05-09
Panel on KB UI with Search functionality for KB entries 


naveen2507:
- [x] Search over KB annotations in search side bar
- [x] Panel on KB UI to show the instance annotations on text
- [x] Utility class for indexing of KB Linked concepts
- [x] List data from KB linking on KB UI

naveen2507:
Done PR #223

reckart:
@naveen2507 the listing of concept / instance mentions is rather confusing:

* the mentions list is always in in the middle column of the screen which generally contains information about the class - even if it shows instance mentions when an instance is selected. Would have expected separate mention lists, one in the "class" column and one in the "instance" column. I can understand that a single list might be more convenient, but then I believe it should be spanning the "class" and the "instance" columns, i.e. appear below both of them.
* the mentions list shows only the annotated word - no left or right context, and no information about in which document the annotation was found.

naveen2507:
> the mentions list is always in in the middle column of the screen which generally contains information about the class - even if it shows instance mentions when an instance is selected. Would have expected separate mention lists, one in the "class" column and one in the "instance" column. I can understand that a single list might be more convenient, but then I believe it should be spanning the "class" and the "instance" columns, i.e. appear below both of them.

@reckart What happens now is, when we click on instances it shows the mention for the instances and as we discussed initially about the possibility of including parent concept for search here once we include parent and sub-super class concept thingy in KBSI. 

WDYS? Should we have two separate **Mentions** for concept and instances OR include the search to one with the parent concept for instance and subclass for concept.

reckart:
This is what we have now

```
+---+------------+----------+
|   | Class      | Instance |
|   |            |          |
|   |------------+----------|
|   | Class      | Instance |
|   | Stmts      | Stmts    |
|   |------------|          |
|   | Mentions   |          |
|   | (cls/inst  |          |
+---+-----------------------+
```

I could imagine this:

```
+---+----------+----------+
|   | Class    | Instance |
|   |          |          |
|   |----------+----------|
|   | Class    | Instance |
|   | Stmts    | Stmts    |
|   |----------|----------|
|   | Mentions | Mentions |
|   | Class    | Instance |
+---+---------------------+
```

or 

```
+---+----------+----------+
|   | Class    | Instance |
|   |          |          |
|   |----------+----------|
|   | Class    | Instance |
|   | Stmts    | Stmts    |
|   |---------------------|
|   | Mentions            |
|   | (class or instance) |
[...]

PR #423, opened 2018-07-26
Bugfix/420 cannot delete kb instances

naveen2507:
When deleting instances in the KB view, it throws an exception


naveen2507:
Issue fixed

reckart:
Why do we have so many commits in this PR?

naveen2507:
@reckart Merged statement validation branch here. As discussed initially about the **Concept Mentions**, wanted to change the search to include subclass and for instances its parent concept. But, will discuss it now in Issue #170

Issue #420, opened 2018-07-25
Cannot delete KB instances


Rentier:
**Describe the bug**
When deleting instances in the KB view, it throws an exception: 

`Error: Failed to invoke @OnEvent method`

**To Reproduce**
1. Create a new knowledge base
2. Go to the knowledge base view
3. Create a concept
4. Create a instance
5. Try to delete the instance

**Expected behavior**
Instance is deleted

**Please complete the following information:**
Master 0f9cf014d4f201d25924aaf9d5c3f056164637b6
Version 67.0.3396.99 (Official Build) (64-bit)

```
java.lang.IllegalStateException: Failed to invoke @OnEvent method
	at org.wicketstuff.event.annotation.AnnotationEventSink.onEvent(AnnotationEventSink.java:173) ~[wicketstuff-annotationeventdispatcher-7.10.0.jar:7.10.0]
	at org.wicketstuff.event.annotation.AnnotationEventSink.onEvent(AnnotationEventSink.java:108) ~[wicketstuff-annotationeventdispatcher-7.10.0.jar:7.10.0]
	at org.wicketstuff.event.annotation.AnnotationEventDispatcher.dispatchEvent(AnnotationEventDispatcher.java:86) ~[wicketstuff-annotationeventdispatcher-7.10.0.jar:7.10.0]
	at org.apache.wicket.settings.FrameworkSettings.dispatchEvent(FrameworkSettings.java:161) ~[wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ComponentEventSender.dispatchToComponent(ComponentEventSender.java:282) ~[wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ComponentEventSender.access$100(ComponentEventSender.java:36) ~[wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ComponentEventSender$ComponentEventVisitor.component(ComponentEventSender.java:329) ~[wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ComponentEventSender$ComponentEventVisitor.component(ComponentEventSender.java:306) ~[wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.util.visit.Visits.visitChildren(Visits.java:144) ~[wicket-util-7.10.0.jar:7.10.0]
	at org.apache.wicket.util.visit.Visits.visitChildren(Visits.java:162) ~[wicket-util-7.10.0.jar:7.10.0]
[...]

PR #424, opened 2018-07-26
#420 Cannot delete KB instance

naveen2507:
When deleting instances in the KB view, it throws an exception:


Rentier:
It works, but it logs 
```2018-07-26 16:34:48 WARN [admin] XmlPartialPageUpdate - Component '[ContentPanel [Component id = content]]' with markupid: 'id138' not rendered because it was already removed from page
```

naveen2507:
@Rentier It logs because of the instance component which is being selected is deleted and it changes the selection to the concept. I think for now, its not an issue for now. I will look into this when i change the UI for instance "concept mentions" .

@Rentier  the PR can be approved.

naveen2507:
@Rentier changed the base to 0.4.x

Rentier:
Why are there so many commits in here?

naveen2507:
@Rentier Opening another pull request.

Issue #421, opened 2018-07-26
Cannot create subclass


reckart:
**Describe the bug**
An exception when trying to create a subclass in the knowledge base page:

```
2018-07-26 12:05:20 ERROR [admin] KnowledgeBasePanel - Unable to find property subclassof.
java.util.NoSuchElementException: No value present
	at java.util.Optional.get(Optional.java:135) ~[?:1.8.0_144]
	at de.tudarmstadt.ukp.inception.ui.kb.SubclassCreationDialog$ContentPanel.actionCreateSubclass(SubclassCreationDialog.java:132) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.support.lambda.LambdaAjaxButton.onSubmit(LambdaAjaxButton.java:52) [classes/:?]
	at org.apache.wicket.ajax.markup.html.form.AjaxButton$1.onSubmit(AjaxButton.java:113) [wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ajax.form.AjaxFormSubmitBehavior$AjaxFormSubmitter.onSubmit(AjaxFormSubmitBehavior.java:218) [wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.markup.html.form.Form.delegateSubmit(Form.java:1311) [wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.markup.html.form.Form.process(Form.java:977) [wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.markup.html.form.Form.onFormSubmitted(Form.java:798) [wicket-core-7.10.0.jar:7.10.0]
	at org.apache.wicket.ajax.form.AjaxFormSubmitBehavior.onEvent(AjaxFormSubmitBehavior.java:174) [wicket-core-7.10.0.jar:7.10.0] 
   ...
```

**To Reproduce**
1. Create a new empty local KB in the project settings (I used the "OWL" profile)
2. Open the KB in the KB page
3. Create a new top-level class `AAA`
4. Press "create subclass" and try creating a class `BBB`

**Expected behavior**
`BBB` is created as a subclass of `AAA`

**Please complete the following information:**
 - Version and build ID: INCEpTION 𝜶 -- 0.5.0-SNAPSHOT (2018-07-26 11:27:24, build 98aaa3c967e31eed6a4270b8139698e7af67e399)

**Additional context**

naveen2507:
[...]

PR #425, opened 2018-07-26
#421 Cannot create subclass

naveen2507:
- Issue #421 
- Issue because of NO subClassOf property found in KB. 

Solution Description: 

- Initialize base properties while registering or updating KB
- Base Properties : subClassOf, comment  and label 
- When updating KB , delete the list of older base properties.


naveen2507:
Issue fixed. Needs review

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -386,18 +386,25 @@ public void deleteConcept(KnowledgeBase kb, KBConcept aConcept)
     @Override
     public KBHandle createProperty(KnowledgeBase kb, KBProperty aProperty)
     {
-        if (StringUtils.isNotEmpty(aProperty.getIdentifier())) {
+        if (StringUtils.isNotEmpty(aProperty.getIdentifier())
+                && (!aProperty.getIdentifier().equals(kb.getSubclassIri().stringValue()))) {
```
reckart: What is happening here? The exception message doesn't seem to match the condition. I also do not understand the condition - there is no comment explaining what happens.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -386,18 +386,25 @@ public void deleteConcept(KnowledgeBase kb, KBConcept aConcept)
...
             throw new IllegalArgumentException("Identifier must be empty on create");
         }
-
         return update(kb, (conn) -> {
-            String identifier = generateIdentifier(conn, kb);
+            String identifier;
+            if (aProperty.getIdentifier() != null && aProperty.getIdentifier().toString()
+                    .equals(kb.getSubclassIri().stringValue())) {
```
reckart: Same here. Why is there an exception for the subclass IRI? There is no comment explaining why this condition is used.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/SubclassCreationDialog.java
```diff
@@ -128,10 +129,20 @@ protected void actionCreateSubclass(AjaxRequestTarget aTarget, Form<KBConcept> a
 
                 // get subclassof property
                 KnowledgeBase kb = kbModel.getObject();
-                KBProperty property = kbService.readProperty(kb, kb.getSubclassIri().stringValue())
-                        .get();
-                KBHandle propertyHandle = new KBHandle(property.getIdentifier(), property.getName(),
-                        property.getDescription());
+                Optional<KBProperty> propertyContainer = kbService.readProperty(kb,
```
reckart: Again there is no explanatory comment here.

reckart: Also, the PR doesn't include a test case.

reckart: I suppose we didn't have these problems when the inferencer was still enabled because it would have initialized the KB with the basic RDF and OWL IRIs. Maybe it would be a good idea that we prep freshly created local KBs with a few basic things to start with, e.g. the foundational RDF, RDFS and OWL classes and properties?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: We do not need any extra message here as `!aProperty.getIdentifier().equals(kb.getSubclassIri().stringValue())` is added as IF condition, to avoid a only check of `StringUtils.isNotEmpty(aProperty.getIdentifier()`. This will only happen when we create subclass.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/SubclassCreationDialog.java
naveen2507: Yeah , we have these problems after we removed the inferencer. As we dont have the base properties such as `subClassOf`. 

>  Maybe it would be a good idea that we prep freshly created local KBs with a few basic things to start with, e.g. the foundational RDF, RDFS and OWL classes and properties?

In that case, we can just initialize the basic properties with out actually tagging them to class instances. The list can be : 
- subClassOf
- instanceOf
WDYS, do we need any other property?

naveen2507: But the idea is where should we actually do that, should we do that when we read the concept tree for the first time? OR in listProperty ?

naveen2507: We will actually then try to replicate the basic inferencer. Need to check if that would create any issues with the medium sized KB's

naveen2507:
@Rentier changed the base to 0.4.x

Issue #428, opened 2018-07-30
SPARQL query does not define required prefixes


reckart:
**Describe the bug**

Unabled to use KB at `http://zbw.eu/beta/sparql/stw/query`:

> Unable to list root concepts: Error 400: Parse error: SELECT DISTINCT ?s ?l WHERE { { ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Class> . } UNION { ?someSubClass <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?s . } . FILTER NOT EXISTS { ?s <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?otherSub . FILTER (?s != ?otherSub) } FILTER NOT EXISTS { ?s owl:intersectionOf ?list . } OPTIONAL { ?s <http://www.w3.org/2000/01/rdf-schema#label> ?l . FILTER(LANG(?l) = "" || LANGMATCHES(LANG(?l), "en")) } } LIMIT 10000 **Line 8, column 13: Unresolved prefixed name: owl:intersectionOf** Fuseki - version 3.8.0 (Build date: 2018-06-25T11:00:34+0000)

**To Reproduce**
1. Add a remote KB with the URL `http://zbw.eu/beta/sparql/stw/query`, no reification, OWL profile
2. Try opening the KB in the KB page

**Expected behavior**
The KB contents show on the KB page.

**Please complete the following information:**
 - Version and build ID: 
   - 0.4.1 (2018-07-20 12:36:20, build de3629c5c054990f5da64b014a719cf2e42e148a) **stable**
   - 0.5.0-SNAPSHOT (2018-07-30 14:50:27, build 72d641f25ca6cadcb578899c78793ed2eb393ba8) **experimental**

Rentier:
I get 

```
2018-08-02 11:36:55 DEBUG [admin] AnnotatedListIdentifiers - Error in the query.
de.tudarmstadt.ukp.inception.search.ExecutionException: Query not executed because index is in invalid state. Try again later.
	at de.tudarmstadt.ukp.inception.search.SearchServiceImpl.query(SearchServiceImpl.java:245) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.SearchServiceImpl$$FastClassBySpringCGLIB$$d6146f50.invoke(<generated>) ~[classes/:?]
[...]

PR #429, opened 2018-07-31
#428 SPARQL query does not define required prefixes

naveen2507:
- Add prefixes to query from SPARQLQueryStore.java


naveen2507:
Resolved

Issue #300, opened 2018-06-17
Introduce LIMIT constant for retrieving Statements from the Knowledge Base


rcffc:
RdfUtils does not have a configurable LIMIT parameter yet. We might also need a limit parameter for other places where a limit is applied.

reckart:
@rcffc Does #435 resolve this issue or are there more places where this constant needs to be introduced?

rcffc:
There are still some limits in WikiDataReification and KnowledgeBaseServiceImpl (line 414).

rcffc:
Okay.

PR #435, opened 2018-08-02
#300 Introduce LIMIT constant for retrieving Statements from the KB

naveen2507:
- Limit in the query


naveen2507:
@rfcc in #300, can you elaborate the issue and why are you limiting just to RDFUtil.java?

naveen2507:
@rcffc Okay. Yeah, I will commit the code after the change on KBSI

reckart:
@naveen2507 Tests don't seem robust enough. I'm removing the "needs review" tag. Please add it again when the tests are passing.

PR #449, opened 2018-08-10
Release/445 merge to 0.4.x

naveen2507:
- Merge to 0.4.x

### DIFF
--- a/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
+++ b/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
@@ -315,12 +315,15 @@ public class KnowledgeBaseServiceImplIntegrationTest  {
     @Test
     public void nonempty_WithEmptyKnowledgeBase_ShouldReturnTrue() {
         sut.registerKnowledgeBase(kb, sut.getNativeConfig());
+        
         sut.defineBaseProperties(kb,false);
+        
         List<KBHandle> listProperties = sut.listProperties(kb, true);
         Stream<String> listIdentifier = listProperties.stream().map(KBHandle::getIdentifier);
-        assertEquals(listProperties.size(),3);
         String[] expectedProps = { kb.getSubclassIri().stringValue(),
                 kb.getLabelIri().stringValue(), kb.getDescriptionIri().stringValue() };
+        
+        assertEquals(listProperties.size(),3);
         assertThat(listIdentifier).as("Check that base properties are created")
                 .contains(expectedProps);
     }
--- a/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
+++ b/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
@@ -319,7 +319,7 @@ public class KnowledgeBaseServiceImplIntegrationTest  {
         List<KBHandle> listProperties = sut.listProperties(kb, true);
         Stream<String> listIdentifier = listProperties.stream().map(KBHandle::getIdentifier);
         assertEquals(listProperties.size(),3);
-        String expectedProps[] = { kb.getSubclassIri().stringValue(),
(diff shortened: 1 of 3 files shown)


inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
```diff
@@ -316,7 +317,12 @@ public void nonempty_WithEmptyKnowledgeBase_ShouldReturnTrue() {
         sut.registerKnowledgeBase(kb, sut.getNativeConfig());
         sut.defineBaseProperties(kb,false);
         List<KBHandle> listProperties = sut.listProperties(kb, true);
+        Stream<String> listIdentifier = listProperties.stream().map(KBHandle::getIdentifier);
```
Rentier: Please split your tests into three parts:
1) Setup
2) Action you want to test
3) Assertions

naveen2507: Because we add the properties in accordance to KB schema settings, when we update KB schema (supposedly form OWL to wikidata) , we need to delete the properties from OWL and and then add default properties from wikidata otherwise it will add up to list of  default properties

naveen2507: We needed kbUpdate flag to tell the method explicitly that, the call has been made from the update and we need to delete the old property values now.

naveen2507: The name of the fields at times are the same, so we cannot just update the properties. It can cause duplicate name on property list and we definitely dont want that.

Rentier: Yes, but you delete all properties from what I see.

naveen2507: > Please split your tests into three parts:
> Setup
> Action you want to test
> Assertions

What do u mean to say here?

Rentier: Look at the other tests and how they are structured. They are written in three blocks.

naveen2507: > Yes, but you delete all properties from what I see. That means, if you call this method, ALL properties are gone.

Yeah, when u change ur schema, it intializes the KB again along with the default properties and reads the local/remote ontology again

naveen2507: Ahhan , you mean spaces for blocks . Okay

Rentier: `listProperties` returns ALL properties, not only ontology properties.

naveen2507: `if (kb.getType() == RepositoryType.LOCAL) {
           kbService.defineBaseProperties(kb,true);
`
and we call the method only for LOCAL type KB. 

So, what happens now is we didn't had the subClass as property anymore and even when we have an empty KB we dont see anything on KB page. To sort this out, we had to add this idea that when we initialize KB we should add the default properties which were earlier introduced by inferencer

naveen2507: > You would also need to change for readOnly in the first block.

I didn't get you. Can you give the reference

Rentier: https://github.com/inception-project/inception/pull/449/files#diff-2c7aa5a57022089e3825589d1f7534c1R165

PR #450, opened 2018-08-10
Bugfix/421 cannot create subclass issue

naveen2507:
- Issue #421 
- Issue because of NO subClassOf property found in KB. 

Solution Description: 

- Initialize base properties while registering or updating KB
- Base Properties : subClassOf, comment  and label 
- When updating KB , delete the list of older base properties. 

- TO DO : what if the schema is custom with some non standard subclass IRI


naveen2507:
@reckart Need ur suggestion on this.

So the approach right now is as follows:

- We add default subClass, label and description property to the local KB (empty as well as in case we import an ontology like wine) depending on the schema. 
- Now if the user updates the schema, we list all the properties and delete the properties which are not the base property for any of the defines schemas. Refer : Line - [170](https://github.com/inception-project/inception/blob/7407b222a03d30b3b4c47eac57a7bfa344cf062d/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java#L170) and [894](https://github.com/inception-project/inception/blob/7407b222a03d30b3b4c47eac57a7bfa344cf062d/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java#L894)

So we have a question here, when the user updates the schema with a custom schema IRI definition, he has the leverage to add any thing as subClass IRI. Am I right? Should we change something on that as we can provide the base property to IRI.

reckart:
I believe that it should work as follows:

When a user changes any of the IRIs in the KB settings, it should be checked if a corresponding IRI definition exists in the KB. So. if the user sets the `label` IRI, then on saving the configuration, it should be checked that the KB actually contains a statement like `(label-iri, type-of-iri, property-iri)` and also corresponding statements for all the other IRIs.

I am not sure if it is a good idea to delete any statements as part of the process. E.g. if we delete the statement that `(label-iri, type-of-iri, property-iri)`, so we also delete all other statements about `label-iri` (e.g. any range definitions)? Since we do not have an "undo" function, a user might easily end up deleting statements that are then difficult to re-create. The user might not even be aware what was deleted.

I am also not sure if it is a good idea to "update" existing data to the changed IRIs. E.g. if the `label` IRI is changed, I believe that statements such as `(X, label-iri, "some label)` should not be touched. I  believe that changing the mapping is one thing and refactoring the KB is another one. We might offer functionality at some point to refactor the KB, e.g. to bulk-change statements - but I believe this should be separate from the configuration of the mapping.

So I believe that when we setup/change the mapping, we should only verify that the KB contains essential statements required for the mapping to work and to create such statements if they are missing. I fear that changing existing statements or deleting existing statements (even if they have been auto-created before) may lead to hard-to-repair problems which could force the user to in the best case simply re-import the KB or in the worst case have to rebuild it almost from scratch. I'd be very careful about such operations and only consider introducing them after we have more experience with how users interact with the KB and whether they request such functionality or not.

It seems that it would be a good idea to inform the user about such missing / auto-created statements. This could happen e.g. via a popup dialog before saving (e.g. with a message like "The following essential statements will be automatically created when you save these settings. Are you sure?") or just by displaying a message after saving.

naveen2507:
@reckart We can list then set of statements but that would be N number of statements (where N can be at times more that no of concepts) . For now, I am not deleting any property/statement on update and add the subClass property depending on the current update of KB schema.

naveen2507:
@reckart you can review this PR now. We have removed the delete thingy for the base property as of now. We can have a detailed discussion on it and open an issue again for the further approach.

reckart:
> @reckart We can list then set of statements but that would be N number of statements (where N can be at times more that no of concepts)

I think we might have different sets of statements in mind. Which statements do you mean?

naveen2507:
> I think we might have different sets of statements in mind. Which statements do you mean?

I mean all the statement defined for subClassOf in case the ontology has the inferred statements included. The same applies for label and comment statements. If the ontology suits OWL schema and in case the user update the schema to Wikidata (or some random Custom IRI) , then we have to list down all the statements for subClassOf and update the user for deleting the statements and this would seriously hamper the concept tree structure as well. Ofcourse we don't want that at all.

reckart:
When I said it would be a good idea to list auto-created / missing statements, I was only talking about the statements that would be created when the user saves the KB settings, e.g.

* `(label-iri, is-a-iri, property-iri)`
* `(description-iri, is-a-iri, property-iri)`
* `(subclassOf-iri, is-a-iri, property-iri)`
* `(subpropertyOf-iri, is-a-iri, property-iri)`
* ...

If any of these statements are missing when saving the KB settings, we'd have to automatically create them - and then we could inform the user either before or after the creation.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -159,6 +159,23 @@ public void registerKnowledgeBase(KnowledgeBase kb, RepositoryImplConfig cfg)
         repoManager.addRepositoryConfig(new RepositoryConfig(repositoryId, cfg));
         entityManager.persist(kb);
     }
+    
+    @Override
+    public void defineBaseProperties(KnowledgeBase kb, boolean kbUpdateFlag)
```
reckart: `aKb`

Parameter `kbUpdateFlag` is not used.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -159,6 +159,23 @@ public void registerKnowledgeBase(KnowledgeBase kb, RepositoryImplConfig cfg)
         repoManager.addRepositoryConfig(new RepositoryConfig(repositoryId, cfg));
         entityManager.persist(kb);
     }
+    
+    @Override
+    public void defineBaseProperties(KnowledgeBase kb, boolean kbUpdateFlag) 
+    {
+        // KB will initialize base properties with base IRI schema properties defined by user
+        if (kb.getType() == RepositoryType.LOCAL) {
+            KBProperty property = new KBProperty(kb.getSubclassIri().getLocalName());
```
reckart: I'd suggest to introduce a constructor for `KBProperty` with two arguments: identifier and name - and then to use that here.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -159,6 +159,23 @@ public void registerKnowledgeBase(KnowledgeBase kb, RepositoryImplConfig cfg)
...
+            property.setIdentifier(kb.getSubclassIri().stringValue());
+            createBaseProperty(kb, property);
+            property = new KBProperty(kb.getLabelIri().getLocalName());
```
reckart: Please avoid re-purposing variables. Anyway, when the constructor mention above has been introduced, this variable is not required anymore.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -859,6 +885,22 @@ private boolean hasImplicitNamespace(String s)
         }
         return false;
     }
+    
+    public boolean isIdentiferBaseProperty(KBHandle handle)
+    {
+
+        List<IRI> listIRI = new ArrayList<IRI>();
```
reckart: `new ArrayList<IRI>()`-> `new ArrayList<>()` - btw. you do this a lot, the generic type is not required in these cases - Java can automatically infer it.

naveen2507:
But what about the statements dependent on the the properties. I tried to do a bit of trial and test thing after me and @Rentier discussed about the idea in detail. The idea of updating the schema has issues as it can effect the ontology and concept tree in general specially when there are statements dependent on the `subClassOf / label / comment` .  And, as we also allow user to set any other random IRI as subClass and after working on part of creating a KB from scratch, what if the user changes the schema. There are several cases which have issues when we update the schema. I agree that we should show the user which statements gets updated. But there can be statement which gets impacted by inference and not directly.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -859,6 +885,22 @@ private boolean hasImplicitNamespace(String s)
         }
         return false;
     }
+    
+    public boolean isIdentiferBaseProperty(KBHandle handle)
```
reckart: This method doesn't seem to be used anywhere. I also think it doesn't do any good because it doesn't consider the IRIs configured for the KnowledgeBase.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseService.java
```diff
@@ -398,5 +398,16 @@ RepositoryImplConfig getKnowledgeBaseConfig(KnowledgeBase kb)
...
+    /**
+     * Define base default properties of comment, label and subClassOf with schema set defined for
+     * KB while initializing the KB
+     * 
+     * @param kb
+     *            The knowledge base to initialize base properties
+     * @param kbUpdateFlag
+     *            Flag variable to indicate the update of KB or not.
+     */
+    void defineBaseProperties(KnowledgeBase kb, boolean kbUpdateFlag);
```
reckart: `aKb` + `aKbUpdateFlag`

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -849,6 +866,15 @@ private ReificationStrategy getReificationStrategy(KnowledgeBase kb)
             return new NoReification(this);
         }
     }
+    
+    // Method to create and define base property
```
reckart: Comment should be JavaDoc comment and also explain what a "base property" and when this method should be used instead of `createProperty`.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Its for letting the function aware of the updating the KB or not. Can be removed for now, but we would definitely need that when we implement a different approach for updating the KB.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Yeah, its an old school habit ;) Will remove it.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Yeah, this method was meant to identify the base IRIs and check if the configured KnowledgeBase IRI matches the base one.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Java doc comment is there in KBService.

reckart:
>But what about the statements dependent on the the properties. 

Which statements do you mean?

> I tried to do a bit of trial and test thing after me and @Rentier discussed about the idea in detail. The idea of updating the schema has issues as it can effect the ontology and concept tree in general specially when there are statements dependent on the subClassOf / label / comment. And, as we also allow user to set any other random IRI as subClass and after working on part of creating a KB from scratch, what if the user changes the schema. 

If a user changes the schema after already having created statements, then - for the time being - it is their problem. Changing the mapping is different from refactoring the KB. We offer support for changing the mapping, but not for refactoring.

**Case 1**: A user should be able to import an existing ontology and then try out which mapping works best for it. It should be possible to do this in a way which does not actually change the ontology. 

**Case 2:** A user should also be able to create a new ontology from scratch (no import). In this case, we must create some basic statements in the KB because otherwise we run into #421. 

Case 2 has a side-effect on Case 1 in the way that using a "wrong" mapping with an imported ontology may cause the creation of statements which were not in the original ontology. But the number of these statements is reasonably small and adding them has (as far as I can see atm) negligible impact on the imported KB.

Independent of this, we may want to support KB refactoring, i.e. renaming IRIs. A refactoring operation may be "change IRI of property X". Such an operation would then go through all the statements which use the property X and replace its IRI with the changed IRI. But this is different from configuring the mapping. If the refactoring were entailed by the changing of the mapping, then it would conflict with Case 1.

> There are several cases which have issues when we update the schema. I agree that we should show the user which statements gets updated. But there can be statement which gets impacted by inference and not directly.

These are issues related to **refactoring**, right? Not to **mapping configuration**. I think we need to consider them when we work on KB refactoring, not now.

naveen2507:
@reckart you have addressed the concerns I had. So we would need the KB refactoring part to fix the issue in complete. But for now the current issue is resolved. We just have some more implicit properties on the list in case the user plays around with schema.

Issue #453, opened 2018-08-13
Unit test case failing


naveen2507:
**Describe the bug**
Test is failing on master for Jenkins check

**To Reproduce**
PR failing due to issue #444 

**Expected behavior**
Jenkins check should pass

**Screenshots**
![image](https://user-images.githubusercontent.com/10420950/44029262-f1e48208-9efc-11e8-81b8-19c750faccf5.png)

naveen2507:
Need to fix this issue for release of 0.4.2

betoboullosa:
Duplicate of #444.

Issue #444, opened 2018-08-09
Document indexed twice


betoboullosa:
**Describe the bug**
When adding a new document in a project that still does not have an index, the document is indexed twice. This happens because the creation of a new index indexes all project documents, and, at the moment the new document indexing is called through SearchService.afterDocumentCreate, the document is already present in the project. 

**To Reproduce**
This error is reproduced in the MtasDocumentIndexTest.testRawTextQuery test.

**Expected behavior**
When a new document is added to a project that still does not have an index, it should not be indexed again. 

**Please complete the following information:**
 - Version and build ID: 0.5.0-SNAPSHOT

betoboullosa:
Fixed in https://github.com/inception-project/inception/pull/456/commits/87ea527019e8e2b90e51a96a789c848e661c76c6

reckart:
See https://github.com/inception-project/inception/commit/87ea527019e8e2b90e51a96a789c848e661c76c6#r30058995

The change in 87ea527 unfixes a previous fix. There must be a better approach.

Please do not commit directly to the branches and stick to doing PRs unless surgery on the git repository is required.

reckart:
Never mind, I found the PR: https://github.com/inception-project/inception/pull/456

@naveen2507 are you really sure you wanted to approve this change?

naveen2507:
Yeah @reckart. The issue couldn't be reproduced in 0.5.0 due to the code at line https://github.com/inception-project/inception/blob/87ea527019e8e2b90e51a96a789c848e661c76c6/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java#L631

This code wasn't there in 0.4.2 , so issue has been addressed here.

reckart:
@naveen2507 ok, great. Then closing this issue again.

PR #454, opened 2018-08-13
#453 Unit test case failing

naveen2507:
- Fixing the test, as the document is indexed twice. Issue fix in #444


reckart:
This is obviously a recognized bug. Why not fix the bug instead of "breaking" the unit test?

betoboullosa:
This PR will be closed. It does not make sense.

PR #455, opened 2018-08-13
Feature/445 prepare release 0.4.2

naveen2507:
- Prepare release for 0.4.2


reckart:
@naveen2507 because of the accident with other PR (i.e. that master was merged into 0.4.x), we have way too many changes now in this PR. In particular, this PR sets all the version numbers of the master branch to 0.4.2-SNAPSHOT.

To fix this, please change the version numbers in the POMs in this branch to from `0.4.2-SNAPSHOT` to `0.5.0-SNAPSHOT` (global search/replace should work).

inception-imls-dl4j/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>de.tudarmstadt.ukp.inception.app</groupId>
     <artifactId>inception-app</artifactId>
-    <version>0.5.0-SNAPSHOT</version>
```
reckart: This version should be set to 0.5.0-SNAPSHOT as well.

inception-imls-mira/pom.xml
```diff
@@ -21,7 +21,7 @@
   <parent>
     <groupId>de.tudarmstadt.ukp.inception.app</groupId>
     <artifactId>inception-app</artifactId>
-    <version>0.5.0-SNAPSHOT</version>
```
reckart: This version should be set to 0.5.0-SNAPSHOT as well.

naveen2507:
@reckart these projects were not there on my local. I will check the PR for more changes

betoboullosa:
As far as I've noticed, the 444 bug appeared because of the introduction of a call to createPhysicalIndex() in MtasDocumentIndex constructor, apparently introduced to circumvent some bug related to the KB search. Nevertheless, the current code of the 0.4.x branch doesn't seem to have the MtasDocumentIndexTest that triggers the bug. In order to simplify things, I've fixed it only in 0.5.x, but I still to have to check with @naveen2507 which are the consequences for the KB search.

naveen2507:
@reckart can you review this PR.

reckart:
Doing it in a new PR so we can better inspect what is happening in the two attempts.

reckart:
https://github.com/inception-project/inception/pull/455

PR #461, opened 2018-08-14
Feature/298 pre defined importable knowledge bases

MdeBoer95:


inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
```diff
@@ -1,6 +1,7 @@
 british_museum:
   name: British Museum
-  sparql-url: http://collection.britishmuseum.org/sparql
+  type: REMOTE
+  access-url: http://collection.britishmuseum.org/sparql
```
naveen2507: Why do you need to change the name here? anything specific?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/wizard/KnowledgeBaseCreationWizard$LocalRepositoryStep.html
```diff
@@ -32,6 +32,22 @@
...
+                      </button>
+                  </wicket:container>
+              </div>
+          </div>
+          <button wicket:id="download" type="button" class="btn btn-outline-secondary"><wicket:container wicket:id="downloadLabel"></wicket:container></button>
+      </div>
+
   </wicket:panel>
 </body>
 </html>
```
naveen2507: No newline at EOF. Take a note to fix this.

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
MdeBoer95: It's more general. For remote KBs it is the sparql endpoint, for local KBs it is the download URL.

reckart: We could indeed include some "built-in" ontologies which we could import from a classpath URL.

naveen2507: @MdeBoer95 As @reckart suggested, it would be a better idea instead to have some built it ontogies on classpath URL. 

@reckart do we see a use case in future to provide user a leverage to set up her own set of predefined importable KB?

reckart: > @reckart do we see a use case in future to provide user a leverage to set up her own set of predefined importable KB?

Consider a research group deploying an INCEpTION instance for internal use. The group has some internal KB. It would be nice if they could put these KBs into a configuration file such that users can simply select them from the list when they start a new project instead of having to enter all the details manually.

But IMHO allowing users to set up a second user-defined configuration file or alternatively adding custom KB profiles to settings.properties would be the subject of an separate issue.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/wizard/KnowledgeBaseCreationWizard.java
```diff
@@ -206,19 +243,93 @@ public IDynamicWizardStep next() {
...
+                        {
+                            // Can not download the same KB more than once
+                            setEnabled(
+                                !downloadedProfiles.containsKey(item.getModelObject().getName()));
+                        }
+                    };
+
+                    String itemLabel = item.getModelObject().getName();
+                    // Adjust label to indicate whether the KB has already been downloaded
+                    if (downloadedProfiles.containsKey(item.getModelObject().getName())) {
```
naveen2507: @MdeBoer95  We can add the ontology in the classpath and it will not be needed to download them into tmp folder. WDYS

naveen2507:
Also please take a look at the conflicting files

MdeBoer95:
So do we only want to include pre-defined local KBs via the classpath and not download any KB at all or do we want to provide both options (Opt. 1: download, then import, Opt. 2: import from classpath)?

reckart:
@MdeBoer95 for the time being, I think including only via the classpath should be sufficient. But on the other hand, downloading would also be nice and probably isn't really much extra effort to implement, right?

MdeBoer95:
The download option is already implemented. So I will add the importing via classpath.

naveen2507:
@MdeBoer95 so we already have the implementation for download, you just have to add the classpath process flow into the code. Apart from that, we can download the files into the classpath. WDYS?

reckart:
@naveen2507 the classpath usually consists of JARs or even of a single JAR. While in principle possible, we should not try to download anything *into the classpath*. A downloadable remote KB should be downloaded into a temporary file, imported, and then deleted.

For classpath resources, I would suggest to use the Spring `PathMatchingResourcePatternResolver`. Even if it is not used, classpath locations should be expected to start with `classpath:`.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/wizard/KnowledgeBaseCreationWizard.java
reckart: Downloaded ontologies should to into a temporary file which is deleted after the import (mind it should also be deleted in case the import fails, so the delete needs to happen in a "finally" block!

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseService.java
```diff
@@ -528,4 +529,16 @@ RepositoryImplConfig getKnowledgeBaseConfig(KnowledgeBase kb)
      */
     List<KBHandle> listAllConcepts(KnowledgeBase kb, boolean aAll) throws QueryEvaluationException;
 
+    SchemaProfile checkSchemaProfile(KnowledgeBaseProfile aProfile);
```
reckart: Please add JavaDoc.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SchemaProfile.java
```diff
@@ -57,6 +57,8 @@ private SchemaProfile(IRI aClassIri, IRI aSubclassIri, IRI aTypeIri, IRI aDescri
         propertyTypeIri = aPropertyTypeIri;
     }
 
+    public String getLabel() { return label; }
```
reckart: Please apply DKPro Core style.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -40,6 +42,8 @@
     public void checkThatDeserializationWorks() throws JsonParseException, JsonMappingException, IOException {
         String name = "Test KB";
         String url = "http://someurl/sparql";
+        KnowledgeBaseAccessType accessType = KnowledgeBaseAccessType.CLASSPATH;
```
reckart: Shouldn't that be "LOCAL" instead of "CLASSPATH"?

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -49,13 +53,22 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
...
+        referenceAccess.setAccessUrl(url);
+        referenceAccess.setAccessType(accessType);
+
         referenceProfile.setMapping(referenceMapping);
         referenceProfile.setName(name);
-        referenceProfile.setSparqlUrl(url);
+        referenceProfile.setAccess(referenceAccess);
+        referenceProfile.setType(type);
         
         String test_yaml = "test_profile:\n    "
```
reckart: I'd suggest to please the test profile into a file under `src/test/resources` which is then loaded by the test.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/ProjectKnowledgeBasePanel.properties
```diff
@@ -41,6 +40,9 @@ kb.wizard.steps.type.name=Name
 kb.wizard.steps.type.type=Type
 kb.wizard.steps.schema=Schema
 kb.wizard.steps.local.fileupload=File(s)
+kb.wizard.steps.local.predefinedKBs = Pre-defined knowledge bases
+kb.wizard.steps.local.schemaOnMouseOver = Schema: {0}\nAccess: {1}
```
reckart: Does `\n` actually have an effect here? Browsers usually ignore line breaks in HTML. If this shows in the browser, it might have to be a  `<br>` instead.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/wizard/KnowledgeBaseCreationWizard.java
```diff
@@ -111,9 +126,29 @@ public KnowledgeBaseCreationWizard(String id, IModel<Project> aProjectModel) {
...
         init(wizardModel);
     }
 
+    private Map<String, KnowledgeBaseProfile> readKbProfiles() {
+        Map<String, KnowledgeBaseProfile> profiles = new HashMap<>();
+        try {
+            profiles = kbService.readKnowledgeBaseProfiles();
+        }
+        catch (IOException e) {
+            error("Unable to read knowledge base profiles" + e.getMessage());
```
reckart: Add `:<space>` at the end of the string before getMessage().

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/wizard/KnowledgeBaseCreationWizard.java
```diff
@@ -206,19 +241,118 @@ public IDynamicWizardStep next() {
...
+                    setKbIRIsAccordingToProfile(model.getObject().getKb(),
+                        selectedKnowledgeBaseProfile);
+                    downloadedProfiles
+                        .put(selectedKnowledgeBaseProfile.getName(), selectedKnowledgeBaseProfile);
+                    aTarget.add(listViewContainer);
+                    selectedKnowledgeBaseProfile = null;
+                }
+            }
+            catch (IOException e) {
+                error("Unable to download or import knowledge base file" + e.getMessage());
```
reckart: getMessage is glued to the error string. Add `:<space>`.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/ProjectKnowledgeBasePanel.properties
MdeBoer95: Yes, the \n works.

MdeBoer95:
>Downloaded ontologies should to into a temporary file which is deleted after the import (mind it should >also be deleted in case the import fails, so the delete needs to happen in a "finally" block!

@reckart Is it not sufficient that the tmp files are deleted by the user's system after a while ? We also treat the uploaded KBs like this.

reckart:
> @reckart Is it not sufficient that the tmp files are deleted by the user's system after a while ? We also treat the uploaded KBs like this.

No, that is not sufficient because "after a while" can be a **very long time** on a server.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
reckart: The URL says it all. If the URL starts with `classpath:`, then we should fetch the inputstream for the data from via the Spring resolver. Otherwise, we should try to use Java URL and get the stream from there. Once the stream has been obtained, the code is basically the same.

reckart:
>> @reckart Is it not sufficient that the tmp files are deleted by the user's system after a while ? We also treat the uploaded KBs like this.
> No, that is not sufficient because "after a while" can be a very long time on a server.

Wrt. Uploaded KBs: I didn't check it but it might be the case that the UI element through which the KBs are uploaded takes care of cleaning them up - or maybe even the wizard. If not, then action would be required there as well. Since neither the server machine nor the server application is restarted on a regular basis, temporary files that "clean up eventually" just pile up higher and higher.

Rentier:
For uploading knowledge base stuff, we use the `IFileCleaner`  of Wicket:

https://github.com/inception-project/inception/blob/a9cf2fa5e96a1f0e97ca99aeea5c80f8a7609fe8/inception-support/src/main/java/de/tudarmstadt/ukp/inception/kb/io/FileUploadHelper.java

reckart:
> So, should I create a "DownloadHelper" that does the same or just add another method to the class and rename it to something like UploadDownloadHelper...?

Why would you add a new class instead of simply using the FileUploadHelper that we already have?

MdeBoer95:
I just added a new method for writing a downloadable source to a tmp file. The upload uses [``FileUpload``](https://ci.apache.org/projects/wicket/apidocs/6.x/org/apache/wicket/markup/html/form/upload/FileUpload.html) from the wicket component [``FileUploadField``](https://ci.apache.org/projects/wicket/apidocs/6.x/org/apache/wicket/markup/html/form/upload/FileUploadField.html) which makes it slightly different from the download.

Issue #463, opened 2018-08-16
Concept features may lose connection to KB on project import


reckart:
**Describe the bug**
When a project is exported and then imported again on another instance of INCEpTION, the traits in the concept features may no longer point to the correct KB. The problem is that a new KB ID is generated on import, but the traits (and possible other references to the KBs) are not updated.

**To Reproduce**
1. Export a project with concept features that are bound to a KB from an instance A of INCEpTION.
2. Import the project to another instance B of INCEpTION

Note that both instances should have different numbers of projects to aggravate the problem.

**Expected behavior**
If a concept feature was bound to a KB on instance A, it should still be bound to that KB after the import to instance B.

**Please complete the following information:**
 - Version and build ID: 0.4.2 (2018-08-14 10:34:34, build 5880cba81d733892d44ca2b26bec50b52b0ebef5)

PR #464, opened 2018-08-17
#463 - Concept features may lose connection to KB on project import

reckart:
- Remap KB IDs in the traits during import


inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseExporterTest.java
```diff
@@ -105,8 +121,56 @@ public void thatExportingWorks() throws Exception
         verify(kbService, times(numOfLocalKBs)).importData(any(),
                 any(), any());
 
-        assertThat(exportedKbs).usingFieldByFieldElementComparator()
-                .containsExactlyInAnyOrderElementsOf(knowledgeBases());
+//        assertThat(exportedKbs).usingFieldByFieldElementComparator()
```
Rentier: Why are you weakening the tests here? We imho really need this check.

reckart: Try putting in the old check and tell me how you'd do it. Thing is that on re-import under normal circumstances actually not all fields are the same, in particular the `project` and `repositoryId` usually mismatch... and they mismatch now that I made these changes to the test.

reckart: @Rentier nevermind... found it :) Thanks for the prodding.

```
assertThat(exportedKbs)
                .usingElementComparatorIgnoringFields("repositoryId", "project")
                .containsExactlyInAnyOrderElementsOf(knowledgeBases());
```

Issue #489, opened 2018-08-24
Better indexing of relations


reckart:
**Is your feature request related to a problem? Please describe.**
It would be nice to slightly better support search of relations. Currently, the target of a relation is indexed and searchable, but not the source.

**Describe the solution you'd like**
Some kind of syntax that allows to address the text and features on the target and source of a relation annotations, e.g. `<Dependency-source="the"/>`.

**Describe alternatives you've considered**
Since Mtas | CQL does not support special operators for relations, I don't seem any alternatives.

PR #490, opened 2018-08-24
#489 - Better indexing of relations

reckart:
- Index source and target of relations specifically
- Use "-" instead of "." to separate special suffixes (this also affects ".exact" which changes to "-exact")
- Update documentation (including some reformatting)
- Upgrade MTAS
- Added profile to facilitate building the docs much faster


reckart:
@betoboullosa @naveen2507 Could either of you please review?

Issue #492, opened 2018-08-27
Ability to configure IRIs for property labels / descriptions


reckart:
**Is your feature request related to a problem? Please describe.**
At times the properties follow a slightly different schema than classes and instances. Thus it would be useful to be able to configure their IRIs separately:

- [x] property descriptions
- [x] property labels
- (subproperty IRI) -> this is actually tracked in a different issue

naveen2507:
DONE

PR #493, opened 2018-08-27
#492 - Ability to configure IRIs for property labels / descriptions

MdeBoer95:
- add comboboxes for configuration of property labels / descriptions
- adjust profiles and schema


inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
```diff
@@ -48,15 +48,22 @@
...
     
     
-    @JsonCreator
-    public KnowledgeBaseMapping(@JsonProperty("class") String classIri,
-            @JsonProperty("subclass-of") String subclassIri,
-            @JsonProperty("instance-of") String typeIri,
-            @JsonProperty("description") String descriptionIri,
-            @JsonProperty("label") String labelIri,
-            @JsonProperty("property-type") String propertyTypeIri)
+    @JsonCreator public KnowledgeBaseMapping(@JsonProperty("class") String classIri,
```
naveen2507: @MdeBoer95 Checkstyle issue.. plz make sure the code follows the DKPro style

MdeBoer95:
For which of the profiles do we need specific IRIs for the property labels/ descriptions ?
Right now I configured it to just use the same IRIs that are used for the concept labels/ descriptions

reckart:
@MdeBoer95 If I remember right then @Rentier mention that we did not get any labels/descriptions for properties for the UKP Wikipedia KB. However, I don't actually see which IRI we should use there. It seems to me almost as if the properties don't have any information associated with them at all.

I tried adding a KB using the official Wikidata SPARQL endpoint to see if we get more infos about properties there, but no. E.g. for `P1838` I only see this:



Even entering a query directly on the Wikidata SPARQL interface doesn't yield more:



It almost seems as if property information would not be included in the "direct" Wikidata -> PDF mapping.

So I guess for the moment, we do not need to update anything - so the PR is probably fine as it is. But we need to investigate the general problem further.

Rentier:
I would merge this if it is ok but comment and leave the issue open.

Issue #419, opened 2018-07-25
Language field in KB editor broken


Rentier:
**Describe the bug**
In the knowledge base statement editor, when setting the language field, several things do not work:

1) Confirmation to save with enter does not save
2) Confirmation to save with the check mark needs many clicks to do something, it also does not save
3) When trying to save, the concept tree label is changed to what looks like a BNode

**To Reproduce**

![language_field](https://user-images.githubusercontent.com/934600/43206757-b7bb0ca8-9026-11e8-8096-e0159a14b492.gif)

**Expected behavior**
The language is saved when I confirm the change either with enter or with the checkmark

**Screenshots**

**Please complete the following information:**
0f9cf014d4f201d25924aaf9d5c3f056164637b6
Version 67.0.3396.99 (Official Build) (64-bit)

Issue #443, opened 2018-08-08
Language Filter for SPARQL Query 


naveen2507:
**Describe the bug**
Language filter for retrieving Label doesn't work properly

**Expected behavior**
 Retrieve the english label and if it does not exist retrieve the label with no language

**Please complete the following information:**
 - Version and build ID: [Technische Universität Darmstadt -- Computer Science Department -- INCEpTION 𝜶 -- 0.5.0-SNAPSHOT (2018-08-06 14:02:37, build e230e9eb1add8be17759fca7ec081d0003d92107)]
 - OS: [Linux]
 - Browser: [chrome]

**Additional context**
Issue with language filter in SPARQL queries.

naveen2507:
PR #407

PR #495, opened 2018-08-27
#419 - language_field_in_kb_editor_broken

MdeBoer95:
- safe language value on change

This PR, together with the changes from #51 fixes the issues when saving a language value.

The reason that the concept tree shows a node-id when the language of the default label is changed to for example 'de' , is that we only look for labels with language en or "" when we query for the root concepts. So getUiLabel won't find any label for that concept and returns the node-id. 
I don't know if and in what way we want to change that.

@naveen2507 Do you have any suggestions?

### DIFF
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/StringLiteralValueEditor.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/value/editor/StringLiteralValueEditor.java
@@ -42,7 +42,8 @@ public class StringLiteralValueEditor
         value.add(new LambdaAjaxFormComponentUpdatingBehavior("change", t -> t.add(getParent())));
         add(value);
 
-        add(new TextField<>("language"));
+        add(new TextField<>("language")
+            .add((new LambdaAjaxFormComponentUpdatingBehavior("change", t -> t.add(getParent())))));
     }
     
     @Override


naveen2507:
@MdeBoer95 Issue #443 will take care of language for labels

reckart:
@MdeBoer95 Actually, we should not need the `LambdaAjaxFormComponentUpdatingBehavior`. The form should be submitted automatically when the "save" button is clicked, since the `StatementEditor` uses `(new LambdaAjaxButton<>("save", StatementEditor.this::actionSave));` which should submit the form. The `LambdaAjaxFormComponentUpdatingBehavior` should only be required in cases when we need to act directly on the input of the user (not only when `save` is pressed). One such case is when the use changes the value of the value type dropdown since we then immediately need to reload the value editor. But for the language field or input field, it should not be needed.

I believe what you may want to try instead is this:

```
        form.add(new AjaxFormSubmitBehavior("onsubmit") {
            @Override
            protected void onSubmit(AjaxRequestTarget target) {
                super.onSubmit(target);
                // HERE ADD THE COMPONENTS TO THE TARGET THAT YOU WANT TO
                //BE RE-RENDERED WHEN THE FORM IS SUBMITTED.
            }
        });
```

reckart:
@naveen2507 I thought we had an issue somewhere related to this language problem described by @MdeBoer95, but I do not find it. Do you remember it?

naveen2507:
> @naveen2507 I thought we had an issue somewhere related to this language problem described by @MdeBoer95, but I do not find it. Do you remember it?

Yeah, I commented that above.. issue #443

MdeBoer95:
> Actually, we should not need the LambdaAjaxFormComponentUpdatingBehavior

@reckart I added the  LambdaAjaxFormComponentUpdatingBehavior to the language field because the Textarea also has the LambdaAjaxFormComponentUpdatingBehavior and it caused some problems that one element had this and the other one not.

But after doing some testing it seems we don't really need the Behavior on any of these two elements.
Removing it also seems to solve the problem that one has two click twice to save a value in the StringEditor.

Do you remember why you added the LambdaAjaxFormComponentUpdatingBehavior to the TextArea?

reckart:
> Do you remember why you added the LambdaAjaxFormComponentUpdatingBehavior to the TextArea?

I have absolutely no idea.

MdeBoer95:
Ok, then I will just remove this in another PR to solve the double click problem.

Issue #504, opened 2018-08-30
Entity linking not working


Rentier:
**Describe the bug**
When trying to entity link, no suggestions are shown in the feature editor.

**Please complete the following information:**
Technische Universität Darmstadt -- Computer Science Department -- INCEpTION 𝜶 -- 0.5.0-SNAPSHOT (2018-08-30 12:02:56, build ef792c7c0535febf60cc185c513dbb55c9299bb8)

**Additional context**

```
java.lang.NullPointerException: null
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method) ~[?:1.8.0_131]
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62) ~[?:1.8.0_131]
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45) ~[?:1.8.0_131]
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423) ~[?:1.8.0_131]
        at java.util.concurrent.ForkJoinTask.getThrowableException(ForkJoinTask.java:598) ~[?:1.8.0_131]
        at java.util.concurrent.ForkJoinTask.reportException(ForkJoinTask.java:677) ~[?:1.8.0_131]
        at java.util.concurrent.ForkJoinTask.invoke(ForkJoinTask.java:735) ~[?:1.8.0_131]
        at java.util.stream.ForEachOps$ForEachOp.evaluateParallel(ForEachOps.java:158) ~[?:1.8.0_131]
        at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateParallel(ForEachOps.java:174) ~[?:1.8.0_131]
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:233) ~[?:1.8.0_131]
        at java.util.stream.ReferencePipeline.forEachOrdered(ReferencePipeline.java:423) ~[?:1.8.0_131]
        at java.util.stream.ReferencePipeline$Head.forEachOrdered(ReferencePipeline.java:593) ~[?:1.8.0_131]
        at de.tudarmstadt.ukp.inception.conceptlinking.recommender.NamedEntityLinker.predictSentences(NamedEntityLinker.java:136) ~[inception-concept-linking-0.5.0-SNAPSHOT.jar!/:?]
[...]

PR #505, opened 2018-08-30
#504 - Entity linking not working

reckart:
- If no allowed value type is set for a ConceptFeatureTrait, then assume INSTANCE (the traditional setting)


reckart:
This fixes a problem that I observed locally when trying to link entities, but it doesn't match the stack trace reported in #504.

The #504 stack trace hints at problems with the entity linking recommender which is probably a different issue than the dropdown not filling.

reckart:
Ok, second commit to this PR fixes the originally reported stack trace.

reckart:
@Rentier @rcffc you can review/merge this

PR #506, opened 2018-08-30
#489 - Better indexing of relations

reckart:
- Reduce logging level of tracing output to "trace"


reckart:
@Rentier @rcffc you can review/merge this

Issue #440, opened 2018-08-07
Duplicate entries for Wikidata


naveen2507:
**Describe the bug**
Duplicate entries for Wikidata in Concept tree 

**To Reproduce**
1. Create a project
2. Configure Wikidata in KB
3. Go to Knowledge Base 
4. Duplicate entries in Wikidata Concept tree

**Expected behavior**
Should not show duplicate entries

**Screenshots**

![image](https://user-images.githubusercontent.com/10420950/43770184-ec278f10-9a3b-11e8-874b-a6737ba19f2e.png)

**Please complete the following information:**
 - Technische Universität Darmstadt -- Computer Science Department -- INCEpTION 𝜶 -- 0.5.0-SNAPSHOT (2018-08-06 14:02:37, build e230e9eb1add8be17759fca7ec081d0003d92107)
 - Linux
 - chrome

**Additional context**

reckart:
I also observed this with at least one other ontology (either `http://zbw.eu/beta/sparql/gnd/query` or `http://zbw.eu/beta/sparql/stw/query` but I don't remember which concept).

naveen2507:
Issue with the #51 : When we add multiple comments to the entity, then it show the concept in duplicity. 

Possible Solution : We should be able to handle this with SPARQL, otherwise it would be computationally very expensive

reckart:
I still seem to see duplicates:

naveen2507:
Ohhk. Need to take a look into this.

reckart:
Opened new issue as it seems to be a new bug.

PR #514, opened 2018-09-05
#440 Duplicate Entries for Wikidata

naveen2507:
- Duplicate entries for Wikidata issue fix
- Add default root class fro DBPedia : http://www.w3.org/2002/07/owl#Thing


naveen2507:
@reckart the problem of duplicate on wikidata is resolved with these changes. Although, I am a bit stuck wondering that the query which is failing for our test case : 

```
PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX owl: <http://www.w3.org/2002/07/owl#>
SELECT DISTINCT ?s (MIN(?label) AS ?l) (MIN(?desc) AS ?d) WHERE { 
  { ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2002/07/owl#Class> . } 
  UNION { ?someSubClass <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?s . } .
  FILTER NOT EXISTS { 
    ?s <http://www.w3.org/2000/01/rdf-schema#subClassOf> ?otherSub . 
    FILTER (?s != ?otherSub) }
  FILTER NOT EXISTS { 
    ?s owl:intersectionOf ?list . }
  OPTIONAL {
    ?s <http://www.w3.org/2000/01/rdf-schema#label> ?label .
    FILTER(LANGMATCHES(LANG(?label), "en"))
 }

  OPTIONAL {
    ?s <http://www.w3.org/2000/01/rdf-schema#comment> ?desc .
    FILTER(LANGMATCHES(LANG(?desc), "en"))
 }

} GROUP BY ?s
LIMIT 1000
```
is running within seconds on https://dbpedia.org/sparql . Any ideas or pointers on this?  I guess there can be issues w.r.t time on `GROUP BY` queries but running okay on https://dbpedia.org/sparql is a bit strange for me

naveen2507:
@reckart yeah. I had a discussion about it with Daniil as well and we tried to figure out the issue and the reasoning behind it but didn't succeed though.

reckart:
When the a query is issued directly through http://dbpedia.org/sparql, then there a `default graph` setting is added to the query:

`http://dbpedia.org/sparql?default-graph-uri=http%3A%2F%2Fdbpedia.org&query=PREFIX+rdf%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23%3E%0D%0APREFIX+rdfs%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX+owl%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23%3E%0D%0ASELECT+DISTINCT+%3Fs+%28MIN%28%3Flabel%29+AS+%3Fl%29+%28MIN%28%3Fdesc%29+AS+%3Fd%29+WHERE+%7B+%0D%0A++%7B+%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23type%3E+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23Class%3E+.+%7D+%0D%0A++UNION+%7B+%3FsomeSubClass+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3Fs+.+%7D+.%0D%0A++FILTER+NOT+EXISTS+%7B+%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3FotherSub+.+%0D%0A++++FILTER+%28%3Fs+%21%3D+%3FotherSub%29+%7D%0D%0A++FILTER+NOT+EXISTS+%7B+%0D%0A++++%3Fs+owl%3AintersectionOf+%3Flist+.+%7D%0D%0A++OPTIONAL+%7B%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23label%3E+%3Flabel+.%0D%0A++++FILTER%28LANGMATCHES%28LANG%28%3Flabel%29%2C+%22en%22%29%29%0D%0A+%7D%0D%0A%0D%0A++OPTIONAL+%7B%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23comment%3E+%3Fdesc+.%0D%0A++++FILTER%28LANGMATCHES%28LANG%28%3Fdesc%29%2C+%22en%22%29%29%0D%0A+%7D%0D%0A%0D%0A%7D+GROUP+BY+%3Fs%0D%0ALIMIT+1000&format=text%2Fhtml&CXML_redir_for_subjs=121&CXML_redir_for_hrefs=&timeout=30000&debug=on&log_debug_info=on&run=+Run+Query+`

This makes the query run quite fast and return only a few results.

When manually removing this parameter from the URL, the query takes quite a bit longer and returns many more results - it doesn't time out though:

`http://dbpedia.org/sparql?query=PREFIX+rdf%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23%3E%0D%0APREFIX+rdfs%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0D%0APREFIX+owl%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23%3E%0D%0ASELECT+DISTINCT+%3Fs+%28MIN%28%3Flabel%29+AS+%3Fl%29+%28MIN%28%3Fdesc%29+AS+%3Fd%29+WHERE+%7B+%0D%0A++%7B+%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23type%3E+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23Class%3E+.+%7D+%0D%0A++UNION+%7B+%3FsomeSubClass+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3Fs+.+%7D+.%0D%0A++FILTER+NOT+EXISTS+%7B+%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3FotherSub+.+%0D%0A++++FILTER+%28%3Fs+%21%3D+%3FotherSub%29+%7D%0D%0A++FILTER+NOT+EXISTS+%7B+%0D%0A++++%3Fs+owl%3AintersectionOf+%3Flist+.+%7D%0D%0A++OPTIONAL+%7B%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23label%3E+%3Flabel+.%0D%0A++++FILTER%28LANGMATCHES%28LANG%28%3Flabel%29%2C+%22en%22%29%29%0D%0A+%7D%0D%0A%0D%0A++OPTIONAL+%7B%0D%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23comment%3E+%3Fdesc+.%0D%0A++++FILTER%28LANGMATCHES%28LANG%28%3Fdesc%29%2C+%22en%22%29%29%0D%0A+%7D%0D%0A%0D%0A%7D+GROUP+BY+%3Fs%0D%0ALIMIT+1000&format=text%2Fhtml&CXML_redir_for_subjs=121&CXML_redir_for_hrefs=&timeout=30000&debug=on&log_debug_info=on&run=+Run+Query+`

This is the URL generated by RDF4J and which times out:

`http://dbpedia.org/sparql/?query=PREFIX+rdf%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23%3E%0APREFIX+rdfs%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%3E%0APREFIX+owl%3A+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23%3E%0ASELECT+DISTINCT+%3Fs+%28MIN%28%3Flabel%29+AS+%3Fl%29+%28MIN%28%3Fdesc%29+AS+%3Fd%29+WHERE+%7B+%0A++%7B+%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F1999%2F02%2F22-rdf-syntax-ns%23type%3E+%3Chttp%3A%2F%2Fwww.w3.org%2F2002%2F07%2Fowl%23Class%3E+.+%7D+%0A++UNION+%7B+%3FsomeSubClass+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3Fs+.+%7D+.%0A++FILTER+NOT+EXISTS+%7B+%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23subClassOf%3E+%3FotherSub+.+%0A++++FILTER+%28%3Fs+%21%3D+%3FotherSub%29+%7D%0A++FILTER+NOT+EXISTS+%7B+%0A++++%3Fs+owl%3AintersectionOf+%3Flist+.+%7D%0A++OPTIONAL+%7B%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23label%3E+%3Flabel+.%0A++++FILTER%28LANGMATCHES%28LANG%28%3Flabel%29%2C+%22en%22%29%29%0A+%7D%0A%0A++OPTIONAL+%7B%0A++++%3Fs+%3Chttp%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23comment%3E+%3Fdesc+.%0A++++FILTER%28LANGMATCHES%28LANG%28%3Fdesc%29%2C+%22en%22%29%29%0A+%7D%0A%0A%7D+GROUP+BY+%3Fs%0ALIMIT+1000`

By adding `&timeout=30000` to that URL, the timeout can be prevented.

reckart:
I have tried hacking in the timeout parameter but that causes yet again a different error (some problem in parsing the server response).

I think the way to address this would be to avoid doing the automatic detection of root concepts in DBPedia and instead explicitly set the root concepts as part of the profile. I would also recommend explicitly setting root concepts for other large-scale resources like Wikidata.

naveen2507:
But wudn't setting the root concepts implicitly will limit them to some specifics and then it would be the same problem of missing out on some entities. But yeah I beleive the option for root concept setting for DBPedia seems most feasible solution right now with its own shoertcomings.

naveen2507:
I have tested the changes. @reckart can you please test this PR in case there are some issues? 

image

naveen2507:
I have tested the changes and compared the DBPedia heirarchy with http://mappings.dbpedia.org/server/ontology/classes .

Owl:Thing was not appearing for us as it doesn't suffice the conditions for our root query with just two statements in it for: label and comment.

@reckart can you please test this PR in case there are some issues?

image

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -807,6 +808,13 @@ public KBHandle update(KnowledgeBase kb, UpdateAction aAction)
                 resultList.add(conceptHandle);
             }
         }
+        else if (aKB.getType() == RepositoryType.REMOTE
```
reckart: Please add a comment briefly explaining what this branch is doing.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -817,6 +825,7 @@ public KBHandle update(KnowledgeBase kb, UpdateAction aAction)
                 tupleQuery.setBinding("pLABEL", aKB.getLabelIri());
                 tupleQuery.setBinding("pDESCRIPTION", aKB.getDescriptionIri());
                 tupleQuery.setIncludeInferred(false);
+                tupleQuery.setMaxExecutionTime(0);
```
reckart: I believe that `0` is actually the default (i.e. no timeout). Mind that on remote SPARQL connections, this setting has no effect. What are you trying to achieve?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/exporter/ExportedKnowledgeBase.java
```diff
@@ -51,6 +51,9 @@
 
     @JsonProperty("property_type_iri")
     private String propertyTypeIri;
+    
+    @JsonProperty("base_concept_iri")
+    private String baseConceptIri;;
```
reckart: One `;` too many.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
```diff
@@ -137,6 +137,12 @@
     @Column(nullable = false)
     private IRI propertyDescriptionIri;
 
+    /**
+     * The IRI for a base concept for instance for DBPedia as owl:Thing
+     */
+    @Column(nullable = false)
```
reckart: Uhm... we already have root concepts in the KB:

```
    /**
     * A List of explicitly defined root concepts that can be used if auto detection takes too long
     */
    @ElementCollection(fetch = FetchType.EAGER)
    @CollectionTable(name = "knowledgebase_root_classes")
    @Column(name = "name")
    private List<IRI> explicitlyDefinedRootConcepts = new ArrayList<>();
```

Why do we need this additional `baseConceptIri`?

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
```diff
@@ -11,6 +11,7 @@ british_museum:
       property-type: http://www.w3.org/1999/02/22-rdf-syntax-ns#Property
       property-label: http://www.w3.org/2000/01/rdf-schema#label
       property-description: http://www.w3.org/2000/01/rdf-schema#comment
+      base-concept : http://www.w3.org/2002/07/owl#Nothing
```
reckart: The root concepts in the KB should be added here:

```
root-concepts:
- http://www.w3.org/2002/07/owl#Nothing
```

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
```diff
@@ -1516,7 +1516,7 @@ public void readKnowledgeBaseProfiles_ShouldReturnValidHashMapWithProfiles() thr
             });
 
     }
-
+
```
reckart: Spurious change - not necessary.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Yeah. Its the default one. Its still WIP. I was checking out some things with the tuple query.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
naveen2507: `explicitlyDefinedRootConcepts` is something which user can specify to limit the concept tree but its not something which can be taken as the base concept IRI. I gave a thought about it, `baseConceptIri` would be the IRI to define the base super class for ontologies and can be used further in the application as a separate variable.  Otherwise we would have to ask the user to change the `explicitlyDefinedRootConcepts` in case the default base concept is already defined and she decides to use some of the sub class concept. WDYS?

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
naveen2507: http://www.w3.org/2002/07/owl#Nothing is taken as default (for now) with KBs without any base concept specified yet. We need to get the root concepts for each one of them .

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
reckart: The intention of introducing the `explicitlyDefinedRootConcepts` was to avoid having to run the auto-detection of root concepts. So if any `explicitlyDefinedRootConcepts` are defined, then `KBSI.listRootConcepts()` would return them instead of trying to auto-detect the root concepts. As far as I can see, it has been implemented in this way as well.

So why do we additionally need a `baseConceptIri`? 

From what I see in https://github.com/inception-project/inception/blob/6b2714e33f618dc553da42124a1b1808573a8e0e/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java#L811 , I'd say that defining a `baseConceptIri` is equivalent to defining a single `explicitlyDefinedRootConcepts`. So I don't see why we would need another field in the KB and another if/else branch in the `KBSI.listRootConcepts()`.

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
reckart: @naveen2507 did you try defining a root IRI for Wikidata (official as well as UKP Virtuoso)?

reckart:
@naveen2507 did you try defining a root IRI for Wikidata (official as well as UKP Virtuoso)?

naveen2507:
yeah, I tried to do it for WIkidata but i didn't found any root as its was a cyclic graph. I will take a look at it again.

reckart:
@naveen2507 @MdeBoer95 we have some conflicts here now.

reckart:
> yeah, I tried to do it for WIkidata but i didn't found any root as its was a cyclic graph. I will take a look at it again.

I believe this is the equivalent of "dbpedia:Thing" in Wikidata: https://www.wikidata.org/wiki/Q35120

naveen2507:
> I believe this is the equivalent of "dbpedia:Thing" in Wikidata: https://www.wikidata.org/wiki/Q35120

Yeah, thats the equivalent but it has agen as subclass and further person as subclass of agent which has subject as subclass which is subclass of entity. Actually I didn't found any hierarchical structure for Wikidata like in DBPedia http://mappings.dbpedia.org/server/ontology/classes/ , let me take a look at the other way round.

reckart:
@naveen2507 We know that Wikidata has some circles - but that doesn't make `Entity` less of a root-concept, does it? Since we added loop cancellation, we shouldn't have any trouble with the circles anymore either, right? The tree doesn't really care either - if a concept appears in multiple branches, then that's like it is - it won't crash because of that.

naveen2507:
@reckart @Rentier Hi Guys, I am stuck with this issue in Wikidata, any pointers on this :
So we have a query which works on https://query.wikidata.org : 
`SELECT * WHERE {wd:Q35120 ?p ?s. }`
whereas this query doesn't 
`SELECT * WHERE {<https://www.wikidata.org/wiki/Q35120> ?p ?s. }`

reckart:
`wd` == `http://www.wikidata.org/entity/`

So the second query must be

`SELECT * WHERE {<http://www.wikidata.org/entity/Q35120> ?p ?s. }`

reckart:
For more information on the default prefixes in Wikidata, check https://www.mediawiki.org/wiki/Wikibase/Indexing/RDF_Dump_Format#Full_list_of_prefixes

naveen2507:
> SELECT * WHERE {<http://www.wikidata.org/entity/Q35120> ?p ?s. }

Ohh, Yeah . Thanks :)

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
```diff
@@ -177,7 +194,8 @@ public boolean equals(Object o)
                 && Objects.equals(propertyTypeIri, that.propertyTypeIri)
                 && Objects.equals(descriptionIri, that.descriptionIri)
                 && Objects.equals(propertyLabelIri, that.propertyLabelIri)
-                && Objects.equals(propertyDescriptionIri, that.propertyDescriptionIri);
+                && Objects.equals(propertyDescriptionIri, that.propertyDescriptionIri)
+                && Objects.equals(rootConcepts.get(0), that.rootConcepts.get(0));
```
reckart: Should in principle compare the entire list of root concepts. Btw. *why* does this class need an equals method anyway?

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -55,17 +59,18 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
         String descriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
         String propertyLabelIri = "http://www.w3.org/2000/01/rdf-schema#label";
         String propertyDescriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
-
+        List<String> rootConcepts =  new ArrayList<>(Arrays.asList(OWL.NOTHING.stringValue()));
```
reckart: `asList(OWL.NOTHING.stringValue())` (with a static import of `Arrays.asList`) is sufficient unless you need the list to be mutable (which I do assume is not the case).

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -55,17 +59,18 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
...
         KnowledgeBaseMapping referenceMapping = new KnowledgeBaseMapping(classIri, subclassIri,
             typeIri, descriptionIri, label, propertyTypeIri, propertyLabelIri,
-            propertyDescriptionIri);
+            propertyDescriptionIri,rootConcepts);
```
reckart: Missing space before `rootConcepts`.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
```diff
@@ -1538,9 +1537,11 @@ public void readKnowledgeBaseProfiles_ShouldReturnValidHashMapWithProfiles() thr
         String descriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
         String propertyLabelIri = "http://www.w3.org/2000/01/rdf-schema#label";
         String propertyDescriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
-
-        KnowledgeBaseMapping testMapping = new KnowledgeBaseMapping(classIri, subclassIri, typeIri,
-            descriptionIri, label, propertyTypeIri, propertyLabelIri, propertyDescriptionIri);
+        List<String> rootConcepts =  new ArrayList<>(Arrays.asList(OWL.NOTHING.stringValue()));
```
reckart: `asList(OWL.NOTHING.stringValue())` (with a static import of `Arrays.asList`) is sufficient unless you need the list to be mutable (which I do assume is not the case).

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplIntegrationTest.java
```diff
@@ -1538,9 +1537,11 @@ public void readKnowledgeBaseProfiles_ShouldReturnValidHashMapWithProfiles() thr
...
+        
+        KnowledgeBaseMapping testMapping = new KnowledgeBaseMapping(classIri, subclassIri,
+            typeIri, descriptionIri, label, propertyTypeIri, propertyLabelIri,
+            propertyDescriptionIri,rootConcepts);
```
reckart: Missing space before `rootConcepts`.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -363,25 +369,29 @@ public static KnowledgeBase setOWLSchemaMapping(KnowledgeBase kb) {
...
     private static class TestConfiguration
     {
         private final String url;
         private final KnowledgeBase kb;
         private final String testIdentifier;
-        public TestConfiguration(String aUrl, KnowledgeBase aKb, String atestIdentifier)
+        private final String rootIdentifier;
+
+        public TestConfiguration(String aUrl, KnowledgeBase aKb, String atestIdentifier,
+                String aRootIdentifier)
```
reckart: Shouldn't this be a set?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
reckart: Well, if there is no need for the equals (hashcode() as well? because there should never be an equals without a hashcode and vice versa), then I would remove it which saves us the discussion whether or not the entire list needs to be compared ;)

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
naveen2507: I don't see a need of Set here as it just checks with root concept.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
naveen2507: Removed

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
reckart: There could be more than one root concept.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
```diff
@@ -416,4 +420,17 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(repositoryId, name);
     }
+    
+    public void applyRootConcept(KnowledgeBaseMapping aMapping) {
+        if (!aMapping.getRootConcepts().isEmpty()
+                && !(aMapping.getRootConcepts().contains(OWL.NOTHING.stringValue()))) {
+            if (explicitlyDefinedRootConcepts.isEmpty()) {
```
reckart: I do not understand why the conditions in lines 425-427 are required. IMHO it would be sufficient to simply have the code from line 428-431 plus a line before that which cleans out any existing explicitlyDefinedRootConcepts. Line 432 seems redundant. I also do not understand why `applyRootConcept` is a separate method - I believe it shouldn't be called independently of `applyMapping`.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
```diff
@@ -63,7 +67,8 @@
         @JsonProperty("label") String labelIri,
         @JsonProperty("property-type") String propertyTypeIri,
         @JsonProperty("property-label") String propertyLabelIri,
-        @JsonProperty("property-description") String propertyDescriptionIri)
+        @JsonProperty("property-description") String propertyDescriptionIri,
+        @JsonProperty("root-concept") List<String> rootConcepts)
```
reckart: `root-concept` should be `root-concepts` here. Best introduce constants for the JSON property names.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/yaml/KnowledgeBaseMapping.java
```diff
@@ -74,7 +79,8 @@
         this.propertyTypeIri = vf.createIRI(propertyTypeIri);
         this.propertyLabelIri = vf.createIRI(propertyLabelIri);
         this.propertyDescriptionIri = vf.createIRI(propertyDescriptionIri);
-    }
+        this.rootConcepts = rootConcepts;
```
reckart: Method argument names should start with `a` (e.g. `aRootConcepts`)  and then the `this.` is not needed.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -55,17 +58,18 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
         String descriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
         String propertyLabelIri = "http://www.w3.org/2000/01/rdf-schema#label";
         String propertyDescriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
-
+        List<String> rootConcepts =  Arrays.asList(OWL.NOTHING.stringValue());
```
reckart: Why do we still need the `OWL.NOTHING`? It should be sufficient if the list is empty. Saying explicitly that there is "no root concept" by adding `OWL.NOTHING` to the list is not required.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -80,6 +84,17 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
         
         KnowledgeBaseProfile testProfile = profiles.get("test_profile");
         
-        assertEquals(testProfile, referenceProfile);
+        assertEquals(testProfile.getAccess(), referenceProfile.getAccess());
```
reckart: I'm sure that assertj has a better way of comparing an object field-by-field.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceRemoteTest.java
```diff
@@ -318,20 +351,6 @@ public void thatParentListCanBeRetireved()
 
     }
 
-    @Test
```
reckart: Why was this test removed?

naveen2507: It was redundant.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
naveen2507: Because we don't want to overwrite the value of the root-concepts if the user explicitly provides root-concept in the configuration settings. What we can do is, remove `OWL.NOTHING` from the mapping file in yaml and set that to pass an empty list. We can remove the rest of the condition barring `if (explicitlyDefinedRootConcepts.isEmpty())` .

naveen2507:
@reckart Tests are failing for 

`org.hibernate.PropertyValueException: not-null property references a null or transient value : de.tudarmstadt.ukp.clarin.webanno.model.Project.mode`

Any idea? They are failing on master as well.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
```diff
@@ -382,7 +381,13 @@ public void applyMapping(KnowledgeBaseMapping aMapping)
...
         setPropertyLabelIri(aMapping.getPropertyLabelIri());
         setPropertyDescriptionIri(aMapping.getPropertyDescriptionIri());
-        applyRootConcept(aMapping);
+        if (explicitlyDefinedRootConcepts.isEmpty()) {
+            ValueFactory vf = SimpleValueFactory.getInstance();
+            for (String rootConcept : aMapping.getRootConcepts()) {
+                explicitlyDefinedRootConcepts.add(vf.createIRI(rootConcept));
+            }
+        }
+        setExplicitlyDefinedRootConcepts(explicitlyDefinedRootConcepts);
```
reckart: `setExplicitlyDefinedRootConcepts(explicitlyDefinedRootConcepts);` translates to `explicitlyDefinedRootConcepts  = explicitlyDefinedRootConcepts`, right? So it's a no-operation?

inception-kb/src/main/resources/de/tudarmstadt/ukp/inception/kb/knowledgebase-profiles.yaml
```diff
@@ -45,8 +44,7 @@ virtuoso:
       description: http://www.schema.org/description
       property-label: http://www.w3.org/2000/01/rdf-schema#label
       property-description: http://www.schema.org/description
-      root-concepts:
-            - http://www.w3.org/2002/07/owl#Nothing
+      root-concepts: [ ]
```
reckart: If the list is empty, you can probably just leave it out entirely. May require to change the constructor to handle the case that `aRootConcepts` is null an in that case set `rootConcepts` to an empty `ArrayList<>()`.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseProfileDeserializationTest.java
```diff
@@ -58,7 +56,7 @@ public void checkThatDeserializationWorks() throws JsonParseException, JsonMappi
         String descriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
         String propertyLabelIri = "http://www.w3.org/2000/01/rdf-schema#label";
         String propertyDescriptionIri = "http://www.w3.org/2000/01/rdf-schema#comment";
-        List<String> rootConcepts =  Arrays.asList(OWL.NOTHING.stringValue());
+        List<String> rootConcepts =  new ArrayList<String>();
```
reckart: Could be `Arrays.emptyList()` or `new ArrayList<>()`;

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
reckart: > Because we don't want to overwrite the value of the root-concepts if the user explicitly provides root-concept in the configuration settings.

Hmm... If the root concepts are part of the pre-configured mapping, then I think it would be strange not to add these root concepts when the mapping is selected. But I also understand that it could be annoying to the user to manually configure some root concepts and loosing them all when selecting some other pre-configured mapping. 

At least this behavior should be documented in the user documentation and there should be a comment in the code as well explaining the rationale behind this.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/model/KnowledgeBase.java
```diff
@@ -382,7 +381,13 @@ public void applyMapping(KnowledgeBaseMapping aMapping)
         setPropertyTypeIri(aMapping.getPropertyTypeIri());
         setPropertyLabelIri(aMapping.getPropertyLabelIri());
         setPropertyDescriptionIri(aMapping.getPropertyDescriptionIri());
-        applyRootConcept(aMapping);
+        if (explicitlyDefinedRootConcepts.isEmpty()) {
```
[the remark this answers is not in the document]
naveen2507: Because, when the user updates it.. we don't want to lost his set of explicitly defined root concept. Does it make sense or is there something I am missing ?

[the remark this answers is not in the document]
naveen2507: Because, when the user updates it.. we don't want to lost his set of explicitly defined root concept. Does it make sense or is there something I am missing ?

Issue #513, opened 2018-09-04
Make SPARQL query limit configurable by user


mromanello:
When querying a KB via SPARQL the number of results retrieved is limited, by default, to 1,000.

I'd like to be able to configure this limit manually via the inception configuration file and, if possible, I'd like to be able to specify that I want all results to be retrieved.

reckart:
I think it would be best to allow configuring this on a per-KB basis.

naveen2507:
@reckart Yeah, that's what we are doing. But the code on 0.5.x doesn't quite align with the specificity. Code on  the master has the structure set up for that.

reckart:
For 0.5.1, add a setting to be added to the `settings.properties` file. For master, we can make it configurable on a KB-basis and use the value from the `settings.properties` file as the default value. WDYT?

reckart:
For adding configuration settings to `settings.properties`cf. e.g. `de.tudarmstadt.ukp.inception.conceptlinking.config.EntityLinkingProperties`.

PR #516, opened 2018-09-11
#513 - Make SPARQL query limit configurable by user

MdeBoer95:


MdeBoer95:
@naveen2507 @reckart 
When I try to use the ``KnowledgeBaseProperties`` like the ``EntityLinkingProperties`` via **@ Resource** in e.g. ``SPARQLQueryStore`` I have the problem that the **@ Resource** can not be used in a static context.
This could be solved by making the ``SPARQLQueryStore`` and ``RDFUtils`` a spring component instead of classes with static methods.
Should I do that or is there a better way to solve this ?

reckart:
> This could be solved by making the SPARQLQueryStore and RDFUtils a spring component instead of classes with static methods.

For 0.6.x, it shouldn't be an issue because the setting is picked up from the `KnowledgeBase` entity anyway.

For 0.5.x, I would suggest to add a limit parameter to the methods in `SPARQLQueryStore` and `RDFUtils` - I expect the code that is calling the methods would not be static and could thus access the `KnowledgeBaseProperties`.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -461,22 +465,23 @@ public KBHandle createProperty(KnowledgeBase kb, KBProperty aProperty)
         });
     }
 
-    @Override
-    public Optional<KBProperty> readProperty(KnowledgeBase kb, String aIdentifier)
+    @Override public Optional<KBProperty> readProperty(KnowledgeBase kb, String aIdentifier)
```
reckart: `@Override` should be on a line by its own.
`kb` -> `aKB`.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -511,7 +516,8 @@ public void deleteProperty(KnowledgeBase kb, KBProperty aType)
         throws QueryEvaluationException
     {
         List<KBHandle> resultList = read(aKB, (conn) -> {
-            String QUERY = SPARQLQueryStore.PROPERTYLIST_QUERY;
+            String QUERY = SPARQLQueryStore
```
reckart: `QUERY` is not a constant (`final`). All capital names should only be used for constants and enum values (which are constants).

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/config/KnowledgeBaseProperties.java
```diff
@@ -0,0 +1,38 @@
...
+package de.tudarmstadt.ukp.inception.kb.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("inception.knowledge-base")
+public class KnowledgeBaseProperties
+{
+    private int sparqlQueryResultLimit = 1000;
```
reckart: Setting should be documented in the asciidoc documentation.

PR #529, opened 2018-10-01
Feature/299 support sub property labels

MdeBoer95:


reckart:
@MdeBoer95 needs resolving conflicts

reckart:
@MdeBoer95 when ready, please request review from @naveen2507

naveen2507:
@reckart can u ask @ukp-svc-jenkins to run a test.

reckart:
@naveen2507 you are whitelisted - you should be able to ask yourself.

naveen2507:

naveen2507:
@reckart Have u added `getPath() in de.tudarmstadt.ukp.clarin.webanno.ui.core.menu.MenuItem`? What's it for?

reckart:
The SPARQL endpoint `http://zbw.eu/beta/sparql/gnd/query` (OWL Schema) works again and uses various sub-properties of `http://www.w3.org/2000/01/rdf-schema#label`. So this can be tested again now.

reckart:
@MdeBoer95 @naveen2507 there are a number of unit test failures here

naveen2507:
@reckart we checked the PR on Friday and its still needs some work with transitive sub-property labels. @MdeBoer95 can we wrap this up today? Or we will move it to 0.6.1

MdeBoer95:
@naveen2507 I tried to solve it by adding the "+" or "*", but it didn't work for some reason. Can you have a look at it? If there is a simple fix for it we can wrap it up today. Otherwise we would need to move it.

naveen2507:
@reckart After having a closer look at the code, here is what I propose: 

- Adding transitive sub property as a part of query w/o inferencer is not the possibility. 
- Therefore I would take the other approach of extracting subProperty of rdfs:label and store it in a List variable for KB
- @MdeBoer95 implementation takes the route of `read(..)` which we would be refactoring as it takes too many queries to process. 
- We have already changed the part for KBConcept in PR #581 

I would suggest t move this PR to 0.6.1 for now.

reckart:
I have had the impression that a path-query along the sub-property IRI might work, something like "give me all <subject, predicate, value> where predicate sub-property* label-property". I guess you have tried that, but why didn't it work?

naveen2507:
Works Now.

naveen2507:
> I have had the impression that a path-query along the sub-property IRI might work, something like "give me all <subject, predicate, value> where predicate sub-property* label-property". I guess you have tried that, but why didn't it work?

There was a problem else where. Although the issue for language thing still prevails, as the prefered name values are mostly without language tags.

reckart:
> Although the issue for language thing still prevails, as the prefered name values are mostly without language tags.

That is https://github.com/inception-project/inception/issues/588, right?

naveen2507:
> That is #588, right?

Yup. But I was looking at the possible solution for #588. We need to have a query which says that if we get a variable with a specific language tag, we get that otherwise we return labels without language tag. Right? Do u have any pointers for #588

reckart:
I'll reply in #588.

naveen2507:
Issue with the `*` part of the query, the same query works for KB like Yago, GND, STW, etc but fails for Wine ontology.  Issue is dependent on #588  as PR #588 will change the structure of the query and then this issue can be worked out.

naveen2507:
@reckart the approach here doesn't work in general because using a * in the query makes the query quite expensive as it has to look for the sub property labels every time. Here is what I suggested when we started this PR. When we read the list of properties, we also read the subProperty relation once and store that as part of the KB configuration/object. WDYS?

reckart:
> When we read the list of properties, we also read the subProperty relation once and store that as part of the KB configuration/object. WDYS?

What do you suggest to be the life-time of this cached data?

naveen2507:
> What do you suggest to be the life-time of this cached data?

Yeah, could be and that's the issue. 

If not, then we can have a list of sub property query separate from the individual query. WDYS?

reckart:
> If not, then we can have a list of sub property query separate from the individual query. WDYS?

It is basically the same question: when do we do that query and how long do we retain the information obtained by it?

naveen2507:
> It is basically the same question: when do we do that query and how long do we retain the information obtained by it?

It would basically be the same as schema IRI's and we can do the query once before initializing the KB page.

reckart:
So we perform the query every time the user switches from one KB to another in the KB page (if the project has multiple KBs)? I guess that might work. But this would not be part of this PR anymore, right?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -616,14 +622,57 @@ public void deleteProperty(KnowledgeBase kb, KBProperty aType)
...
+    {
+        Set<KBHandle> subPropertyLabels = new HashSet<KBHandle>();
+        KBHandle labelHandle = new KBHandle(aKB.getLabelIri().stringValue());
+        getIterativeSubPropertyLabels(subPropertyLabels, aKB,
+                labelHandle);
+        return subPropertyLabels;
+       
+    }
+    
+    public Set<KBHandle> getIterativeSubPropertyLabels(Set<KBHandle> subPropertyLabelSet,
```
reckart: You mean `getSubPropertyLabelsRecursively`?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1105,12 +1165,15 @@ else if (identifierKBObj.get() instanceof KBInstance) {
...
             tupleQuery.setBinding("pLABEL", aKB.getLabelIri());
             tupleQuery.setBinding("pDESCRIPTION", aKB.getDescriptionIri());
+            tupleQuery.setBinding("pSUBPROPERTY", aKB.getSubPropertyIri());
             tupleQuery.setIncludeInferred(false);
 
             return evaluateListQuery(aKB, tupleQuery, false, aAll, "s");
         });
 
-        resultList.sort(Comparator.comparing(KBObject::getUiLabel));
+        if (resultList.size() > 1) {
```
reckart: Why extra check - if the list is empty or contains a single element, then the sort is a NOP - performance hit should be minimal.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1365,10 +1433,16 @@ public boolean statementsMatchSPO(KnowledgeBase akb, KBStatement mockStatement)
...
                 }
             }
-            else if (!stmts.hasNext()) {
-                Optional<KBInstance> kbInstance = readInstance(aKb, aIdentifier);
-                if (kbInstance.isPresent()) {
-                    return kbInstance.flatMap((p) -> Optional.of(p));
+            else {
+                Optional<KBProperty> kbProperty = readProperty(aKb, aIdentifier);
+                if (kbProperty.isPresent()) {
+                    return kbProperty.flatMap((p) -> Optional.of(p));
```
reckart: Isn't this a NOP? Can you please add a comment why this is necessary? Same a few lines below.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
```diff
@@ -69,34 +72,87 @@ private static final String optionalLanguageFilteredValue(String aProperty, Stri
...
+        fragment.append(languageFilterForVariable(aVariable, aLanguage));
+        fragment.append(" }\n");
+        return fragment.toString();
+    }
+  
+    /**
+     * adds an OPTIONAL block which looks for a value that is declared with a
+     * subproperty of the given property
+     */
+    private static final String optionalSubPropertyLabelFilter(Set<KBHandle> labelProperties,
```
reckart: The method name sounds as if the returned value is the label of a sub-property. But my understanding is that the query retrieves the label taking sub-properties of the label IRI into account. Also, it doesn't seem to be a "filter" because it doesn't really filter anything. Can we not simply all it `queryForOptionalLabel()`.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
```diff
@@ -69,34 +72,87 @@ private static final String optionalLanguageFilteredValue(String aProperty, Stri
...
     /** 
      * Query to list all instances from a knowledge base.
      */
-    public static final String listInstances(KnowledgeBase aKB)
+    public static final String listInstances(KnowledgeBase aKB, Set<KBHandle> labelProperties)
     {
         return String.join("\n"
                 , SPARQL_PREFIX
-                , "SELECT DISTINCT ?s ?l ?labelGeneral WHERE {"
+                , "SELECT DISTINCT ?s ?l ?labelGeneral ?spl WHERE {"
```
reckart: We have `?l` (label with language) and `?labelGeneral` (label without language). Why do we need an additional `?spl`? IMHO the `queryForOptionalLabel` should be overlapping with `optionalLanguageFilteredValue`, so let's bind `queryForOptionalLabel ` to `?l` and drop `?spl`.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/ConceptInstancePanel.java
```diff
@@ -134,17 +136,31 @@ public void actionStatementChanged(AjaxStatementChangedEvent event)
...
+                statement.getInstance().getIdentifier());
+        if (kbInstance.isPresent()) {
+            instanceHandle.setName(kbInstance.get().getName());
+        }
         selectedInstanceHandle.setObject(instanceHandle);
         event.getTarget().add(this);
     }
 
+    private boolean isRenamingEvent(KBStatement aStatement)
+    {
```
reckart: Can you please add JavaDoc here - what is this method doing? The argument is a statement, not an event - so I don't understand how you can check if the statement is a renaming event.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
naveen2507: Oh yeah.. Sorry . Will change that.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
naveen2507: yeah, that wud be better.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
naveen2507: then the problem would be that it can at times ignore the label present with rdfs:label and take the label provided with its sub property label. Thats why I did it separate (for the worst case scenario)

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/ConceptInstancePanel.java
naveen2507: No idea on that. @MdeBoer95 what was the logic behind this method and do we still need that?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
reckart: Why would it ignore the rdfs:label?

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/ConceptInstancePanel.java
MdeBoer95: It determines whether the event is about renaming a specific Instance. So yes it would make more sense to pass an event as a parameter here and then get the statement from that event.
I think we still need that because without it changing the label might not update the Instance name at the top of the Panel. 
But the ``isSubpropertyLabel`` method in the ``kbService`` needs to be updated since it doesn't take transitive subproperties into account right now.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/SPARQLQueryStore.java
naveen2507: Because if we take the same variable, that would be a possibility. For now, in the list of sub property label list we do not include label.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/ConceptInstancePanel.java
naveen2507: Ok. Yeah. I would need to do some changes here.

naveen2507: @MdeBoer95 can u please guide me to why exactly do we need this in context of sub property labels here. Also,  can you please add a java doc to the method here.

MdeBoer95: Whenever a label of a selected instance is changed we want to update the ``selectedInstanceHandle``object. So each time a statement is changed, we need to check whether this change involves the label of the selected instance. Since the label could now also be a subproperty label and not only a direct label like ``rdfs:label``, we need to adjust the check method (``isRenamingEvent()``)  to take subproperty labels into account .

naveen2507:
Jenkins, Can you test this please?

reckart:
@MdeBoer95 could you please review the latest changes from @naveen2507 and say if you consider this good for merging or not?

MdeBoer95:
@reckart I'm on it but I have some problems running the application, see gitter.

MdeBoer95:
There are some rendering issues because the ```KBInstance.read```, ```KBConcept.read```, ```KBProperty.read``` don't look for subproperty labels. I will fix it. Shouldn't take too much time.

MdeBoer95:
There are some rendering issues because the ```KBInstance.read```, ```KBConcept.read```, ```KBProperty.read``` don't look for subproperty labels.
These methods also add the statements that are used to define the Concept to the ```originalStatments``` list (which is a field in ```KBConcept```, etc.).
I don't know how I can extract the propertyIRI of the subpropertylabel in the current implementation. It would be necessary in order to add the corresponding statement to the ```originalStatements```  list.
I am also not sure why exactly we need this list. It is used for updating Concepts by deleting the ```originalStatements``` and then writing everything new again. I think we could also have a method that queries for all statements where the identifier is the subject when we want to update something instead of using this list.

Rentier:
We use the list for reification. In wikidata reified, one kbobject can consist of more than one statement. The subject there is not the subject, but the statement identifier.

reckart:
@MdeBoer95 looks like we got conflicts because of the KB refactoring PR. Could you please have a look at them?

MdeBoer95:
https://github.com/inception-project/inception/blob/00014ae41da9c3569e1d7e5323a04bb317d226d7/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java#L654-L661

@reckart why exactly do we need to check ```  if (!hasImplicitNamespace(id) && id.contains(":")) ```  ?

reckart:
If checks if the identifier is a "blank node" - i.e. one that does not have a proper IRI. We currently do not support blank nodes.

MdeBoer95:
Can be reviewed/tested.
For manual testing:
local KB:
1. Create a new property **A**
2. Create a Statement ``` A  subPropertyOf label``` for **A** (where label is e.g. rdfs:label) 
3. Set a new label for a Concept/Instance/Property by using the subproperty label **A** <br> (e.g. ConceptX A "NewLabel")
If there is no other label defined for the Concept/Instance/Property the subproperty label should be shown in the tree/list panel and in the heading.

remote KB:
GND can be used.
See if  Concepts/Instances/Properties are displayed with correct labels.
(e.g. Person names are defined with *Prefered Name for Person* which is a subproperty of label)

Issue #248, opened 2018-06-04
Elevated KB properties


MdeBoer95:
A small set of important KB properties should always be positioned at the top of the statement group list. The statements could be subClassOf, label and description (tbd).
- [ ] add discreet visual highlighting to these statement groups
- [ ] if these statements are missing, show placeholders of statement groups, which directly open a statement editor when clicked

PR #530, opened 2018-10-02
#248 - elevated kb properties

MdeBoer95:
- highlight statement group for certain properties

### DIFF
--- a/inception-ui-kb/pom.xml
+++ b/inception-ui-kb/pom.xml
@@ -161,6 +161,10 @@
       <artifactId>wicket-spring</artifactId>
     </dependency>
     <dependency>
+      <groupId>de.agilecoders.wicket</groupId>
+      <artifactId>wicket-bootstrap-core</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.wicketstuff</groupId>
       <artifactId>wicketstuff-annotation</artifactId>
     </dependency>
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.html
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.html
@@ -54,7 +54,7 @@
             <div class="statement-editor" wicket:id="statement"></div>
           </div>
         </div>
-        <div class="statement-group-footer">
+        <div wicket:id="statementGroupFooter" class="statement-group-footer">
           <button wicket:id="add" type="button" class="btn btn-link pull-right"><i class="fa fa-plus" aria-hidden="true"></i> <wicket:container wicket:id="label"></wicket:container></button>
         </div>
       </div>
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.java
@@ -27,6 +27,7 @@ import java.util.stream.Collectors;
 
 import org.apache.wicket.Component;
 import org.apache.wicket.ajax.AjaxRequestTarget;
+import org.apache.wicket.behavior.AttributeAppender;
 import org.apache.wicket.event.Broadcast;
 import org.apache.wicket.markup.html.WebMarkupContainer;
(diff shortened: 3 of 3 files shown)


MdeBoer95:
@reckart I changed the type of the Comparator for the statementGroups from Comparator < StatementGroupBean > to ImportantStatementComparator. Do you think that's ok or do you see any problems?
I needed to access the important-Function of the ImportantStatementComparator for the highlighting of the statementGroups.
https://github.com/inception-project/inception/blob/34e4c998b1b0856cb2109655babe32e6ce9c3931/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/stmt/StatementGroupPanel.java#L283-L288

reckart:
> @reckart I changed the type of the Comparator for the statementGroups from Comparator < StatementGroupBean > to ImportantStatementComparator. Do you think that's ok or do you see any problems?

@MdeBoer95 that's one option - However, I would suggest to extract the coloring into a separate class. Also `getImportant()` would not be sufficient anyway. So having a `StatementColoringStrategy` interface and for the start a `DefaultColoringStrategyImpl` as an implementation with methods such as `getColor(statement)` would IMHO be preferable. 

In WebAnno, I am using a method to automatically determine the text color based based on the background color (i.e. what `getColor()` returns) - the below is JavaScript. I would suggest to port it to Java and use it for deciding on the text color automatically:

```
// WEBANNO EXTENSION BEGIN - #820 - Allow setting label/color individually
// http://24ways.org/2010/calculating-color-contrast/
// http://stackoverflow.com/questions/11867545/change-text-color-based-on-brightness-of-the-covered-background-area
    var bgToFgColor = function(hexcolor) {
      var r = parseInt(hexcolor.substr(1,2),16);
      var g = parseInt(hexcolor.substr(3,2),16);
      var b = parseInt(hexcolor.substr(5,2),16);
      var yiq = ((r*299)+(g*587)+(b*114))/1000;
      return (yiq >= 128) ? '#000000' : '#ffffff';
    }    
// WEBANNO EXTENSION END
```

reckart:
@MdeBoer95 `getImportant()` would not be sufficient, because we want to assign different colors for different types of important statements, e.g. one color for labels, one for descriptions, etc.

MdeBoer95:
Would we implement this similar to the ValueTypes by having a ``StatementColoringStrategyRegistry`` that looks for implementations of ``StatementColoringStrategy`` which support a given property?
```java
        for (ValueTypeSupport s : getValueSupports()) {
            if (s.accepts(aStatement, aProperty) || s.accepts(range, null)) {
                support = s;
                supportCache.put(datatype, s);
                break;
            }
        }
...
```
Or would the Implementation of StatementColoringStrategy handle this by e.g. returning a color for a given property. ``getColor(String aPropertyIdentifier)``

reckart:
@MdeBoer95 we would use a registry if we want the coloring strategies to be pluggable by modules. We could do that. But I have to admit, I didn't have that in mind yet - I was thinking more along the lines of a hard-coded set of strategies to start with (cf. `de.tudarmstadt.ukp.clarin.webanno.api.annotation.coloring.ColoringStrategy`). If you like, we can already go for a registry :)

MdeBoer95:
Ok, then we go for a registry.

MdeBoer95:
@naveen2507 I can't compile the project:
Error:(48, 60) java: package de.tudarmstadt.ukp.inception.recommendation.api.type does not exist
Error:(93, 55) java: cannot find symbol
               symbol:   class PredictedSpan
  location: class de.tudarmstadt.ukp.inception.recommendation.imls.stringmatch.StringMatchingRecommender

This also happens for me on master. Do you know how to fix it?

naveen2507:
@MdeBoer95 Update the maven project including the jars and check if it works

Rentier:
@MdeBoer95  You need to add the `jcas` folder in recommender-api as a generated source folder from what I remember.

MdeBoer95:
Works again now. Thanks.

naveen2507:
@MdeBoer95  works for me. Although I didn't like the color for label in general. @reckart any comments?

MdeBoer95:
@naveen2507 feel free to change it. Just put another hexcode for the frame color in LabelColoringStrategyImpl.

reckart:
@naveen2507 @MdeBoer95 I think that the pastel tones in general are a good idea. There may be room for improvement, but for the moment, I'd just merge the PR so we can wrap up the release.

Issue #543, opened 2018-10-09
Unit tests fail for Hucit due to SPARQL service being unavailable


munnellg:
The unit tests in `de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceRemoteTest` fail when running `mvn clean compile package` due to the Hucit SPARQL service being unavailable. I am not sure if this is due to the service being unavailable in my area (I have tried on several different networks) or because the service itself is down. Either way, the unavailability of the service arguably should not cause the entire unit test to fail and result in maven terminating the build as this failure is not due to the INCEpTION code base.

reckart:
@munnellg Thanks for the report. We have multiple tests which check against remote KBs. This is not the best way in principle, but we want to be able to check that our tests really work against the latest live systems. 

@mromanello Probably we shouldn't test against your SPARQL endpoint? I guess it is quite a bit less stable than say Wikidata or DBPedia. WDYT?

@Rentier @naveen2507 Any comments?

reckart:
@munnellg If you want to be bleeding edge, you could also get pre-compiled binaries of the latest automatic builds from our integration branch (master) from here: https://zoidberg.ukp.informatik.tu-darmstadt.de/jenkins/job/INCEpTION%20(GitHub)%20(master)/

Rentier:
We could check in the test whether it is available or not and depending on that run the tests or not and print a warning.

reckart:
@Rentier sounds good. We can use the JUnit `assume` mechanism to conditionally skip the test if the endpoint is not there.

munnellg:
I tend to build from source as a way to get to grips with a new project. It's more of a personal process than a need to be on the bleeding edge, but thanks for the link!

Let me know if these issue reports get spammy. Just trying to flag things as I spot them. Seems like a good idea to have a record of them somewhere, even if the long term plan is not to address them.

reckart:
@munnellg we're definitely happy go be getting your issue reports! The last two issues you reported are already fixed.

For a true bleeding-edge build of the INCEpTION master, you also need a local checkout of the master branch from the WebAnno repo since the two co-evolve. E.g. I had to fix the two issues you have reported before in the WebAnno codebase, even if they didn't really directly affect the WebAnno application itself.

It is good to receive all issues via the INCEpTION tracker - we can then triage and decide whether they are to be fixed in the INCEpTION code itself or in upstream code.

reckart:
[...]

PR #545, opened 2018-10-10
#543 - Unit tests fail for Hucit due to SPARQL service being unavailable

MdeBoer95:
- assume that remote sparql endpoint is available before running tests against it

### DIFF
--- a/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/util/TestFixtures.java
+++ b/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/util/TestFixtures.java
@@ -17,6 +17,8 @@
  */
 package de.tudarmstadt.ukp.inception.kb.util;
 
+import static org.junit.Assume.assumeTrue;
+
 import java.io.IOException;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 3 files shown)


inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/util/TestFixtures.java
```diff
@@ -140,4 +143,23 @@ public KBQualifier buildQualifier(KBStatement kbStatement, KBHandle propertyHand
...
+     */
+    public boolean isEndpointAvailable(String aEndpointURL, int aTimeout) {
+        try {
+            URLConnection connection = new URL(aEndpointURL).openConnection();
+            connection.setConnectTimeout(aTimeout);
+            connection.connect();
+            return true;
+        }
+        catch (IOException e) {
+            e.printStackTrace();
```
reckart: A stack trace is a bit of an overkill here. You can print (or log) a simple message here that the endpoint is not available. But actually, I'd suggest to rename this to `assumeEndpointIsAvailable(...)` and instead of returning false, you just call `assume` at the end using the signature of the method which allows including a message. In that message you can include the URL of the endpoint as well as the reason for failure from the exception message. That way the information why the test was skipped directly ends up having all the details.

reckart:
@naveen2507 please review and merge when ready.

naveen2507:
yeah , will merge today.

Issue #547, opened 2018-10-11
Split dashboard into user and admin areas


reckart:
**Is your feature request related to a problem? Please describe.**
Right now, the dashboard has user-related and administration-related stuff all-in-one.

**Describe the solution you'd like**
 It would be better to set up separate areas for users and administrators/project managers where specific functionalities can be found.

PR #548, opened 2018-10-11
#547 - Split dashboard into user and admin areas

reckart:
- Started the split by moving user management and project management to a separate page

- [x] The link to the help page next to the user name is broken
- [x] When there are no projects, one should display a text like "You have no projects yet", else this white page looks as if I did something wrong


Rentier:
It looks really nice, but there are some minor issues:
- When I click on my user name, it does not redirect me to the profile page
- The link to the help page next to the user name is broken
- When there are no projects, one should display a text like "You have no projects yet", else this white page looks as if I did something wrong
- The liquibase complains about "characters" not being an object when using an embedded database:

```
Caused by: liquibase.exception.DatabaseException: user lacks privilege or object not found: characters [Failed SQL: update annotation_type set 
      anchoring_mode = "characters" 
      where lockToTokenOffset = false and multipleTokens = false]
```

reckart:
@Rentier the "characters" issue is not related to this PR - it will happen on master as well. It is related to https://github.com/webanno/webanno/issues/393

reckart:
@Rentier other half of the DB problems are covered via the changeset refactoring in https://github.com/inception-project/inception/pull/561

reckart:
Fixed non-working link to the user profile via https://github.com/webanno/webanno/commit/6130a88f0eb045c7415c97b7da9b8f5f00455c95

Essentially, the link should only be enabled if profile self-service is enabled. (`user.profile.accessible=true` in `settings.properties`).

reckart:
@Rentier Ok - all your comments should have been addressed now. Mind that one of the DB-related fix is in #561 some other fixes were in upstream code.

reckart:
@Rentier @naveen2507 any more feedback / change requests? Otherwise, I'd I'm looking forward to approval so this can be merged.

Issue #157, opened 2018-05-03
Enable full-text search on internal KBs


reckart:
For internal KBs, we should enable full-text search. I guess we can just always enable it instead of making it an option. 

http://docs.rdf4j.org/programming/#Full_text_indexing_with_the_Lucene_SAIL

The problem here is that we cannot wrap the NativeStoreConfig with a LucenesailConfig. rdf4j-storage still relies on Lucene 6.x and not Lucene 7.x which we use. There is already an open PR for upgrading to Lucene 7.x (eclipse/rdf4j-storage#78) that needs to be waited for.

----
### Todos

- [x] @reckart: Fork rdf4j and rdf4j-storage repos
- [x] @reckart: Create a new branch called " inception-master" based on the master branch, protect it, set it as the default for the repo
- [x] @rcffc: Change group ID
- [x] @rcffc: Upgrade Lucene/Solr version to same as MTAS (currently 7.3.0)
- [x] @reckart: Set PR build and main build Jenkins
- [x] @rcffc: Testing
- [x] @rcffc: Find out which properties were used 
- [x] @reckart: update documentation

rcffc:
- After importing an internal KB, query all instances of the database and store each in the Lucene Index.
- Create a LuceneSail in KnowledgeBaseService.
- Modify Entity Linking queries.
- KB should be indexed again whenever it is updated.

reckart:
> KB should be indexed again whenever it is updated.

Adding the LuceneSail in the KnowledgeBaseService for local KBs should take care of everything. I.e. when data is added/updated/removed, the SAIL should automatically keep the indexes up-to-date.

However, search might not work for local KBs that were created before adding the LuceneSAIL.

rcffc:
@reckart I am getting this error during startup, although the dependencies should be there: 
```Caused by: java.lang.IllegalStateException: Failed to introspect Class [de.tudarmstadt.ukp.inception.kb.KnowledgeBaseServiceImpl] from ClassLoader [sun.misc.Launcher$AppClassLoader@18b4aac2]
	at org.springframework.util.ReflectionUtils.getDeclaredMethods(ReflectionUtils.java:659) ~[spring-core-5.0.4.RELEASE.jar:5.0.4.RELEASE]
	at org.springframework.util.ReflectionUtils.doWithMethods(ReflectionUtils.java:556) ~[spring-core-5.0.4.RELEASE.jar:5.0.4.RELEASE]
[...]

PR #561, opened 2018-10-14
#157 - Enable full-text search on internal KBs

reckart:
- Changed formulation a bit
- Fixed Liquibase changeset for HSQLDB
- Still need to update screenshot of the KB settings ...


reckart:
@Rentier if you approve ("looks good") you should also press the approve button :)

Issue #552, opened 2018-10-12
Duplicates in concept feature editor


reckart:
**Describe the bug**
Concept feature editor seems to show duplicates (similar to #440)

**To Reproduce**
1. Set up KB based on the UKP-internal Wikidata SPARQL endpoint; **enable concept linking support**
2. Create a named entity annotation
3. Search for something

**Expected behavior**
Every item should be shown only once unless there are really multiple IRIs.

**Screenshots**
![2018-10-12_17-03-25](https://user-images.githubusercontent.com/1410238/46877322-dca79880-ce40-11e8-9a12-019289d17857.png)

<img width="506" alt="2018-10-12_17-15-03" src="https://user-images.githubusercontent.com/1410238/46878001-699f2180-ce42-11e8-9c74-ed39d2f906d5.png">

**Please complete the following information:**
 - Version and build ID: INCEpTION 𝜶 -- 0.6.0-SNAPSHOT (2018-10-12 16:39:08, build bc3be70efa774a0b1a7223910603e2584163ab86)

reckart:
In order to debug this better, it would be good to implement #553 first

naveen2507:
@reckart PR #564 takes #553. I am not able to replicate the issue. Can you please check it once more. If you have the same issue or was it a different IRI?

reckart:
I can reproduce it with this project. For instructions, see the project description ;)

[%28ISSUE+%23552%29+Duplicates+in+concept+feature+editor_2018-10-15_1157.zip](https://github.com/inception-project/inception/files/2478244/28ISSUE%2B.23552.29%2BDuplicates%2Bin%2Bconcept%2Bfeature%2Beditor_2018-10-15_1157.zip)

naveen2507:
The issue comes up from the line in `getConcepts()` and `getInstances()` method in `ConceptFeatureEditor.java` : 
`handles.addAll(listLinkingInstances(kb.get(), aState,
                        () -> getEditorCas(aHandler), aTypedString));` 

It doesn't distinguish between a concept and instance when `kb.get().isSupportConceptLinking()` is true.

naveen2507:
The issue prevails only in the case if we enable support concept linking.

PR #566, opened 2018-10-15
#552 Duplicates in concept feature editor

naveen2507:
- Fix to remove duplicates

### DIFF
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
@@ -24,6 +24,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
```diff
@@ -289,6 +291,17 @@ protected void onInitialize() {
     public Component getFocusComponent() {
         return focusComponent;
     }
+    
+    private List<KBHandle> distinctByIri(List<KBHandle> aHandles)
+    {
+        Map<String, KBHandle> hMap = new HashMap<>();
```
reckart: Potentially changes the order of the list in unpredictive ways. Not sure if this is an issue, but it could be prevented by using a LinkedHashMap().

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
```diff
@@ -289,6 +291,17 @@ protected void onInitialize() {
...
+        for (KBHandle h: aHandles) {
+            if (!hMap.containsKey(h.getIdentifier())) {
```
MdeBoer95: Is this *if* necessary? Iris would still be distinct if we just use *hMap.put(h.getIdentifier(), h)*, since HashMap doesn't allow duplicate keys.

naveen2507: yeah, its not. actually, I was playing around with some else if but removed it later on.

PR #569, opened 2018-10-16
#547 - Split dashboard into user and admin areas

reckart:
- Allow access to project settings directly from project dashboard


Rentier:
@reckart Can you tell me what that actually changed? I searched but do not see it when running INCEpTION.

reckart:
>  Allow access to project settings directly from project dashboard

* Go to the project overview
* Select a project which will take you to the project dashboard
* There is now a new item "Settings" in the menu sidebar of the project dashboard which takes you directly to the project page with the current project preloaded. In this view, the project list and project import panel should not be visible. However, the latter (list/import invisible) may depend on changes in WebAnno which Jenkins might still be building/deploying as we write.

Rentier:
I can go to the settings from a project. But for me the project is then not preloaded/preselected.

reckart:
@Rentier for me, it works

reckart:
I think the help item may still not be working - checking.

reckart:
Broken help link was an upstream issue. Fixed.

Issue #577, opened 2018-10-24
Unable to add root concept to KB


reckart:
**Describe the bug**
When I add an explicit root concept to a KB in the project settings, save, switch somewhere else and come back, the root concept it gone. It is also not used on the KB page.

There is no error message and I also did not notice any error in the log.

**To Reproduce**
1. Go to a KB
2. Add an explicit root concept
3. Press save button
4. Press cancel button
5. Select the KB again
6. Check if the root concept is still there (in my case, it is not)

**Expected behavior**
The concept should still be there.

**Please complete the following information:**
 - Version and build ID: 4ef41a7884ae10fe7389f030de09b69937ab2582

naveen2507:
Tested. The issue exists for the concepts which doesn't have label and description field.

reckart:
I can add the root concept now, but the label of the root concept on the KB page is now off.

<img width="686" alt="2018-10-24_13-45-21" src="https://user-images.githubusercontent.com/1410238/47450932-bf6eb480-d7c6-11e8-9ff5-4a5218791625.png">

As you can see in the screenshot, the label *should* be "Authority Resource", but it is for some reason "GND Subject Categories Value" in the tree and also in the class detail editor. The problem seems to affect only the explicitly defined root concept. The children seem to have the correct labels.

naveen2507:
@reckart I have changed the querying function, which works now. Can you please post the KB configuration for the above screenshot, so I can test it or may be you can check if it works for you now?

reckart:
@naveen2507 the root concept has the correct label now, but the properties no longer show up - only a line "unable to list properties".

Here is a screenshot of the KB settings so you can try yourself:

![2018-10-25_10-57-46](https://user-images.githubusercontent.com/1410238/47488536-e075ea80-d844-11e8-8b46-92971ab46d75.png)

naveen2507:
@reckart I can see the properties listed. Can't replicate the property listing issue.

![image](https://user-images.githubusercontent.com/10420950/47490928-005bdd00-d84a-11e8-9038-0c769a5b5da2.png)

![image](https://user-images.githubusercontent.com/10420950/47490976-179aca80-d84a-11e8-9217-5a79b2825844.png)

reckart:
@naveen2507 seems to work for me as well now. Might have been a glitch.

naveen2507:
DONE

PR #581, opened 2018-10-24
#577 Unable to add root concept to KB

naveen2507:
- Changed the method to read the concept
- Currently reading concepts take 3 queries which can be replace by 1 query. 
- Issue: Concepts which do not have any statements associated with it, is not read by the current implementation 
- If everyone is okay with this approach, we can then remove the readConcept()


reckart:
@naveen2507 Jenkins still fails:

```
[ERROR] /home/svc_jenkins/workspace/INCEpTION (GitHub) (PR builder)/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseService.java:573: error: @param name not found
[ERROR] * @param kb
[ERROR] ^
```

reckart:
> If everyone is okay with this approach, we can then remove the readConcept()

If the original `readConcept()` is no longer used and we remove it, how about then changing the name of the new `readConceptExist()` to `readConcept()`?

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -445,21 +445,61 @@ public KBHandle createConcept(KnowledgeBase kb, KBConcept aConcept)
...
+                }
+                Binding label = bindings.getBinding("l");
+                Binding description = bindings.getBinding("d");
+                handle.setIdentifier(aIdentifier);
+                if (label != null) {
+                    handle.setName(label.getValue().stringValue());
+                    if (label.getValue() instanceof Literal) {
+                        Literal literal = (Literal) label.getValue();
+                        Optional<String> language = literal.getLanguage();
+                        language.ifPresent(handle::setLanguage);
```
MdeBoer95: Can we move this part to ``evaluateListQuery(...)`` and then call ``evaluateListQuery(...)`` inside ``readConcept(...)`` to avoid redundant code?

naveen2507:
Added language from retrieved labels through SPARQL query to the KBHandle in `evaluateListQuery(...) `

Issue #586, opened 2018-10-26
Unable to configure KB without language


reckart:
**Describe the bug**
If a KB does not make use of language-specific labels, it must be possible to configure it without a language. However, right now the language is mandatory.

**To Reproduce**
1. Try deleting the value from the language field of the KB settings
2. Save

**Expected behavior**
Empty value should be saved.

**Please complete the following information:**
 - Version and build ID: 85d08aea4fe58433b749f387e45e796de471c1e6

PR #587, opened 2018-10-26
#586 - Unable to configure KB without language

reckart:
- Made the language field not mandatory
- Removed unnecessary LambdaAjaxFormComponentUpdatingBehavior
- Ensure that feedback messages are shown if the form is not valid


reckart:
@naveen2507 good for merge?

Issue #590, opened 2018-10-26
Sentence-level recommender


reckart:
**Is your feature request related to a problem? Please describe.**
Add a recommender for sentence-level annotations.

**Describe the solution you'd like**
The OpenNLP Document Categorizer should serve well for a start.

PR #591, opened 2018-10-26
#590 - Sentence-level recommender

reckart:
- Basic implementation of a recommender based on the OpenNLP Document Categorizer

### DIFF
--- a/inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderTest.java
+++ b/inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderTest.java
@@ -86,7 +86,7 @@ public class OpenNlpDoccatRecommenderTest
 
         assertThat(context.get(OpenNlpDoccatRecommender.KEY_MODEL))
             .as("Model has been set")
-            .isNotNull();
+            .isPresent();
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderTest.java
```diff
@@ -0,0 +1,233 @@
...
+    public void thatTrainingWorks() throws Exception
+    {
+        OpenNlpDoccatRecommender sut = new OpenNlpDoccatRecommender(recommender, traits);
+        List<CAS> casList = loadDevelopmentData();
+
+        sut.train(context, casList);
+
+        assertThat(context.get(OpenNlpDoccatRecommender.KEY_MODEL))
+            .as("Model has been set")
+            .isNotNull();
```
Rentier: I think that this test does not make sense, as you return now an Optional

inception-imls-opennlp/src/test/resources/de/tudarmstadt/ukp/dkpro/core/api/datasets/lib/sentence-classification-en.yaml
```diff
@@ -0,0 +1,56 @@
+groupId: org.dkpro.core.datasets.sentenceclassification
```
Rentier: Why do we need this file？

reckart: It is used to download the data used in the unit test.

inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderTest.java
reckart: Was copy-pasted from the other test - probably needs to be changed everywhere then.

reckart:
@Rentier good for merge?

reckart:
@Rentier I tested it in the application and it seems to work in general. However, whether it works well or not is a different question (and obviously depends on the task). The accuracy of 0.85 for the argumentative zoning task from the unit test looks nice at a first glance, but there might be a majority class (MISC). Anyway, it is not the purpose of this issue to log more evaluation details or to optimize the results - just to provide the basic functionality.

So, I'll merge it once the checks are good.

Issue #554, opened 2018-10-12
Features can use disabled KBs


reckart:
**Describe the bug**
A concept linking feature may refer to a KB which is disabled.

**To Reproduce**
1. Create a KB
2. Create a concept linking feature and choose the previously created KB in the "Knowledge base" dropdown
3. Disable the KB (uncheck the "enabled" checkbox)
4. Switch back to the concept linking feature - the disabled KB is still listed

**Expected behavior**
If a concept linking feature links to a specific KB and that KB is disabled, then I would expect the feature to stop working. I would say this is expected behavior (not sure if this is actually the case though). However, the user should be warned in the concept feature editor as well as in the feature editor that the KB is disabled and the feature therefore does not work.

**Screenshots**

**Please complete the following information:**
 - Version and build ID: INCEpTION 𝜶 -- 0.5.2 (2018-10-04 15:37:48, build 08f5facb646bb5c007962f4cf37e6daa68db2bbb)

naveen2507:
Works

PR #595, opened 2018-10-28
#554 - features can use disabled kbs

MdeBoer95:
- add warning label in conecept feature editor
- display warning in concept feature settings


MdeBoer95:
@reckart How can I add a warning label for the fact linking outside of the ``QualifierFeatureEditor``,``SubjectObjectFeatureEditor``, etc.? 
So that there is only one label at the bottom and not one for each editor.
Is there a Webanno class or something?

reckart:
@MdeBoer95 If I understand the question correctly, I guess you'd have to add the warning label inside the `<div wicket:id="content">` in the HTML files of the two editors.

MdeBoer95:
No, I mean I want to add the label to the surrounding Panel.

reckart:
You could add the marker in the title section of the panel:

```
<div class="panel-heading">
      <h3 class="panel-title">
        <span wicket:id="feature"></span> **HERE**
      </h3>
    </div>
```

... but it would need to be a very small marker. If I saw it right, you currently add a rather longish message explaining the problem. I would suggest instead to use a small warning icon (e.g. https://fontawesome.com/v3.2.1/icon/warning-sign) and put a tooltip on it with the explanatory text.

MdeBoer95:
> I would suggest instead to use a small warning icon (e.g. https://fontawesome.com/v3.2.1/icon/warning-sign) and put a tooltip on it with the explanatory text.

makes sense

> You could add the marker in the title section of the panel:

But wouldn't I need to add the Label/Icon for every Editor in that case?
If I just use a small Icon it might be okay. I just wanted to avoid displaying a long message multiple times.

reckart:
> But wouldn't I need to add the Label/Icon for every Editor in that case?
If I just use a small Icon that might be okay. I just wanted to avoid displaying a long message multiple times.

Right - every editor to which the problem applies would exhibit a warning. Mind that an annotation could have multiple editors, some to which the problem applies and some to which it does not apply.

reckart:
The issue number referenced in the commits has a typo - it should be #554, not #544

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
```diff
@@ -165,6 +175,25 @@ private JCas getEditorCas(AnnotationActionHandler aHandler) throws IOException {
         return handles;
     }
 
+    private boolean featureUsesDisabledKB(ConceptFeatureTraits aTraits)
+    {
+        Optional<KnowledgeBase> kb = null;
```
Rentier: Please do not assign `null` to an `Optional`, use `Optional.empty()`

naveen2507:

naveen2507:
@MdeBoer95 I dont get any warning message, while I look over the annotations from the KB which has been disabled. Although debugging shows that it goes through  `add(createDisabledKbWarningLabel());` in  `ConceptFeatureEditor.java`

naveen2507:
Works.

Issue #594, opened 2018-10-28
Load CASes during training/prediction/evaluation lazily


reckart:
**Describe the refactoring action**
Currently, the CASes are loaded preemptively at the beginning of training/evaluation/prediction tasks. However, there are many cases where the CASes are loaded but then not even used, e.g. if there is only one recommender which does not require evaluation or if all recommenders are disabled, etc. Thus, it would be better to load the CASes only when they are actually needed. The easiest way to accomplish this is to leave the code largely as-is, but to load the CASes only when they are used for the first time.

**Expected benefit**
In cases the CASes are not required, the training/evaluation/prediction tasks can complete much faster.

PR #597, opened 2018-10-29
#594 - Load CASes during training/prediction/evaluation lazily

reckart:
- Implemented the lazy loading

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
@@ -86,36 +86,17 @@ public class PredictionTask
         Predictions model = new Predictions(project, getUser());
         List<SourceDocument> documents = documentService.listSourceDocuments(project);
 
-        for (SourceDocument document : documents) {
-            JCas jCas;
-            try {
-                jCas = documentService.readAnnotationCas(document, user.getUsername());
-                annoService.upgradeCas(jCas.getCas(), document, user.getUsername());
-            } catch (IOException e) {
-                log.error(
-                        "Cannot read annotation CAS for user [{}] of document "
-                                + "[{}]({}) in project [{}]({}) - skipping document",
-                        user.getUsername(), document.getName(), document.getId(), project.getName(),
-                        project.getId(), e);
-                continue;
-            } catch (UIMAException e) {
-                log.error(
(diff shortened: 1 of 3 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -141,42 +122,71 @@ public void run()
...
+                        }
+                        try {
+                            annoService.upgradeCasIfRequired(jCas.get().getCas(), document,
+                                    user.getUsername());
+                        }
+                        catch (UIMAException | IOException e) {
+                            log.error("Cannot upgrade annotation CAS for user [{}] of document "
+                                    + "[{}]({}) in project [{}]({}) - skipping document",
+                                    user.getUsername(), document.getName(), document.getId(),
+                                    project.getName(), project.getId(), e);
```
Rentier: I think the error `e` will not be shown, because it is at the end of some var args.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
```diff
@@ -141,42 +122,71 @@ public void run()
...
+                            continue nextDocument;
+                        }
+                    }
+                    
                     try {
-                        recommendationEngine.predict(ctx, jCas.getCas());
+                        recommendationEngine.predict(ctx, jCas.get().getCas());
```
Rentier: I would `get` the Jcas once and assign it to a variable, that looks nicer

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
reckart: If the last argument of an SFL4J log call is an exception and if there is no `{}` covering it, then it is logged as a stack trace.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/PredictionTask.java
reckart: Might defy the lazy evaluation because then I introduce a fixed evaluation point again. It might work in this case, but I'd prefer not having to worry about it. That's why I just call `get()` whenever I might need the value.

Issue #560, opened 2018-10-14
Entity search only finds concepts with labels


reckart:
**Describe the bug**
When a concept does not have a label (as is the case for most of the concepts in the Wine ontology), they cannot be found via the local Lucene-based indexing.

**To Reproduce**
1. Import wine ontology (local FTS enabled)
2. Try linking via a Named Entity annotation
3. Only "wine" and "Wine ontology" are shown in auto-complete field

**Expected behavior**
~~All items that show up on the KB page should also be retrievable via the auto-complete field (assuming that it is configured correspondingly)~~ (see https://github.com/inception-project/inception/issues/560#issuecomment-477513570)

- [x] KB creators should always include labels for their KB items ( classes/instances/properties) and the INCEpTION documentation should make explicit that these labels are mandatory in order to find the items.
- [x] it should be possible to search for items by entering their **full** IRI.

**Please complete the following information:**
 - Version and build ID: e93744a9a169c83a8226a4fef6098447a88395c1

-----

>>> @reckart When I import the wine ontology into a local KB (OWL mapping) with FTS enabled and create a layer with a KB-linked feature that allows classes/instances from any KB/scope - I still only get auto-completions for "Wine Ontology" and "wine", but for nothing else from the KB.

>> @rcffc In the Wine ontology, these two are the only items that have used `rdfs:label`. I am having a look at the build.

> @reckart I see. I wonder if we can maybe hook into/configure the LuceneSail indexing process somehow so that it also allows us to index the IRI suffixes that we show in the UI if no label is available.

@reckart Looks like that is not easily possible. We'd have to make changes directly to the RDF4J Lucene indexing code. The code already stores an "URI" field in the lucene index, but that cannot be queried - at least not via SPARQL as the analyzer used to parse the query tries to remove special characters. We'd get something like this when trying

```
ERROR LuceneIndex - There was a problem evaluating query 'uri:"http://mbugert.de/pets#animal"' for property 'null!
java.lang.IllegalStateException: field "uri" was indexed without position data; cannot run PhraseQuery (phrase=uri:"http mbugert.de pets animal")
```

Rentier:
There is a property `INDEXEDFIELDS` which we might be able to leverage:

```
Defining the indexed Fields
The property INDEXEDFIELDS is to configure which fields to index and to project a property to another. Syntax:
 # only index label and comment
 index.1=http://www.w3.org/2000/01/rdf-schema#label
[...]

PR #602, opened 2018-10-29
#560 - Local full text search only finds concepts with labels

rcffc:
- also return concepts without label
- if there is no label, use iri as label
- simplified queries by throwing out unnecessary things. renaming "e2" to "iri"


Rentier:
Tests are failing. Can you also add unit tests for your code?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
```diff
@@ -164,15 +164,23 @@ private static String getFullTextMatchingQueryPartDefault(int aLimit)
...
-            "    SELECT DISTINCT ?e2 ?altLabel ?description WHERE",
+            "    SELECT DISTINCT ?iri ?altLabel ?description WHERE",
             "    {",
-            "      ?e2 search:matches ?match .",
-            "      ?match search:query ?string ;",
-            "             search:property ?labelIri ;",
-            "             search:snippet ?altLabel",
+            "      {  ?iri ?a ?concept .",
+            "          FILTER NOT EXISTS { ?iri ?labelIri [] }",  // only concepts without label
+            "          FILTER regex( str(?iri) , ?string, \"i\") ",  // case-insensitive matching
```
reckart: Here, we are no longer using full-text search. Using `regex` does not really scale. Have you tried this with large-scale KBs like Wikidata on Virtuoso?

rcffc: This query snippet is only used for local knowledge bases that are indexed by Lucene.

Rentier:
It works for me: I deleted the label from an instance and it could be still found. @reckart Could you try it also? You opened the bug.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/util/QueryUtil.java
reckart: Ok, but Lucene dos not speed up `regex` - so even for local KBs, this will be quite slow unless the KB is really small. Using `regex` is *not* a full text query as the title of the method suggests. The old code was using the full-text index from Lucene, but this new code does not.

Rentier:
I close this PR as we might need to change things in rdf-storage directly.

Issue #588, opened 2018-10-26
Support labels/descriptions without languages even if KB has language


reckart:
**Is your feature request related to a problem? Please describe.**
It may be the case that the KB does not consistently make use of the possibility to associate languages with string literals. This is currently a problem because if a KB is configured to use a certain language, then labels without a language are not considered, even if there is no label with the desired language - and we might just get no reasonable label in that case.

**Describe the solution you'd like**
If the KB has a language configured and if a label in that language is available, we should use it. However, if there is no label in the configured language but there is one without a language, then we should use that. Only if there is no label, we should fall back to extracting one from the URI.

INCEpTION 𝜶 -- 0.6.0-SNAPSHOT (2018-10-26 12:36:10, build 85d08aea4fe58433b749f387e45e796de471c1e6)

reckart:
> @naveen2507:
> We need to have a query which says that if we get a variable with a specific language tag, we get that otherwise we return labels without language tag. Right? Do u have any pointers?

Hm. If remember correctly, we currently include the language filter for the optional label/description in the query - and then we use a `min` to remove duplicates.

What I was thinking is that we could include two language filters in the query, like:

* give me all labels with language X
* give me all labels without a language

Then we apply `min` to each of them separately to remove duplicates within the group of labels with a language and within the group of labels without a language.

Then we could change `evaluateListQuery` such that it would prioritize in the following way
- use the label with language (if available)
- use the label without language (if available)

PR #603, opened 2018-10-29
#588 Support labels/descriptions without languages even if KB has lang

naveen2507:
- Add language support for labels and description when the language for
entity is not specified

- Removed description field for queries where not required


naveen2507:
@reckart the approach gives us timeout exception for Wikidata queries. I guess its overly burdened with the OPTIONAL query?

naveen2507:
> @naveen2507 which queries to time out? get children, get parents, etc.? Is it all of them or specific ones?

Only the root concepts one. You can check the list in the failed tests.

reckart:
@naveen2507 I see three options:

1. we make the support for (transitive) sub-properties optional
2. we document the fact that the root concept detection may time out for large KBs and recommend that people define explicit roots for large KBs
3. we change the way that root concepts are retrieved into a two-phase process: first a simpler query which retrieves **only** the subject IRIs of the root concepts, but no labels/descriptions, etc. Then we run a second query where we retrieve the labels/descriptions of the root concepts.

What do you think? Does any of them make more/less sense to you? A there alternatives?

naveen2507:
> we make the support for (transitive) sub-properties optional

We don't have that in this PR yet. 

> we document the fact that the root concept detection may time out for large KBs and recommend that people define explicit roots for large KBs

Can be an option but it open up the vulnerability of the platform. 

> we change the way that root concepts are retrieved into a two-phase process: first a simpler query which retrieves only the subject IRIs of the root concepts, but no labels/descriptions, etc. Then we run a second query where we retrieve the labels/descriptions of the root concepts.

I was thinking of this option. Suits to me the best for our case.

reckart:
wrt sub-properties: right, was mixing up things here.

So let's try improving the root detecting then by turning it internally into a two-phase process.

naveen2507:
@reckart any idea why is the test failing for `MTASIndexTest.java` for webanno dependedncy.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -1111,19 +1177,76 @@ else if (identifierKBObj.get() instanceof KBInstance) {
...
+            if (labelGeneral != null) {
+                handle.setName(labelGeneral.getValue().stringValue());
+                if (labelGeneral.getValue() instanceof Literal) {
+                    Literal literal = (Literal) labelGeneral.getValue();
+                    Optional<String> language = literal.getLanguage();
+                    language.ifPresent(handle::setLanguage);
+                }
+            }
+            
+            // needs to be fixed
```
reckart: What needs to be fixed here?

naveen2507: Nothing. Its an old comment when I was working on it. Will remove this.

naveen2507:
#529 Dependent on this PR

naveen2507:
@reckart can u review this PR and close. It would impact the other issue of sub property labels as well.

naveen2507:
Jenkins, can you test this please ?

reckart:

naveen2507:
@reckart there is a bit of inconsistency with YAGO SPARQL point. I am checking it. For now its running and test cases have passed. You can review this PR now.

inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImpl.java
```diff
@@ -924,11 +930,61 @@ public KBHandle update(KnowledgeBase kb, UpdateAction aAction)
...
+        List<KBHandle> resultLabelList = readLabelsWithoutLanguage(aKB, aAll, resultList);
+        resultLabelList.sort(Comparator.comparing(KBObject::getUiLabel));
+        return resultLabelList;
+    }
+    
+    private List<KBHandle> readLabelsWithoutLanguage(KnowledgeBase aKB, boolean aAll,
+            List<KBHandle> resultList)
+    {
+        List<KBHandle> resultLabelList = new ArrayList<>();        
+        for (KBHandle result : resultList) {
```
reckart: This loop here generates **many** queries. E.g. when used on `http://zbw.eu/beta/sparql/gnd/query` it takes really long to load. Do you think it would be possible to send a single query where all IDs for which we want to have labels are included - or at least to batch-query for like 100 labels at a time?

naveen2507: Yeah, I thought about it. Ideally when u have an IRI present for specific concept/instance, it's better comparatively. But yeah, there is overhead of queries in here. You wanna put in some pointers here?

reckart: @naveen2507 What kind of pointer would you like?

naveen2507: I meant, do u hv some alternative approach in mind??

reckart:
From testing the code, seems ok.

The ` // needs to be fixed` comment is still there.

Issue #239, opened 2018-05-28
Log search actions


reckart:
Add action logs for in-project queries.

betoboullosa:
Must look at ActiveLearningRecommendationEvent

PR #604, opened 2018-10-29
#239 - Log search actions

reckart:
- Added query logging


reckart:
@Rentier @naveen2507 Jenkins gives thumbs up - approval still required.

Issue #593, opened 2018-10-27
Add PDFAnno sources


wpmp:
**Is your feature request related to a problem? Please describe.**
Currently pdf files are only processable in brat view.
There is a pdf view using pdf.js which does not support annotation of the document.

**Describe the solution you'd like**
PDFAnno is a JavaScript application for annotating pdf files in the browser.
Inception should be extended to provide annotation support for pdfs.
This will be achieved by modifying and using PDFAnno in the pdf view.
Therefore PDFAnno sources need to be added to the inception project.

**Describe alternatives you've considered**
An alternative annotation module is hypothes.is.
However it does not provide support for relation annotations, which PDFAnno does.

PR #606, opened 2018-10-30
#593 - add pdfanno sources

wpmp:
Adds PDFAnno sources to inception-pdf-editor and replaces previous PDF.js


inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
```diff
@@ -0,0 +1 @@
+SERVER_PORT=1000
```
reckart: What server is meant here?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
```
reckart: Shouldn't this be at the root of the project?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.eslintrc.js
```diff
@@ -0,0 +1,59 @@
+// http://eslint.org/docs/user-guide/configuring
```
reckart: Is it important that this is a hidden file? I'm afraid people working on the project may miss it.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/docs/.vuepress/config.js
```diff
@@ -0,0 +1,11 @@
+module.exports = {
```
reckart: Another hidden file?

reckart:

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
wpmp: I can't tell you exactly as I just copied the PDFAnno code over to inception.
There existed a server component in PDFAnno in version 0.4.1, however it was removed and not available in 0.5.0.
Maybe they forgot to delete this file.
Should I remove it?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.eslintrc.js
wpmp: According to the ESLint Configuration page this seems correct.
But the page states that it could also be added to a field `eslintConfig` within package.json file.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
reckart: I guess if we don't need it, we should not have it in the codebase. I hope that we can remove more code that we don't need later - once we figured out that we don't need it ;)

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.editorconfig
wpmp: This is another file copied over from PDFAnno.
So I am not sure, I guess it should stay where it is as it applies to PDFAnno source only, and not the rest of this project.
I could also remove it, what do you think?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/docs/.vuepress/config.js
wpmp: According to VuePress Configuration page this seems also correct.
I don't see alternatives exception other file formats (`yml` or `toml`).
However, as this documentation for pdfanno does not contain much and also some wrong information, I could remove the complete `docs` folder.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
wpmp: I can do that.
I was not sure whether the whole project should be copied over or only what is really required.
Should I remove unnecessary stuff in this issue / PR or should I open a new one?
I am still a bit uncertain about how small or large an issue / PR should be. :)

reckart: This is less about how large/small a PR should be but rather about not having useless files lying around. So we should not commit/merge code/resources that we don't use.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.editorconfig
reckart: It looks like some configuration file for the source code formatter in IntelliJ. @Rentier any idea? I think you added this kind of file somewhere as well.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/docs/.vuepress/config.js
reckart: Removing the docs folder sounds sensible.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.editorconfig
Rentier: It is a cross IDE specification in how your project source code should be formatted. It should be unimportant.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
wpmp: What is the best approach now to get rid of the files we don't want?
Should I just make another commit to this branch where the files are removed or is there a nicer and cleaner way?

Rentier: We can squash merge this pull request at the end I think.

reckart:
@wpmp there seems to be much too much stuff being packaged into the `inception-pdf-editor-0.6.0-SNAPSHOT.jar` (size 147.138K) - it seems the entire node server ends up in there.

I believe the maven-frontend-plugin should be set up in such a way that it builds the JS code somewhere **outside** of `src/main/resources` / `src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources` and then just copy the things that are actually needed for running the stuff into `src/main/resources`/ `src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources` so that it can be served at runtime.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources/.env.example
reckart: @wpmp what @Rentier said ;) Normally we do not squash, but in this case, I'd certainly squash - yes.

wpmp:
@reckart I removed unnecessary files and moved pdfanno code from `src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources` to `src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/src`.
On generating resources only the `dist` folder ends up in`src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/resources`.
The file `inception-pdf-editor-0.6.0-SNAPSHOT.jar` should now be much smaller, around 1.4M

reckart:
Can we move the PDFanno stuff maybe to some place outside the Java folders like `src/main/js`?

reckart:
Optimally, the "compiled" JS code should to to some place like `target/js` and then we could add this `target/js` folder as an additional resource folder (https://maven.apache.org/plugins/maven-resources-plugin/examples/resource-directory.html).

wpmp:
I think this should be no problem.
So PDFAnno sources end up in `src/main/js/pdfanno` I guess.
If the compiled code should end up in the `target` folder, I should also move the maven phase for compiling it from `generate-resources` to `compile`, shouldn't I?

reckart:
> So PDFAnno sources end up in src/main/js/pdfanno I guess.

I have never put JS code that is compiled into a Maven project, but it sounds good.

> If the compiled code should end up in the target folder, 

In a subfolder of the target folder, yes. The compiled Java stuff ends up in `target/classes`. Generated resource usually end up somewhere like `target/ generated-resources/XXX` or `target/generated-sources/XXX`.

> I should also move the maven phase for compiling it from generate-resources to compile, shouldn't I?

I believe you can choose either - I do not think it will make a difference.

reckart:
@wpmp on a local build, I still see a huge number of resource files being processed (copied into the JAR) by Maven:

```
[INFO] --- maven-resources-plugin:3.0.2:resources (default-resources) @ inception-pdf-editor ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.
[INFO] skip non existing resourceDirectory /Users/bluefire/git/inception-application/inception-pdf-editor/src/main/resources
[INFO] skip non existing resourceDirectory /Users/bluefire/git/inception-application/inception-pdf-editor/src/main/resources
[INFO] Copying 24376 resources
```

Commit: 3a7b7343e088e376eaaa152c7a1d70db73dd257d

reckart:
Probably most of these files are in the `de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/src` folder inside the JAR.

wpmp:
@reckart you are right.
I must have made a mistake while checking this, sorry.
I will fix this.

reckart:
> It seems resource can only be appended to parent pom.xml if another plug-in is used.
Is the solution I chose ok?

I assume you are referring to the Maven Build Helper Plugin - yes, can be used - no problem.

Having the output in `target/generated-resources` would be preferable because then it wouldn't go to the git repo (we wouldn't need an git exclude for the folder).

wpmp:
@reckart on `generate-resources` phase, pdfanno source is now copied to `target/js/pdfanno` and afterwards compiled in this folder.
This has the benefit that `build`, `node` and `node_modules` folder don't get created within `src/main/js/pdfanno` but only in the `target` folder.
I removed those folders from `.gitignore`.
I hope everything is fine now.

wpmp:
I removed `cmaps`, `locale` and `images` folders in `pdfanno/src` as they are duplicates and already exist in `pdfanno/pages`.
They also are never used when compiling and are useless.
I think this is all for now that I can remove from PDFAnno.
I hope everything is fine now.

reckart:

wpmp:
@reckart Why did it fail?

reckart:

reckart:
@wpmp good question - the build server seems to have had trouble installing node.js. I'm triggering the build again in case it was just a transient network issue.

reckart:
There is a file which Jenkins chokes on when trying to get the test coverage statistics: 

```
./inception-pdf-editor/target/js/pdfanno/node_modules/async-limiter/coverage/coverage.json
```

reckart:
See: https://github.com/jenkinsci/github-pr-coverage-status-plugin/issues/37

reckart:

PR #617, opened 2018-11-05
#577 - Switching between projects does not hide kb details panel

MdeBoer95:
- set selected model object to null when project selection changes ( KbPanel and RecommendersPanel)


MdeBoer95:
@naveen2507 can be reviewed

Issue #622, opened 2018-11-06
No projects on overview page


reckart:
**Describe the bug**
Project overview page only works for project managers, not for regular users.

**To Reproduce**
1. Log in as a user who has only annotator permissions
2. See that there are no projects to choose from

**Expected behavior**
All projects in which the current user has any role should be listed.

**Please complete the following information:**
 - Version and build ID: 9e1e11082da6482c09541dd8a7a7fd00e34c8bea

PR #623, opened 2018-11-06
#622 - No projects on overview page

reckart:
- Overview page should list all accessible projects, not only manageable projects.


naveen2507:
Tests are failing. I am working on #588 , it will fix the issues with test case

Issue #631, opened 2018-11-08
RDF files with relative IRIs cannot be imported


reckart:
**Describe the bug**
Some RDF files may only contain relative IRIs - but RDF4J does not allow importing relative IRIs. The best practice seems to be to resolve the IRIs against the URL from which the RDF file was downloaded.

**To Reproduce**
1. Import an RDF file containing relative IRIs

```
2018-11-08 14:23:11 ERROR [rec] ParseErrorLogger - [Rio fatal] Not a valid (absolute) IRI: #CF350B291A038871A435028D4E41B7BCA45DEC57672C6275F5B2D6C75EE8 (1, -1)
```

**Expected behavior**
Since we do not download RDF files but import them from disk, the best solution seems to be to use the base prefix configured in the KB settings to resolve the relative IRIs.

**Screenshots**

**Please complete the following information:**
 - Version and build ID: 0.6.1-SNAPSHOT (8d94d73460b192d90645b6e50668f032bcb248d0)

PR #632, opened 2018-11-08
#631 - RDF files with relative IRIs cannot be imported

reckart:
- Use the prefix configured in the KB to resolve relative IRIs


Rentier:
You could write a test for this.

reckart:
> You could write a test for this.

I have considered it but considering the other todos on my list, it fell through. Do you believe testing this is important?

Issue #639, opened 2018-11-11
Refactoring KB Module - Change UI structure


MdeBoer95:
**Describe the refactoring action**
- Change the structure of the UI. 
- KnowledgeBaseCreationWizard and KnowledgeBaseDetailedPanel use the same panel and we have a redundant task for simpler modifications.
 Eg. Use a type step panel or something and use it everywhere, that way we don't have redundant work
- remove view mode in project settings
- should be done before https://github.com/inception-project/inception/issues/598

This is a part of #572.

**Expected benefit**
Make modifications easier and faster.

PR #641, opened 2018-11-11
#639 - Refactoring KB Module - Change UI structure

MdeBoer95:


reckart:
Changing this to 0.7.x (master) because the PR is against master already and master has already been merged into this PR branch.

reckart:
@MdeBoer95 where are we with this one?

MdeBoer95:
@reckart I have a strange bug and I don't know why it happens. When I create a knowledge base and then switch to the KBPage the kb appears empty. When I then go to the kb settings again, click on save and then go back to the KBPage it works. 
The rest is done.

reckart:
@MdeBoer95 is it done from your perspective? Should this now be reviewed/tested?

reckart:
@MdeBoer95 Changing the full text mode setting doesn't work for me. If I change it, leave the page and come back it always resets to "No full text search support".

MdeBoer95:
@reckart fixed it.

Issue #611, opened 2018-10-31
Add new project via the project overview page


reckart:
The project overview page only lists the projects. It would be convenient for users who can create projects if they could do that directly from that page as well without first having to go into the admin area.

Issue #651, opened 2018-11-13
Import project via the project overview page


reckart:
It would be nice if project managers could import projects directly from the project overview page.

reckart:
Included in #611.

reckart:
Also, the import via the project overview currently criples the project somewhat unexpectedly (although documented).

* as a user trying out the tool - when importing an example project (even as "admin"), the only role I have in the project after import is that of a project manager. Consequently, the annotation and curation menu items are not available to me unless I give me the respective permissions. This is quite inconvenient for somebody who just wants to try out example projects.
* as a project manager - I want to create a clone of a project, i.e. to export it and then to re-import it and give it a new name. I can do this, but I loose all the permissions, even if the users exist in the instance.

reckart:
> as a user trying out the tool with the default "admin" account ...

Fixed via https://github.com/webanno/webanno/pull/1121 - permissions of the importing user are always imported, even if general permission import is not requested/possible.

reckart:
If I import an example project via the project list page as "someuser" (i.e. not as "admin" by whom most example projects were created), then I only get the "manager" role now. I have to explicitly give myself the "annotator" and "curator" roles as well. This is an extra hurdle.

Probably it would be good to give the importing user all permissions and let the user remove those again that the user doesn't want.

reckart:
WebAnno commit https://github.com/webanno/webanno/commit/96a4950b897c9f7fe6d2f3b0844da1b08b2aecf8 adds all roles for a user importing a project via the INCEpTION project overview page.

PR #662, opened 2018-11-15
#611 - Add new project via the project overview page

reckart:
- Added button to create new project
- Added upload field to import projects
- Rearranged dashboard package structure a bit
- Reduce use of deprecated methods from SecurityUtil


reckart:

reckart:

reckart:
Upon reconsideration #651 is included in #611 - and it is not really possible to rename branches. Just ignore the branch name - it gets deleted on merge anyway.

reckart:
> I selected files to upload and then clicked to create a new project, but it did not add the files as documents in the new document.

Sorry, I didn't get it... the upload field titled "Browse project files to import ..." is used to import entire projects... not documents. Is the caption that misleading?

> I personally would not add the upload button in that screen, just the "Create new" and then hide uploading documents in the page opening when clicking on the button.

When you press "create new project", then a new project is created. If I hide the option to upload documents there, then there is no way of uploading documents to the project...?

Rentier:
project files for me sounds like documents
I could also multi select file and it did not warn me that I uploaded txt.
I would want 1 button in the project overview to add a project, there you can either import or create a new one.

reckart:
> project files for me sounds like documents

We could call it "Project archives"?

> I could also multi select file and it did not warn me that I uploaded txt.

It should be possible to import multiple projects in one go. I am surprised that you could even select text because the widget is configured to accept only zip files:

```
config.allowedFileExtensions(asList("zip"));
```

> I would want 1 button in the project overview to add a project, there you can either import or create a new one.

How would you do that? Importing a project requires an upload form and creating a project requires switching to the project page. How would you do those two things with a single button?

Rentier:
I can select it, but then it turns red. I did not see it the first time around. I would just rename the button to "Import projects", not "Browse files" and the label to "Import already existing projects" or so

reckart:
The "browse" button is really to browse. If I change it to "import", then it will be confusing because:

* it doesn't really import - it opens a browser files dialog
* after you have selected the files in the dialog, you still need to press the *upload* button :) I could maybe that to "Import" if that helps?

I would agree to changing the label to "Import project archives...". "Import existing projects" confuses me because if the project already exists, then why should I import it? ;)

Rentier:
I honestly would leave the import button in the admin area
A project to import existed before, else you could not have exported it.

reckart:
It might be that somebody sent me the project or that I downloaded it from the website. In that case, it never existed on my machine before.

I like having it on the projects lists page. If I am just a project creator and not an administrator, I find that more convenient.

So we have slightly different opinions. For the sake of reaching approval, how about I change the label to "Import project archives..." and we leave it at that for the time being and see how well it works out for us?

Rentier:
ok

Issue #642, opened 2018-11-11
Document cannot be indexed by Mtas, unable to read CAS


rcffc:
**Describe the bug**
Document cannot be indexed by Mtas because CAS unable to be read.
Search sidebar does not find anything.

```
2018-11-11 21:49:40 ERROR [admin] MtasUimaParser - Unable to decode CAS
org.apache.uima.cas.impl.XCASParsingException: Error parsing XCAS or XMI-CAS from source <unknown> at line 72, column 154: unknown type: de.tudarmstadt.ukp.clarin.webanno.api.type.CASMetadata.
	at org.apache.uima.cas.impl.XmiCasDeserializer$XmiCasDeserializerHandler.createException(XmiCasDeserializer.java:1635) ~[uimaj-core-2.10.2.jar:2.10.2]
	at org.apache.uima.cas.impl.XmiCasDeserializer$XmiCasDeserializerHandler.createException(XmiCasDeserializer.java:1657) ~[uimaj-core-2.10.2.jar:2.10.2]
	at org.apache.uima.cas.impl.XmiCasDeserializer$XmiCasDeserializerHandler.readFS(XmiCasDeserializer.java:490) ~[uimaj-core-2.10.2.jar:2.10.2]
	at org.apache.uima.cas.impl.XmiCasDeserializer$XmiCasDeserializerHandler.startElement(XmiCasDeserializer.java:409) ~[uimaj-core-2.10.2.jar:2.10.2]
	at com.sun.org.apache.xerces.internal.parsers.AbstractSAXParser.startElement(AbstractSAXParser.java:509) ~[?:1.8.0_152]
	at com.sun.org.apache.xerces.internal.parsers.AbstractXMLDocumentParser.emptyElement(AbstractXMLDocumentParser.java:182) ~[?:1.8.0_152]
	at com.sun.org.apache.xerces.internal.impl.XMLNSDocumentScannerImpl.scanStartElement(XMLNSDocumentScannerImpl.java:351) ~[?:1.8.0_152]
	at com.sun.org.apache.xerces.internal.impl.XMLDocumentFragmentScannerImpl$FragmentContentDriver.next(XMLDocumentFragmentScannerImpl.java:2784) ~[?:1.8.0_152]
	at com.sun.org.apache.xerces.internal.impl.XMLDocumentScannerImpl.next(XMLDocumentScannerImpl.java:602) ~[?:1.8.0_152]
[...]

PR #667, opened 2018-11-16
#642 - Document cannot be indexed by Mtas, unable to read CAS

reckart:
- Changed method of obtaining the type system to include internal types


reckart:
@naveen2507 You need this fix ;)

Issue #648, opened 2018-11-12
Error indexing zero-width annotations with MTAS


reckart:
**Describe the bug**
MTAS does not support indexing zero-width annotations.

**To Reproduce**
1. Create a zero-width annotation
2. Check logs:
```
2018-11-12 11:56:06 ERROR [SYSTEM] MtasUimaParser - Unable to create token collection
java.lang.IllegalArgumentException: Start offset after end offset
	at mtas.analysis.token.MtasToken.setOffset(MtasToken.java:422) ~[mtas-7.3.0.3.jar!/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.indexAnnotationText(MtasUimaParser.java:373) ~[inception-search-mtas-0.6.2-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.indexAnnotation(MtasUimaParser.java:309) ~[inception-search-mtas-0.6.2-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.createTokenCollection(MtasUimaParser.java:218) ~[inception-search-mtas-0.6.2-SNAPSHOT.jar!/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.createTokenCollection(MtasUimaParser.java:174) ~[inception-search-mtas-0.6.2-SNAPSHOT.jar!/:?]
	at mtas.analysis.MtasTokenizer.constructTokenCollection(MtasTokenizer.java:237) ~[mtas-7.3.0.3.jar!/:?]
	at mtas.analysis.MtasTokenizer.reset(MtasTokenizer.java:171) ~[mtas-7.3.0.3.jar!/:?]
	at org.apache.lucene.index.DefaultIndexingChain$PerField.invert(DefaultIndexingChain.java:732) ~[lucene-core-7.3.0.jar!/:7.3.0 98a6b3d642928b1ac9076c6c5a369472581f7633 - woody - 2018-03-28 14:31:58]
	at org.apache.lucene.index.DefaultIndexingChain.processField(DefaultIndexingChain.java:430) ~[lucene-core-7.3.0.jar!/:7.3.0 98a6b3d642928b1ac9076c6c5a369472581f7633 - woody - 2018-03-28 14:31:58]
[...]

PR #670, opened 2018-11-16
#648 - Error indexing zero-width annotations with MTAS

reckart:
- Skip indexing of zero-width annotations
- Document behavior
- Updated unit test to check that zero-width annotations are skipped


reckart:
@naveen2507 maybe you have a moment to review/approve?

Issue #674, opened 2018-11-19
"Show all" in recommender side bar is global, not per project


Rentier:
**Describe the bug**
When changing "Show all" for during annotating in one project, it is changed for all projects.

**To Reproduce**
1. Create two projects A and B
2. Open Project A, Open the recommender side bar, change the "Show all", remember the new state A
3. Open Project B, Open the recommender side bar, change the "Show all", remember the new state B
4. Open Project A, the state now is state B

**Expected behavior**
The state should be state A

**Please complete the following information:**
INCEPTION Master

PR #675, opened 2018-11-19
#674 - "Show all" in recommender side bar is global, not per project

reckart:
- Refactored RecommendationUserState to be simpler without internal maps
- Store preferences (including showAll flag) per project

Not fixing this in 0.6.x because I took the opportunity to also refactor the managing of the state internally a bit.

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
@@ -29,6 +29,9 @@ import javax.persistence.PersistenceContext;
 import org.apache.commons.collections4.MapIterator;
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
+import org.apache.commons.lang3.Validate;
+import org.apache.commons.lang3.builder.EqualsBuilder;
+import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.apache.uima.jcas.JCas;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -78,43 +81,39 @@
     private @Autowired RecommenderFactoryRegistry recommenderFactoryRegistry;
     private @Autowired RecommendationScheduler scheduler;
     
-    private Map<String, RecommendationUserState> states = new ConcurrentHashMap<>();
+    private Map<RecommentationStateKey, RecommendationUserState> states = new ConcurrentHashMap<>();
```
Rentier: RecommendationStateKey, not RecommentationStateKey

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -312,15 +304,57 @@ public void setFeatureValue(AnnotationFeature aFeature, Object aPredictedValue,
...
     }
 
+    private static class RecommentationStateKey
+    {
+        private final String user;
+        private final long projectId;
+        
+        public RecommentationStateKey(String aUser, long aProjectId)
+        {
+            super();
```
Rentier: Why do you need super?

reckart: Some people think (including Eclipse) that it is good style to always call super or this in constructors to make the initialization hierarchy explicit. Removed it (as I usually always do).

Issue #624, opened 2018-11-07
Integration of PDFExtractor


wpmp:
**Is your feature request related to a problem? Please describe.**
PDFAnno requires a *.txt.gz file which contains characters, including their positions, of a pdf file.

**Describe the solution you'd like**
PDFExtractor produces the *.txt.gz file for a pdf file.
Therefore PDFExtractor is integrated into inception pdf editor to provide PDFAnno the required file.

PR #678, opened 2018-11-19
#624 - integration of pdfextractor

wpmp:


inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
```diff
@@ -42,28 +48,55 @@
...
+
+    private final Logger log = LoggerFactory.getLogger(getClass());
     
     public PdfAnnoPanel(String aId, IModel<AnnotatorState> aModel)
     {
         super(aId, aModel);
 
+        SourceDocument doc = aModel.getObject().getDocument();
+
+        File pdfFile = documentService.getSourceDocumentFile(doc);
```
reckart: This is probably a problem because the PDF file name is obtained only when the PdfAnnoPanel is first created but not when e.g. switching from one document to another. I'd leave it were it was in `onRequest()` below.

inception-pdf-editor/src/main/js/pdfanno/src/page/pdf/PDFAnnoPage.js
```diff
@@ -613,37 +613,42 @@ export default class PDFAnnoPage {
       method : 'GET',
       mode   : 'cors'
     }).then(response => {
+// BEGIN PDFANNO EXTENSION - #624 - Integration of PDFExtractor
```
reckart: Within such a change-marking block, I'd recommend to keep the entire original code in a `/* */` block - that makes it easier to see what changed when merging in upstream changes. Cf. visualizer.js in WebAnno.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
wpmp: So should I get it twice then?
I need it in the second `onRequest()`, for pdfextract.

inception-pdf-editor/src/main/js/pdfanno/src/page/pdf/PDFAnnoPage.js
wpmp: I changed the extensions.
Is it fine like this?

wpmp:
@reckart I can't merge, but everything is approved or am I missing something?

reckart:
@wpmp No, that is correct. The set of people who can merge to master is currently restricted.

Issue #692, opened 2018-11-21
PDF module makes use of library not on Maven Central


reckart:
**Describe the refactoring action**
PDF module makes use of library not on Maven Central: pdfextract - need to add UKP OSS repo to POM.

**Expected benefit**
Being able to build without having to upload pdfextract to Maven Central for the time being.

PR #693, opened 2018-11-21
#692 - PDF module makes use of library not on Maven Central

reckart:
- Added UKP SNAPSHOT repo


reckart:
@Rentier actually - the SNAPSHOT repo is already defined in the pom (the PR now introduces a duplicate declaration which fails the build)...

So if your system fails obtaining the pdfextract SNAPSHOT there must be some other kind of problem.

Rentier:
It looked to me like it tries to go to scruffy internal and gets access denied (as I did not configure it), that is why I wondered.

Issue #696, opened 2018-11-22
Support document-level annotations


reckart:
**Is your feature request related to a problem? Please describe.**
In addition to the annotations on particular parts of the text, it would be good to have annotations that apply to the document as a whole.

**Describe the solution you'd like**
- [x] New type of layer (in addition to span/relation/chain): document metadata
- [x] Ability to view and edit document-level annotations via a left sidebar on the annotation page
- [x] Select a suitable icon for the sidebar

**Describe alternatives you've considered**
- In order to be able to distinguish between text-level annotations and document-level annotations without having to know details about the layer configuration, it seems sensible that the document-level annotations inherit from `AnnotationBase` and not from `Annotation`. Might have to be careful with chain head types though because they also inherit from `AnnotationBase`

**Additional context**

PR #697, opened 2018-11-22
#696 - Support document metadata

reckart:
PR aims to provide a basic support for document-level annotations, in particular

* New layer type for document metadata annotations
* Ability to view/create/edit/delete them in a sidebar on the annotation page

**How to test manually**
- Enable the feature by adding `documentmetadata .enabled=true` to the `settings.properties` file
- Create a project and import two texts (to test the reaction of the sidebar when switching between documents)
- Create at least two layers of the *Document metadata* type in the project settings
- Add one or more String features to each of the layers 
- Go to the annotation screen and open the metadata sidebar (it currently has the looking glass icon - still needs to be changed)
- Add annotations, modify annotations, delete them
- Switch between documents

It is known that there are still bugs that the UI doesn't properly update, there might even be some exceptions.


reckart:
@Rentier this is implemented as a MVP - still with some bugs/no documentation but it seems to me to be working well enough to get some user feedback on the functionality. Therefore I would like to merge this into 0.7.0. The feature is by default disabled and can be enabled by adding `documentmetadata .enabled=true` to the `settings.properties` file.

reckart:
> * Does granularity make sense when configuring DocumentMetaData?

No - it'll require more upstream refactoring to allow not displaying the granularity for certain layers...

> * When I switch between the layers, it does not update properly

What do you mean by "switching between layers"? The layer dropdown at the top is only to add new annotations of the selected type, not to switch between layers. The annotation list below the dropdown always shows all annotations from all metadata layers.

> * It feels strange that I do not know when it is saved. as there is no button for it

We always auto-save annotations - same as in the annotation detail panel on the right side. Having a save button is terribly annoying.

> * Required not checked?

No, I guess at the moment there is no validation at all for the metadata layers.

> * Cannot be searched (should they?)

There is no search support yet, but it should eventually be possible to search over the metadata annotations.

Issue #570, opened 2018-10-18
NPE during multi-document import


reckart:
**Describe the bug**
A NPE can be thrown when documents are imported. It might be this only occurs when importing multiple documents at once because I didn't observe it yet when importing a single one.

**To Reproduce**
1. Go to the project settings - document tab
2. Import multiple documents

**Expected behavior**
No exception

**Log**
```
2018-10-18 19:52:39 INFO [admin] DocumentServiceImpl - Loading initial CAS for source document [adi3767.conll](195) in project [Corpus: LeNER-BR](92)
2018-10-18 19:52:39 INFO [admin] Conll2002Reader - Scanning [file:/home/user/inception/repository/project/92/document/195/source/]
2018-10-18 19:52:39 INFO [admin] Conll2002Reader - Found [1] resources to be read
2018-10-18 19:52:39 INFO [admin] Conll2002Reader - 0 of 1: file:/home/user/inception/repository/project/92/document/195/source/adi3767.conll
2018-10-18 19:52:39 INFO [admin] CasStorageServiceImpl - Updated annotations for user [INITIAL_CAS] on document [adi3767.conll](195) in project [Corpus: LeNER-BR](92)
2018-10-18 19:52:39 INFO [admin] EventRepositoryImpl - LoggedEvent [created=Thu Oct 18 19:52:39 CEST 2018, event=AfterDocumentCreatedEvent, user=admin, project=92, document=195]
2018-10-18 19:52:39 INFO [admin] DocumentServiceImpl - Imported source document [adi3767.conll](195) to project [Corpus: LeNER-BR](92)
2018-10-18 19:52:39 ERROR [admin] TransactionSynchronizationUtils - TransactionSynchronization.afterCompletion threw exception
java.lang.NullPointerException: null
	at de.tudarmstadt.ukp.inception.search.scheduling.IndexScheduler.isIndexingDocument(IndexScheduler.java:191) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.scheduling.IndexScheduler.enqueue(IndexScheduler.java:112) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.scheduling.IndexScheduler.enqueueIndexDocument(IndexScheduler.java:78) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.SearchServiceImpl.afterDocumentCreate(SearchServiceImpl.java:372) ~[classes/:?]
```

**Please complete the following information:**
[...]

PR #702, opened 2018-11-23
#570 - NPE during multi-document import

reckart:
- Refactoring indexing scheduling code
- Seprate task classes for all indexing tasks
- Using instanceof for the moment instead of deducing task type from which fields are set
- Fix bad access to annotation document ID when dealing with a source document (cause of the NPE)
- Simplified detection of whether a given task is already scheduled


reckart:
More refactoring in the scheduler code is necessary, but not in this PR. Rather when centralizing the whole scheduler stuff into a scheduler module.

reckart:

Issue #706, opened 2018-11-24
External recommender tests fail on Windows


reckart:
**Describe the bug**
External recommender tests fail on Windows. There are probably multiple reasons. At least one of them is that Windows doesn't use UTF-8 as the default encoding which causes the payload to currently get garbled.

**To Reproduce**
* Run unit test on Windows.

**Expected behavior**
Should work.

PR #707, opened 2018-11-24
#706 - External recommender tests fail on Windows

reckart:
- Explicitly set charset in more places.

### DIFF
--- a/inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
+++ b/inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
@@ -20,9 +20,9 @@ package de.tudarmstadt.ukp.inception.recommendation.imls.external;
 import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -190,9 +191,11 @@ private String serializeTypeSystem(CAS aCas) throws RecommendationException
 
     private String serializeCas(CAS aCas) throws RecommendationException
     {
-        try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
-            XmiCasSerializer.serialize(aCas, null, out, true, null);
-            return new String(out.toByteArray(), UTF_8);
+        try (StringWriter out = new StringWriter()) {
+            XmiCasSerializer xmiCasSerializer = new XmiCasSerializer(null);
```
Rentier: I would add a comment here that we do not limit the serialization by the type system, the `null` else requires to read the UIMA code.

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -190,9 +191,11 @@ private String serializeTypeSystem(CAS aCas) throws RecommendationException
...
+            XMLSerializer sax2xml = new XMLSerializer(out, true);
+            xmiCasSerializer.serialize(aCas, sax2xml.getContentHandler(), null, null, null);
```
Rentier: Why do we needed to change this?

reckart: Need is relative. By avoiding the raw byte representation, we can (try to) avoid the encoding issue altogether.

Issue #649, opened 2018-11-13
Learning curve diagram


reckart:
**Is your feature request related to a problem? Please describe.**
We want to be able to see the performance of the recommenders configured on a given project in the recommendation sidebar.

**Describe the solution you'd like**
Display the performance as a "learning curve" diagram with time on the x axis and score on the y axis.

**Steps**
- [x] add a method to the RecommendationService to load RecommenderEvaluationResultEvent from the LoggedEvent table + unit test 
- [x] Decide on a JS library to use for rendering the chart
- [x] If a recommender is disabled in the project settings, then no line for it should be drawn
- [x] Create a new Wicket panel which loads the JS library and uses it to render the time/score information from the RecommenderEvaluationResultEvents retrieved in the first step + unit test
- [x] document the new feature in the user documentation (recommender section)
- [x]  on-mouse-over-infobox is drawing below the sidebar tab are, causing it to appear cut off. 
- [x] show only 4 digits behind the comma for scores

Sibgha360:
"t-50" to "t-0" takes a lot of space as compared to if we show sequence 50-0. The panel width is small for displaying categories on the x axis. So in the latest commit I showed 0-50 sequence. Should there be a better idea, I will change it.

The graph is done for the multiple recommenders. Thinking how to handle layers as well as recommenders!

reckart:
@Sibgha360 We said the multiple layers support should go to a separate issue / PR. So that means this PR should be good for review and probably merging now, right?

Sibgha360:
Yes. It is good for the review.

reckart:
> * Create a new Wicket panel which loads the JS library and uses it to render the time/score information from the RecommenderEvaluationResultEvents retrieved in the first step + unit test

@Sibgha360 The PR implements the new functionality directly in the RecommendationSidebar. Could you please move it to a separate `LearningCurveChartPanel` (java+html) which you then instantiate and add to the `RecommendationSidebar`. The panel should include only the chart, not the surrounding box, so the HTML template for the `LearningCurveChartPanel` is very basic and basically only contains the `chart-container` element.

These two still need to be done:

> * If a recommender is disabled in the project settings, then no line for it should be drawn
> * document the new feature in the user documentation (recommender section)

reckart:
@Sibgha360 can we hide the legend below the chart (i.e. the names of the recommenders) such that they only appear when we move the mouse over a line?

[...]

PR #708, opened 2018-11-24
#649 - Learning curve diagram

Sibgha360:
**What's in the PR**
- Added Learning Curve Diagram 

**How to test manually**
* Add show.learning.curve.diagram=true in the settings.properties
* Create project
* Add document.
* Add recommender (must be enabled and no-evaluation unselected)
* Annotate the document on the layer recommender was added.
* Learning curve will appear on the recommender sidebar

**Automatic testing**
* [x] PR includes integration tests of the DAO functions.

**Documentation**

**Todo**
- [x] Need to fix checkstyle
- [x] Adhere to DkPro Code Style
- [x] Fix error when there are no evaluations
- [x] Update documentation
- [x] Unit tests?
- [x] Add configuration setting to show chart (hidden by default)
- [x] Use recommender id to distinguish recommenders and show recommender name on the learning chart.
- [x] The specified number of points for each recommender must be plotted.


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -166,6 +167,16 @@ public void deleteRecommender(Recommender aRecommender)
         return settings;
     }
     
+    @Override
+    @Transactional
+    public List<LoggedEvent> listLoggedEvents(Project aProject, User user)
```
reckart: Let's put this method into the  `EventRepository` in the `inception-log` module. Also let's change it such that the type of event to be retrieved can be passed as a method argument.

I'd suggest to change the `User aUser` argument into a `String aUsername` argument because in that way we do not necessarily have to retrieve a `User` object before the method can be used.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.html
```diff
@@ -49,6 +49,8 @@ <h3 class="panel-title">Recommendation Information</h3>
         <input wicket:id="save" type="submit" class="btn btn-primary"  wicket:message="value:apply"/>
       </div>
     </form>
+	<div wicket:id="chart-container" id="chart-container-js"></div>
```
reckart: We must not set the HTML `id` attribute ourselves. Wicket has to do it and we can request this by calling `setOutputMarkupId(true)` on the associated Wicket component. That way we can ensure that we can put as many instances of a component on a page as we wish.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
                 aAnnotationPage.actionRefreshDocument(_target)));
         
         add(form);
-    }
+
+		chartContainer = new Label("chart-container" );
+		add(chartContainer);
+
+		add(new AbstractAjaxTimerBehavior(Duration.milliseconds(5000)) {
```
reckart: I assume the problem you are solving with the timer behavior is to ensure that the chart regularly updates itself as the user is performing actions. But instead on relying on an unconditional timer, I'd suggest we listen for events generated in the backed an/or hook into the rendering pipeline for the main editor window.

E.g. you can add a method like this to the sidebar and this is triggered whenever the main editor view is refreshed:

```
    @OnEvent
    public void onRenderAnnotations(RenderAnnotationsEvent aEvent)
    {
       ...
    }
```

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
+						try {
+							scoreDouble = Double.valueOf(detail.getScore());
+							listDetailJson.add(scoreDouble);
+						} catch (NumberFormatException e) {
+							log.debug("Skipping logged Event due to invalid score, Detail:{}", detail);
+							continue;
+						}
+
+					} catch (IOException e) {
+						log.error(e.toString(), e);
```
reckart: When in UI code, in addition to logging, please always also show the errors to the user:

```
                error("Unable to switch to document : " + e.getMessage());
```

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
+		aResponse.render(JavaScriptHeaderItem.forReference(C3JsReference.get()));
+		aResponse.render(JavaScriptHeaderItem.forReference(D3JsReference.get()));
+
+		//import Css
+		aResponse.render(CssHeaderItem.forReference(C3CssReference.get()));
+	}
+}
+
+//TODO: Move it to a separate file?
+@JsonIgnoreProperties(ignoreUnknown = true)
```
reckart: Let's re-use the class `de.tudarmstadt.ukp.inception.recommendation.log.RecommenderEvaluationResultEventAdapter.Details` instead of defining a new one.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
+						aModel.getObject().getUser());
+
+				// iterate over the logged events to extract the scores only.
+				List<Double> listDetailJson = new ArrayList<Double>();
+
+				for (LoggedEvent loggedEvent : loggedEvents) {
+					String detailJson = loggedEvent.getDetails();
+
+					ObjectMapper mapper = new ObjectMapper();
+					Double scoreDouble;
```
reckart: No reason to use `Double` as `null` should not be a valid value. Better use `double`.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
+							log.debug("Skipping logged Event due to invalid score, Detail:{}", detail);
+							continue;
+						}
+
+					} catch (IOException e) {
+						log.error(e.toString(), e);
+					}
+				}
+
+				String data = "";
```
reckart: Better use Jackson (i.e. `JSONUtil.toJson(...)`) to serialize the list to JSON.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -67,5 +91,98 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
+			private static final long serialVersionUID = -3782208159226605584L;
+
+			@Override
+			protected void onTimer(AjaxRequestTarget target) {
+
+				List<LoggedEvent> loggedEvents = recommendationService.listLoggedEvents(aModel.getObject().getProject(),
+						aModel.getObject().getUser());
+
+				// iterate over the logged events to extract the scores only.
+				List<Double> listDetailJson = new ArrayList<Double>();
```
reckart: If we know the size of the list to start with, I'd recommend to use an array of double (`double[]`) instead of a collection class.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.html
Sibgha360: I put the chart div outside the flex-content, and it still doesn't show.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: I want to be able to append javascript to the page. With this event listener I don't get target object where I can append the javascript. Am I missing something?

reckart: You can call `event.getRequestHandler()` which returns a `IPartialPageRequestHandler` that's about the same thing as an `AjaxRequestTarget` (i.e. `AjaxRequestTarget` implements `IPartialPageRequestHandler`).

reckart: The difference is that `IPartialPageRequestHandler` would also be available for web-socket-based communication if we switch to that one day - `AjaxRequestTarget` is limited to AJAX. In most case, a `IPartialPageRequestHandler` is sufficient.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.html
reckart: Try giving it an explicit minimum height.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Done

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Sometimes the score in the database is "NAN". Should we plot it as 0 or skip it? In case of skipping, we dont know the size of the array of valid scores to start with.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.html
Sibgha360: Not working. I tried a few more things to the div and no luck. Finally, while fiddling with the css of the page I noticed when the script 
`<link rel="stylesheet" type="text/css" href="./wicket/resource/de.tudarmstadt.ukp.clarin.webanno.brat.resource.BratCssVisReference/style-vis-ver-1D8B90DD5044623823E65D2592A1297C.css">`

is removed from the head of the page then the graph shows up. Does this give any clue?

reckart: There may be a clash between the CSS defined in that style sheet and the CSS required by the graph - but it needs closer investigation. The style-vis.css might be too broad and take over all SVG elements on a page including the graph.

reckart:

reckart:
@Sibgha360 trying to fix the CSS issue upstream: https://github.com/webanno/webanno/issues/1144

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
reckart: I have removed the check for 0 - how does the chart behave if there is a NAN in between?

reckart: I'd say probably render it as 0 if the chart fails treating the NAN natively.

Sibgha360: It ignores the NaN value and does not plot the point on the graph

reckart: Sounds ok to me for the time being.

reckart:

reckart:
@Sibgha360 Some of the files do not contain license headers:

```
[WARNING] Files with unapproved licenses:
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/D3JsReference.java
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/c3.min.js
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/C3JsReference.java
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/C3CssReference.java
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/c3.min.css
inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/resource/d3.min.js
```

Could you please copy a license header from one of our other Java files over to `D3JsReference`.

You could check if we can simply depend on WebJars for these resources, e.g.

* https://search.maven.org/search?q=org.webjars%20d3js
* https://search.maven.org/search?q=org.webjars%20c3js

E.g. we use *animate.js* as a webjar in `webanno-brat` like this:

*Maven*
```
    <dependency>
      <groupId>de.agilecoders.wicket.webjars</groupId>
      <artifactId>wicket-webjars</artifactId>
    </dependency>
    <dependency>
      <groupId>org.webjars</groupId>
      <artifactId>animate.css</artifactId>
      <version>3.2.6</version>
    </dependency>
```

*Java*
```
        aResponse.render(CssHeaderItem
                .forReference(new WebjarsCssResourceReference("animate.css/current/animate.css")));
```

If that is not possible and you if you cannot find a suitable header to add to the JS/CSS files, please exclude them from the license check. See e.g. the configuration for `apache-rat-plugin` in the `inception-pdf-editor` module on how to do that. Also, please add a `NOTICE.txt` file to the root of the `inception-recommendation` module explaining what the D3 and C3 files are and where they come from. Again, cf. the corresponding file in the `inception-pdf-editor` module.

reckart:
As discussed: implement one line per recommender, then open a new issue for a refined version where we have the option of showing one line per layer.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+		form.add(new NumberTextField<Integer>("maxPredictions", Integer.class).setMinimum(1).setMaximum(10).setStep(1));
+
+		form.add(new CheckBox("showAllPredictions"));
+
+		form.add(new LambdaAjaxButton<>("save", (_target, _form) -> aAnnotationPage.actionRefreshDocument(_target)));
+
+		add(form);
+
+		chartContainer = new Label("chart-container");
+		chartContainer.setOutputMarkupId(true);
```
Rentier: It would be nice to extract wicket ids into `private static final String`, e.g. like we do it in `RecommenderEditorPanel`.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+		this.aModel = aModel;
+		IModel<Preferences> model = LambdaModelAdapter.of(
+				() -> recommendationService.getPreferences(aModel.getObject().getUser(),
+						aModel.getObject().getProject()),
+				(v) -> recommendationService.setPreferences(aModel.getObject().getUser(),
+						aModel.getObject().getProject(), v));
+
+		Form<Preferences> form = new Form<>("form", CompoundPropertyModel.of(model));
+
+		form.add(new NumberTextField<Integer>("maxPredictions", Integer.class).setMinimum(1).setMaximum(10).setStep(1));
```
Rentier: I would write each setter in its extra line.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+
+			StringBuilder chartType = new StringBuilder();
+			
+			HashMap<String, List<Double>> recommenderScoreMap = getLatestScores(aEvent);
+
+			// iterate over the recommender score map to create data arrays to feed to the
+			// c3 graph
+			StringBuilder dataColumns = new StringBuilder();
+			Iterator<Entry<String, List<Double>>> it = recommenderScoreMap.entrySet().iterator();
+			while (it.hasNext()) {
```
Rentier: Is there a reason why you dont use a `for-each` here?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+				Entry<String, List<Double>> pair = it.next();
+
+				String jsonString = toJsonString(pair);
+				jsonString = jsonString.replaceAll("\".*?\"", "");
```
Rentier: It would be nice if you could comment in the code why you need to replace stuff here.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+				dataColumns.append(data);
+				dataColumns.append("]");
+				dataColumns.append(",");
+
+				// avoids a ConcurrentModificationException
+				it.remove();
+			}
+
+			String javascript = createJSScript(dataColumns.toString(), chartType.toString());
+			log.info(javascript);
```
Rentier: Please log it as debug or trace, info is really verbose.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+
+			aEvent.getRequestHandler().prependJavaScript(javascript);
+		} catch (IOException e) {
+			log.error("Unable to render chart", e);
```
Rentier: Where exactly is the error thrown? This try-catch looks really long, I would try to make it as short as possible.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+	 * Creates the js script to render graph with the help of given data points. Sample value of dataColumns:
+	 * ['recommender1', 1.0, 2.0, 3.0 ], ['recommender2', 2.0, 3.0, 4.0]. Also creates an xaxix of a sequence
+	 * from 0 to maximumNumberOfPoints (50)
+	 * 
+	 * @param dataColumns
+	 * @param chartType 
+	 * @return
+	 * @throws IOException
+	 */
+	private String createJSScript(String dataColumns, String chartType) throws IOException {
```
Rentier: Prefix arguments with `a`

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+		int[] intArray = IntStream.range(0, maxPointsToPlot).map(i -> i).toArray();
+
+		String xaxisValues = "[ 'x' ," + substring(Arrays.toString(intArray), 1, -1) + "]";
+		String data = toJsonString(dataColumns).substring(1, dataColumns.toString().length());
+
```
Rentier: Newline too much

Rentier:
The code contains tabs. Please install the DkPro checkstyle as described in https://zoidberg.ukp.informatik.tu-darmstadt.de/jenkins/job/INCEpTION%20(GitHub)%20(master)/de.tudarmstadt.ukp.inception.app$inception-app-webapp/doclinks/3/#_checkstyle . The Jenkins will not accept a PR unless checkstyle goes through. You can test it by building on the command line with `mvn -U   install -DskipTests`.

Rentier:
When I run it without any evaluations, then the diagram looks not so nice:

Also, it throws an Exception with `Error: No entity found for query` which does not yield too much information for users. I would maybe display a message instead of the chart that there are no evaluations yet  instead of throwing an exception.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+	private @SpringBean EventRepository eventRepo;
+
+	IModel<AnnotatorState> aModel;
+
+	int maxPointsToPlot = 50;
+
+	public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel, AnnotationActionHandler aActionHandler,
+			JCasProvider aJCasProvider, AnnotationPage aAnnotationPage) {
+		super(aId, aModel, aActionHandler, aJCasProvider, aAnnotationPage);
+		this.aModel = aModel;
```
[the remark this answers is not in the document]
reckart: Eclipse allows configuring this default prefix "a" under Preferences -> Java -> Code Style set the "prefix list" column in the "parameters" row to "a".

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+					list = new ArrayList<Double>();
+
+					addScoreToMap(recommenderScoreMap, detail.score, detail.tool, list);
+					continue;
+				}
+
+				addScoreToMap(recommenderScoreMap, detail.score, detail.tool, list);
+			
+			} catch (IOException e) {
+				log.error("Invalid logged Event detail. Skipping record with logged event id: " + loggedEvent.getId(),
```
Rentier: You can use String formatting for arguments in logging, e.g.
`log.error("Invalid logged Event detail. Skipping record with logged event id [{}]", loggedEvent.getId(), e);`
The `{}` are replaced with the arguments in the vararg, the stacktrace of the exception is logged if it is last, it does not need a placeholder.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
 import de.tudarmstadt.ukp.clarin.webanno.support.lambda.LambdaModelAdapter;
 import de.tudarmstadt.ukp.clarin.webanno.ui.annotation.AnnotationPage;
 import de.tudarmstadt.ukp.clarin.webanno.ui.annotation.sidebar.AnnotationSidebar_ImplBase;
+import de.tudarmstadt.ukp.inception.log.EventRepository;
+import de.tudarmstadt.ukp.inception.log.model.LoggedEvent;
 import de.tudarmstadt.ukp.inception.recommendation.api.RecommendationService;
 import de.tudarmstadt.ukp.inception.recommendation.api.model.Preferences;
+import de.tudarmstadt.ukp.inception.recommendation.log.RecommenderEvaluationResultEventAdapter.Details;
+
+public class RecommendationSidebar extends AnnotationSidebar_ImplBase {
```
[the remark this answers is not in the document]
Sibgha360: I am trying to add the checkstyle. It gives a parsing error. Any hint?

screen shot 2018-12-06 at 11 50 40 am

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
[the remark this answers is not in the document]
Sibgha360: @Rentier  I get it why it should be this.model instead of this.aModel. But must every argument to every function start with an a or only the one where we have another object of the same type as in the argument?

reckart:
@Sibgha360 You do not have to explicitly configure the Checkstyle plugin in Eclipse. Just install it as described in the [developer documentation](https://inception-project.github.io//releases/0.6.4/docs/developer-guide.html#_checkstyle) and Eclipse will automatically configure it based on the checkstyle settings in Maven.

The ["DKPro style file"](http://dkpro.github.io/contributing/) is *not* a checkstyle configuration file. It is an Eclipse formatter configuration file to be imported under *Preferences -> Java -> Code Style -> Formatter*.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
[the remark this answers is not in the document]
Rentier: All the arguments. Then you can also leave out the `this`

reckart:
We need to fix the recommender's self-evaluation before including this in the UI - otherwise it will be quite confusing to users. So for the time, being, I would suggest we add a configuration setting to explicitly enable this and if the setting is not turned on the chart won't be in the UI.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Done

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: No reason. For each is better. changed.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+				dataColumns.append(recommenderName);
+
+				// append data columns
+				dataColumns.append("', ");
+				dataColumns.append(data);
+				dataColumns.append("]");
+				dataColumns.append(",");
+
+				// avoids a ConcurrentModificationException
+				it.remove();
```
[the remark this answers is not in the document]
Sibgha360: if we use iterator to traverse the map, It avoids concurrent modification exception. Now I changed iterator to Foreach. It is simpler.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Rentier: You have two newlines here, you need at max one.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -17,55 +17,241 @@
...
+			// c3 graph
+			StringBuilder dataColumns = new StringBuilder();
+			Iterator<Entry<String, List<Double>>> it = recommenderScoreMap.entrySet().iterator();
+			while (it.hasNext()) {
+				Entry<String, List<Double>> pair = it.next();
+
+				String jsonString = toJsonString(pair);
+				jsonString = jsonString.replaceAll("\".*?\"", "");
+				jsonString = jsonString.replaceAll("'.*?'", "");
+				jsonString = jsonString.replaceAll("`.*?`", "");
```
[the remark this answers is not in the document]
reckart: I see. We need to extend the evaluation logging event then with the ID/name of the recommender instead or in addition to the tool. What the chart shows now is the tool name, but the recommender name could be different, in particular after #763 has been merged.

reckart:
@Sibgha360 The Maven dependency checker observes now that two dependencies are in the POM that appear not to be used:

```
[WARNING]    org.webjars:d3js:jar:5.5.0:compile
[WARNING]    org.webjars:c3:jar:0.6.6:compile
```

The Maven dependency checker can only detect the use of dependencies if they contain Java code that any code in the module refers to. Non-Java JARs, need to be excluded from the check, e.g. by adding a section such as this to the `pom.xml` file:

```
  <build>
    <pluginManagement>
      <plugins>
        <plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-dependency-plugin</artifactId>
          <configuration>
            <usedDependencies>
              <!--
                - Maven doesn't detect the the webjar dependencies because there is
                - no actual code dependency
              -->
              <usedDependency>org.webjars:d3js</usedDependency>
              <usedDependency>org.webjars:c3</usedDependency>
            </usedDependencies>
          </configuration>
        </plugin>
      </plugins>
    </pluginManagement>
  </build>
```

inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventRepository.java
```diff
@@ -17,11 +17,17 @@
...
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
 import de.tudarmstadt.ukp.inception.log.model.LoggedEvent;
 
 public interface EventRepository
 {
     static final String SERVICE_NAME = "eventRepository";
 
     void create(LoggedEvent aE);
+
+    List<LoggedEvent> listLoggedEvents(Project aProject, String aUsername, String eventType,
```
reckart: `eventType` and `size` miss the initial `a`.

inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventRepositoryImpl.java
```diff
@@ -43,4 +46,24 @@ public void create(LoggedEvent aEvent)
         log.info("{}", aEvent);
         entityManager.persist(aEvent);
     }
+
+    @Override
+    @Transactional
+    public List<LoggedEvent> listLoggedEvents(Project aProject, String aUsername, String eventType,
```
reckart: `eventType` and `size` miss the initial `a`.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<!--
+#Copyright 2017
```
reckart: Year should be 2018

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+    private @SpringBean EventRepository eventRepo;
+
+    IModel<AnnotatorState> model;
+
+    int maxPointsToPlot = 50;
+
+    public LearningCurveChartPanel(String aId, IModel<AnnotatorState> aModel)
+    {
+        super(aId);
+        this.model = aModel;
```
reckart: `this` not required.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+        aResponse.render(
+                CssHeaderItem.forReference(new WebjarsCssResourceReference("c3/current/c3.css")));
+    }
+
+    @OnEvent
+    public void onRenderAnnotations(RenderAnnotationsEvent aEvent)
+    {
+        log.info("rendered annotation event");
+
+        HashMap<String, List<Double>> recommenderScoreMap = getLatestScores(aEvent);
```
reckart: HashMap -> Map

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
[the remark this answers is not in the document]
reckart: Why remove at all?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+            aEvent.getRequestHandler().addChildren(getPage(), IFeedback.class);
+        }
+    }
+
+    /**
+     * Creates the js script to render graph with the help of given data points. Sample value of
+     * dataColumns: ['recommender1', 1.0, 2.0, 3.0 ], ['recommender2', 2.0, 3.0, 4.0]. Also creates
+     * an xaxix of a sequence from 0 to maximumNumberOfPoints (50)
+     * 
+     * @param aDataColumns
```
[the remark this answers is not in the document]
Sibgha360: You mean I should remove the following line from the comment,

`Sample value of dataColumns: ['recommender1', 1.0, 2.0, 3.0 ], ['recommender2', 2.0, 3.0, 4.0].`

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+    private HashMap<String, List<Double>> getLatestScores(RenderAnnotationsEvent aEvent)
+    {
+        // we want to plot RecommenderEvaluationResultEvent for the learning curve. The
+        // value of the event
+        String eventType = "RecommenderEvaluationResultEvent";
+
+        List<LoggedEvent> loggedEvents = eventRepo.listLoggedEvents(model.getObject().getProject(),
+                model.getObject().getUser().getUsername(), eventType, maxPointsToPlot);
+
+        if (loggedEvents == null || loggedEvents.isEmpty()) {
```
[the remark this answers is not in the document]
Sibgha360: What if there is no event in the database and eventRepo.listLoggedEvents returns null.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
reckart: No. I mean we do not need JavaDoc tags if they do not have a description:

```
     * @param aDataColumns NO DESCRIPTION HERE
```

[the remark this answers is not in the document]
reckart: But the embedded JSON is actually something I missed. Such literals should be escaped as e.g. `{@ code}` or `<code>...</code>` or some similar JavaDoc construct.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+
+        HashMap<String, List<Double>> recommenderScoreMap = new HashMap<>();
+
+        // iterate over the logged events to extract the scores and map
+        // it against its corresponding recommender.
+        for (LoggedEvent loggedEvent : loggedEvents) {
+            String detailJson = loggedEvent.getDetails();
+            try {
+                Details detail = fromJsonString(Details.class, detailJson);
+                List<Double> list = recommenderScoreMap.get(detail.tool);
```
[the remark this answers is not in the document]
Sibgha360: Are you referring to the null check on list?

```
                Details detail = fromJsonString(Details.class, detailJson);
                List<Double> list = recommenderScoreMap.get(detail.tool);
                if (list == null) {
                    list = new ArrayList<Double>();
                }
```

Here we are checking if the key(i.e. recommender) exists in the map. If not, then we add a new key with an empty list.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
reckart: `listLoggedEvents` calls `getResultList()` at the end to return the results. This never returns null. If there are no results, it returns an empty list.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
reckart: Use a `MultiValuedMap` like `org.apache.commons.collections4.multimap.ArrayListValuedHashMap<K, V>` - that saves you from this.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
Sibgha360: Got it.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+import de.tudarmstadt.ukp.inception.recommendation.log.RecommenderEvaluationResultEventAdapter.Details;
+
+public class LearningCurveChartPanel
+    extends Panel
+{
+    private static final long serialVersionUID = 4306746527837380863L;
+
+    private static final String CHART_CONTAINER = "chart-container";
+
+    private Logger log = LoggerFactory.getLogger(getClass());
```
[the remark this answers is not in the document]
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+
+        // import Css
+        aResponse.render(
+                CssHeaderItem.forReference(new WebjarsCssResourceReference("c3/current/c3.css")));
+    }
+
+    @OnEvent
+    public void onRenderAnnotations(RenderAnnotationsEvent aEvent)
+    {
+        log.info("rendered annotation event");
```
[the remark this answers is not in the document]
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+        }
+
+        // iterate over the recommender score map to create data arrays to feed to the
+        // c3 graph
+        StringBuilder dataColumns = new StringBuilder();
+        StringBuilder chartType = new StringBuilder();
+
+        recommenderScoreMap.forEach((key, value) -> {
+
+            ArrayList<Double> listScores = (ArrayList<Double>) value;
```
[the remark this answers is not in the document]
Sibgha360: done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+
+            // should not happen ideally
+            if (listScores == null || listScores.size() <= 0) {
```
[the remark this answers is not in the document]
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
Sibgha360: Used <pre></pre>

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+                + "',data:{ x:'x', columns:[" + xaxisValues + " ," + data + "],types:{" + aChartType
+                + "}}});;";
+        return javascript;
+    }
+
+    /**
+     * Fetches a number of latest evaluation scores from the database and save it in the map
+     * corresponding to each recommender for which the scores have been logged in the database
+     */
+    private HashMap<String, List<Double>> getLatestScores(RenderAnnotationsEvent aEvent)
```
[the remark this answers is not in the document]
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
Sibgha360: Got it.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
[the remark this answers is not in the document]
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+    {
+        // we want to plot RecommenderEvaluationResultEvent for the learning curve. The
+        // value of the event
+        String eventType = "RecommenderEvaluationResultEvent";
+
+        List<LoggedEvent> loggedEvents = eventRepo.listLoggedEvents(model.getObject().getProject(),
+                model.getObject().getUser().getUsername(), eventType, maxPointsToPlot);
+
+        if (loggedEvents == null || loggedEvents.isEmpty()) {
+            return null;
```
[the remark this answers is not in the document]
Sibgha360: `Collections.emptyMap()` doesn't return multi valued map. I used `new ArrayListValuedHashMap<>()`.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+     */
+    private void addScoreToMap(HashMap<String, List<Double>> aRecommenderScoreMap, Double aScore,
+            String tool, List<Double> list)
+    {
+        // sometimes score values NaN. Can result into error while rendering the graph on UI
+        if (!Double.isFinite(aScore)) {
+            return;
+        }
+
+        list.add(aScore);
```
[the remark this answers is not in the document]
Sibgha360: Done.

inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventRepository.java
Sibgha360: Done. Should be careful

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.html
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
Sibgha360: Got it.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+    private static final long serialVersionUID = 4306746527837380863L;
+
+    private static final String CHART_CONTAINER = "chart-container";
+
+    private Logger log = LoggerFactory.getLogger(getClass());
+    private WebComponent chartContainer;
+
+    private @SpringBean EventRepository eventRepo;
+
+    IModel<AnnotatorState> model;
```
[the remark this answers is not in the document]
Sibgha360: I made all the field final that are only set in the constructor.

inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventRepositoryImpl.java
Sibgha360: Done.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,259 @@
...
+            dataColumns.append(recommenderName);
+
+            // append data columns
+            dataColumns.append("', ");
+            dataColumns.append(data);
+            dataColumns.append("]");
+            dataColumns.append(",");
+        });
+
+        // should not happen ideally
```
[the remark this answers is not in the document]
Sibgha360: I guess no. I removed the check.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -208,12 +212,20 @@ private String createJSScript(String aDataColumns, String aChartType) throws IOE
             try {
                 Details detail = fromJsonString(Details.class, detailJson);
 
+                //do not include the scores from disabled recommenders
+                List<Recommender> recommenderIfActive = recommendationService.getRecommenderIfActive(model.getObject().getProject(), detail.layer, detail.tool);
+                if(recommenderIfActive.size()<=0)
```
reckart: Use `isEmpty()` instead of a size comparison.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -186,6 +187,24 @@ public Recommender getRecommender(long aId)
         return entityManager.find(Recommender.class, aId);
     }
 
+    @Override
+    @Transactional
+    public List<Recommender> getRecommenderIfActive(Project aProject, String aLayer, String aTool)
```
reckart: We have "active" and "enabled" recommenders. A recommender is "active" if it is used to generate suggestions (i.e. if its evaluation score surpasses the threshold or if it is "always active"). Thus better rename to `getEnabledRecommenders()` to avoid confusion.

inception-log/src/test/java/de/tudarmstadt/ukp/inception/log/EventRepositoryImplTest.java
```diff
@@ -0,0 +1,114 @@
...
+
+    private EventRepositoryImpl sut;
+    private Project project;
+    private User user;
+    private LoggedEvent le;
+
+    @BeforeClass
+    public static void setUpOnce()
+    {
+        System.setProperty("org.eclipse.rdf4j.repository.debug", "true");
```
reckart: This is not required - no RDF4J used in this part of the code.

reckart:
@Sibgha360 looks like your the checkstyle plugin for your IDE is still not working. I'm pushing a few checkstyle fixes.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
```diff
@@ -0,0 +1,242 @@
...
+                    continue;
+                }
+
+                // sometimes score values NaN. Can result into error while rendering the graph on UI
+                if (!Double.isFinite(detail.score)) {
+                    continue;
+                }
+
+                recommenderScoreMap.put(detail.tool,
+                        Double.valueOf(new DecimalFormat("#.0000").format(detail.score)));
```
reckart: This causes an exception on my machine: 

```
Caused by: java.lang.NumberFormatException: For input string: ",4444"
	at jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054) ~[?:?]
	at jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110) ~[?:?]
	at java.lang.Double.parseDouble(Double.java:543) ~[?:?]
	at java.lang.Double.valueOf(Double.java:506) ~[?:?]
	at de.tudarmstadt.ukp.inception.recommendation.sidebar.LearningCurveChartPanel.getLatestScores(LearningCurveChartPanel.java:225) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.recommendation.sidebar.LearningCurveChartPanel.onRenderAnnotations(LearningCurveChartPanel.java:105) ~[classes/:?]
```

reckart: Instead of handling this on the Java side, better pass the proper number to the client and handle it in JS (cf. https://github.com/c3js/c3/issues/2000#issuecomment-291787879).

Sibgha360: I will move it to the JS. But I don't understand how the input string ended up to be ",4444". It is the value coming from database and isn't it always a double value?

inception-log/src/test/java/de/tudarmstadt/ukp/inception/log/EventRepositoryImplTest.java
Sibgha360: Removed

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
Sibgha360: Updated

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
Sibgha360: Done

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/LearningCurveChartPanel.java
reckart: If one doesn't pay close attention and always specifies an explicit format, then the float-to-string conversion uses the system locale. I guess with my system locale settings, floats are rendered with a comma instead of a dot by default.

Sibgha360: Resolved.

Issue #713, opened 2018-11-26
Add project role filter to project overview list


reckart:
**Is your feature request related to a problem? Please describe.**
The list of projects can at times be a bit crowded. Adding an option to filter the projects by the role I have would be nice.

**Describe the solution you'd like**
Add a set of toggle-buttons at the top of the list would be nice.

**Describe alternatives you've considered**
Adding a input field with the option to search e.g. for `role:manager` would of course also be nice.

PR #714, opened 2018-11-26
#713 - Add project role filter to project overview list

reckart:
- Added toggle buttons to filter projects by role
- Documented new functionality

### DIFF
--- a/inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.java
+++ b/inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.java
@@ -141,6 +141,9 @@ public class ProjectsOverviewPage
         config.allowedFileExtensions(asList("zip"));
         config.showPreview(false);
         config.showUpload(true);
+        config.removeIcon("<i class=\"fa fa-remove\"></i>");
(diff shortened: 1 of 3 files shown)


inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.java
```diff
@@ -98,12 +107,17 @@
     
     private BootstrapFileInputField fileUpload;
     private WebMarkupContainer projectListContainer;
+    private WebMarkupContainer roleFilters;
+    private IModel<Set<PermissionLevel>> activeRoleFilters;
```
Rentier: You could use a fancy `EnumSet`, but it is not too important.

reckart:
Make sure you have the very latest WebAnno SNAPSHOTs when building.

Rentier:
I cleaned my installation. Is that not enough?

reckart:
What do you mean by "cleaned"? If you have WebAnno checked out, you need to pull first. Depending on how you build, you may then need to "mvn clean install" WebAnno before INCEpTION can make use of it. If you do not have WebAnno checked out, then building inception with "mvn -U clean install" should help.

Rentier:
It works for me, although I do not like the large gap on the left side between "Create new project" and project List the new selector introduced.

reckart:

Issue #715, opened 2018-11-26
Support basic markdown in project description


reckart:
**Is your feature request related to a problem? Please describe.**
Sometimes having a project description with some document structure (enumerations, formatting, etc.) would be nice.

**Describe the solution you'd like**
Include some kind of markdown processor.

reckart:
Example of how this looks like in action:

![2018-11-26_16-32-14](https://user-images.githubusercontent.com/1410238/49024067-f0148600-f198-11e8-9a2a-8eccf8f5e171.png)

PR #716, opened 2018-11-26
#715 - Support basic markdown in project description

reckart:
- Process project description using txtmark (Markdown) before rendering
- Refresh project from DB instead of directly using the session object so we always display the latest description / title

### DIFF
--- a/inception-ui-core/pom.xml
+++ b/inception-ui-core/pom.xml
@@ -124,6 +124,11 @@
     </dependency>
     
     <dependency>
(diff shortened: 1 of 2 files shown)


reckart:
> Why is this a bugfix (see branch name)?

Naming mistake noticed only after creating the PR.

> It breaks (overflows the footer) if you put more text into the page than the page is high.

Thanks, checking...

reckart:
Page overflow fixed.

Rentier:
Works for me now, the markdown looks really nice.

Issue #303, opened 2018-06-19
Configure document states to consider for training


reckart:
Add a setting to recommenders allowing the user to configure on which document states / annotation states to train. E.g. the manager may want to import pre-annotated data to overcome a recommender cold start but mark the respective documents as locked for all annotations. Similarly, a manager may require that a document is curated before being considered for recommendation.

- [x] Add to backend (recommender model and training task)
- [x] Add to UI
- [x] Update export/import
- [x] Update documentation

reckart:
@Rentier this is the issue about including only documents in certain states into the recommender training, e.g. documents which are marked as finished or marked as curated.

Rentier:
Do you want to have a set of `AnnotationDocumentState` which should be considered, making the configuration a list of checkboxes?

reckart:
no, because `AnnotationDocumentState` doesn't capture whether a document is curated or not.

reckart:
Well, maybe the case for the curated documents needs to be handled separately. In that case, a checkboxgroup might work.

Rentier:
How would we save these options in the database? As it is a collection, I would need a new table, don't I?

reckart:
@MdeBoer95 has used a primitive collection for the KB root concepts. That might be an option.

Rentier:
I will do that. It still creates a table though.

PR #725, opened 2018-11-27
#303 - Configure document states to consider for training

Rentier:
- Add states considered for training to recommender
- Add configure UI for that
- Altered training task to skip documents with wrong state


Rentier:
Checking whether the import works still needs to be done, also I am not sure whether I chose the best way for the default values of the new field.

Rentier:
I will add the documentation once the code part of the PR is done, as I need to update images of the UI.

reckart:
Removed "Needs review" and added "WIP" since the PR apparently is not complete yet.

Rentier:
The test that fails is

``java.lang.AssertionError: expected:<1> but was:<0>
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasDocumentIndexTest.testAnnotationQuery(MtasDocumentIndexTest.java:374)```

which is strange.

reckart:
Does it fail consistently? I had a similar failure recently and just asked Jenkins to rebuild and the second time it worked.

Rentier:
Now the error is different:
`java.io.UTFDataFormatException: encoded string too long: 70441 bytes`

reckart:
Very odd error - let's see if this is consistently happening... it happens somehow as the DB schema is set up in the in-memory DB... no idea what's going on here really.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/TrainingTask.java
```diff
@@ -133,17 +142,28 @@ public void run()
...
-                casses.add(jCas.getCas());
+                casses.add(new AnnotationCas(jCas.getCas(), doc.getState()));
             } catch (IOException e) {
                 log.error("Cannot read annotation CAS.", e);
             }
         }
         return casses;
     }
+
+    private static class AnnotationCas
```
reckart: I find the name confusing. How about `TrainingDocument`?

inception-recommendation/src/test/java/de/tudarmstadt/ukp/inception/recommendation/exporter/RecommenderExporterTest.java
```diff
@@ -193,4 +225,8 @@ private Recommender buildRecommender(String id)
         return recommender;
     }
 
+    private static <T> Set<T> asSet(T... a)
```
reckart: It's funny that Java doesn't offer such a method, isn't it?

Rentier: Guava has it, but we dont like guava.

reckart: I have opened an issue with Apache Commons Collections - let's see if somebody implements it (or if one of us maybe sends them a patch)  https://issues.apache.org/jira/projects/COLLECTIONS/issues/COLLECTIONS-706

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/scheduling/tasks/TrainingTask.java
Rentier: I am thinking right now that it would be easier to store the states we want to ignore instead of the states we want to use for training. The changeset is not so trivial to do.

reckart: We could implement a custom task in Java which does what we want. I'm not sure if reversing the logic would bite us back later at some point. Although in the worst case, we could try it with the reversed logic and implement a custom task to again reverse it later if it bites us...

Rentier: Then I implement the task for the non-reversed logic.

Rentier: I reversed the logic.

Rentier:
From my side, this can be reviewed.

reckart:

Rentier:
I hope all goes well with this.

reckart:
@Rentier BTW did you eventually add the documentation for this new feature?

Rentier:
Of course. I also sneaked the documentation for max suggestions and show all.

Rentier:
I forgot to push it... :(

Issue #724, opened 2018-11-27
Dependency relations cause NPE during MTAS indexing


reckart:
**Describe the bug**
The relation indexing code is unable to handle relations with attach-features and this causes an NPE when indexing dependency relations:

**To Reproduce**
1. Open any document with dependency relations and watch the console

```
2018-11-27 14:29:03 ERROR [SYSTEM] MtasUimaParser - Unable to create token collection
java.lang.NullPointerException: null
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.indexFeatures(MtasUimaParser.java:325) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.indexAnnotation(MtasUimaParser.java:291) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.createTokenCollection(MtasUimaParser.java:212) ~[classes/:?]
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasUimaParser.createTokenCollection(MtasUimaParser.java:170) ~[classes/:?]
	at mtas.analysis.MtasTokenizer.constructTokenCollection(MtasTokenizer.java:237) ~[mtas-7.3.0.3.jar:?]
	at mtas.analysis.MtasTokenizer.reset(MtasTokenizer.java:171) ~[mtas-7.3.0.3.jar:?]
	at org.apache.lucene.index.DefaultIndexingChain$PerField.invert(DefaultIndexingChain.java:732) ~[lucene-core-7.3.0.jar:7.3.0 98a6b3d642928b1ac9076c6c5a369472581f7633 - woody - 2018-03-28 14:31:58]
	at org.apache.lucene.index.DefaultIndexingChain.processField(DefaultIndexingChain.java:430) ~[lucene-core-7.3.0.jar:7.3.0 98a6b3d642928b1ac9076c6c5a369472581f7633 - woody - 2018-03-28 14:31:58]
	at org.apache.lucene.index.DefaultIndexingChain.processDocument(DefaultIndexingChain.java:392) ~[lucene-core-7.3.0.jar:7.3.0 98a6b3d642928b1ac9076c6c5a369472581f7633 - woody - 2018-03-28 14:31:58]
[...]

PR #730, opened 2018-11-29
#724 - Dependency relations cause NPE during MTAS indexing

reckart:
- Support attach-feature on relation layers
- Added unit test


naveen2507:
Dependency issues in PR

Issue #731, opened 2018-11-29
MTAS tests failing occasionally


reckart:
**Describe the refactoring action**
We have some PRs that fail because of an unstable MTAS test. It looks like the timeout configured in the test for async indexing is too tight and has to be relaxed a bit.

**Expected benefit**
Unit test should become more stable.

reckart:
@wpmp hit this issue again on Jenkins:

```
java.lang.AssertionError: expected:<1> but was:<0>
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasDocumentIndexTest.testAnnotationQuery(MtasDocumentIndexTest.java:359)
```

reckart:
Have not seen this for a long time. Assuming it is resolved again.

PR #732, opened 2018-11-29
#731 - MTAS tests failing occasionally

reckart:
- Give the async indexing more time
- Switch to use Awaility instead of hand-coded timeout checks

### DIFF
--- a/inception-search-mtas/pom.xml
+++ b/inception-search-mtas/pom.xml
@@ -217,7 +217,12 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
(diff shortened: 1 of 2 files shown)


inception-search-mtas/pom.xml
```diff
@@ -217,7 +217,12 @@
       <artifactId>assertj-core</artifactId>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <version>3.1.3</version>
```
Rentier: Dont we put version numbers in the master pom?

PR #734, opened 2018-11-30
Bugfix/733 subclasses have random name

MdeBoer95:


reckart:
Why did you remove the cancel button?

MdeBoer95:
Because closing the modal window with the X does the same job.

reckart:
> Because closing the modal window with the X does the same job.

We usually have the cancel button nevertheless in our dialogs. I'll restore it then.

reckart:
@MdeBoer95 never mind - I like how it looks, even though it doesn't match the other dialogs. Talking about the cancel button though, I note that there is no way of canceling the creation of a top-level class. Btw. do you remember why we decided to use a popup dialog for the creation of subclasses instead of using the same approach as when we create top-level classes? It seems to me that creating a top-level class is basically a special case of creating a subclass, namely one where the parent class is not yet, right?

reckart:
Getting this exception when actually trying to create a subclass:

```
org.apache.wicket.WicketRuntimeException: No Page found for component: [LambdaAjaxButton [Component id = createSubclass, page = <No Page>, path = details:info:createSubclass:content:form:createSubclass.LambdaAjaxButton], children =  [Component id = createLabel]]. You probably forgot to add it to its parent component.
     at org.apache.wicket.Component.getPage(Component.java:1685)
     at de.tudarmstadt.ukp.clarin.webanno.support.lambda.LambdaAjaxButton.onSubmit(LambdaAjaxButton.java:59)
     at org.apache.wicket.ajax.markup.html.form.AjaxButton$1.onSubmit(AjaxButton.java:113)
     at org.apache.wicket.ajax.form.AjaxFormSubmitBehavior$AjaxFormSubmitter.onSubmit(AjaxFormSubmitBehavior.java:223)
     at org.apache.wicket.markup.html.form.Form.delegateSubmit(Form.java:1271)
     at org.apache.wicket.markup.html.form.Form.process(Form.java:964)
     at org.apache.wicket.markup.html.form.Form.onFormSubmitted(Form.java:788)
     at org.apache.wicket.ajax.form.AjaxFormSubmitBehavior.onEvent(AjaxFormSubmitBehavior.java:176)
...
```

MdeBoer95:
> @MdeBoer95 never mind - I like how it looks, even though it doesn't match the other dialogs. Talking about the cancel button though, I note that there is no way of canceling the creation of a top-level class. Btw. do you remember why we decided to use a popup dialog for the creation of subclasses instead of using the same approach as when we create top-level classes? It seems to me that creating a top-level class is basically a special case of creating a subclass, namely one where the parent class is not yet, right?

I had the same thoughts but I don't remember why we decided to use the popup dialog.
Regarding cancelling the creation of a top-level class, there is this comment in the code ;)
```java 
// there exists functionality to cancel the "new KBObject" prompt, but in hindsight, MB
// thinks this functionality is not really needed in the UI, so the button is hidden
// here
    LambdaAjaxLink cancelButton = new LambdaAjaxLink("cancel",
        AbstractInfoPanel.this::actionCancel)
            .onConfigure((_this) -> _this.setVisible(false)); ```

**Comment by MdeBoer95** [comment]
I don't get that exception when I create a subclass. 
Does the application crash?
I only get:

``` can not update component because its page is not the same as the one this handler has been created for. Component: [ContentPanel [Component id = content]]
java.lang.IllegalArgumentException: Cannot update component because its page is not the same as the one this handler has been created for. Component: [ContentPanel [Component id = content]]
```


It was also there before this PR but the subclass is still created and everything is rendered as expected.

reckart:
There shouldn't be any exception. Trying to get rid of it here...

reckart:
Ok, I think I nailed it. The code was first firing an event which causes the page layout to change and was then closing the dialog. This caused an exception in the Wicket code that closes the dialog. First closing the dialog and then sending the event fixes this issue.

Issue #739, opened 2018-12-03
Cookie warnings spam log


reckart:
**Describe the bug**
There are tons of messages like the following in the log if entity linking is used:

```
2018-12-02 19:15:57 WARN [SYSTEM] ResponseProcessCookies - Invalid cookie header: "Set-Cookie: WMF-Last-Access-Global=02-Dec-2018;Path=/;Domain=.wikidata.org;HttpOnly;secure;Expires=Thu, 03 Jan 2019 12:00:0
0 GMT". Invalid 'expires' attribute: Thu, 03 Jan 2019 12:00:00 GMT
```

**To Reproduce**
1. Configure the official Wikidata endpoint as a KB and link against it

**Expected behavior**
Actually, the `expires` attribute seems ok... not sure why it is complaining.

**Please complete the following information:**
 - Version and build ID:  0.6.4 (2018-12-03 12:25:04, build 0ce42a9ee58f9ef43ecdaafc954064ebafe18588)

reckart:
The best solution might be to disable the warnings via the log configuration. I the problems seems to be that the the expires value doesn't fit the default configuration of httpclient and I find no way of changing the policy from the outside (i.e. without changing RDF4J code).

reckart:
It is possible to explicitly set a http client on the repository manager, but I'm not sure how to configure it with a more sensible cookie policy:

```
        repoManager.setHttpClient(HttpClientBuilder.create()
                .setDefaultCookieSpecRegistry(...)
                .useSystemProperties()
                .disableAutomaticRetries()
                .build());
```

PR #740, opened 2018-12-03
#739 - Cookie warnings spam log

reckart:
- Decrease log level for ResponseProcessCookies to avoid log being spammed on cookies that httpclient considers malformed (happens a *lot* with Wikidata)


reckart:

Issue #742, opened 2018-12-04
Get rid of TokenObject


reckart:
**Describe the refactoring action**
Remove the TokenObject class - conflate all information into AnnotationObject.

**Expected benefit**
One class less to worry about.

PR #743, opened 2018-12-04
#742 - Get rid of TokenObject

reckart:
- Removed TokenObject class and conflated relevant info into AnnotationObject
- Remove deprecated fields from Offset class


inception-active-learning/src/main/java/de/tudarmstadt/ukp/inception/active/learning/sidebar/ActiveLearningSidebar.java
```diff
@@ -560,8 +560,8 @@ private void writeLearningRecordInDatabaseAndEventLog(LearningRecordUserAction u
         record.setSourceDocument(sourceDoc);
         record.setTokenText(currentRecommendation.getCoveredText());
         record.setUserAction(userAction);
-        record.setOffsetTokenBegin(currentRecommendation.getOffset().getBeginToken());
-        record.setOffsetTokenEnd(currentRecommendation.getOffset().getEndToken());
+        record.setOffsetTokenBegin(-1);
```
Rentier: Why?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/model/Offset.java
```diff
@@ -27,30 +27,16 @@
...
-    @Deprecated
-    private int endToken = -1;
-
-    @Deprecated
-    public Offset()
-    {
-    }
-
-    public Offset(int beginCharacter, int endCharacter, int beginToken, int endToken)
+    public Offset(int beginCharacter, int endCharacter)
```
Rentier: Should we deprecate the whole class?

reckart: I tried deleting it, but it's used quite all over the place. This needs a bit more investigation before deprecating.

Issue #738, opened 2018-12-03
External recommender should not send empty training requests


Rentier:
The external recommender sends training requests even when there are no annotations yet. It would be nice to not do that.

reckart:
> Do you think it makes sense to look in the TrainingTask whether a CAS contains annotations and then just do not hand it to the recommenders?

Yes, I think that makes sense.

Rentier:
@Horsmann INCEpTION should not send empty training requests anymore.

PR #747, opened 2018-12-04
#738 - External recommender should not send empty training requests

Rentier:
- Filter out CAS in TrainingTask that do not contain target annotations

Depends on #303


inception-doc/src/main/resources/META-INF/asciidoc/user-guide.adoc
```diff
@@ -143,7 +143,7 @@ include::{include-dir}projects_layers.adoc[leveloffset=+1]
 
 include::{include-dir}projects_knowledge-base.adoc[leveloffset=+1]
 
-include::{include-dir}projects_recommendation.adoc[leveloffset=+1]
+include::{include-dir}projects_recommendation.adoc[leveloffset=+2]
```
reckart: This is confusing because the section is at the same level as the KB documentation. Better leave it at +1.

Rentier: I would rather change the kb documentation to +2, because it starts at `==` inside.

reckart: The main file (containing the includes) is the basis. Sub-sections should be included as +1. If a sub-section has additional sub-sections, these should be included as +2. KB and Recommender are both sub-sections of the main file, so they should be +1.

Rentier: I think KB needs to be at 2, as it is `Main->Projects->Knowledge Bases` and `Main->Projects->Recommender`

reckart: Hm, right - I missed the prefix. Then let's change the KB to +2.

Issue #754, opened 2018-12-07
String recommender doesn't activate for zero threshold


reckart:
**Describe the bug**
The string recommender doesn't activate even if the threshold is set to 0. This is somewhat surprising because one would expect that things cannot get worse than 0 as a score. Thing is that the string recommender evaluates itself using Krippendorff Alpha unitizing which apparently can return scores lower than 0.

**Expected behavior**
I would expect that a threshold of 0 will always run the recommender *plus* do evaluation (which does not happen if *always active* is turned on).

**Please complete the following information:**
 - Version and build ID: f4f1c1612f3f054c5418de6e87adf06343017d8e

PR #755, opened 2018-12-07
#754 - String recommender doesn't activate for zero threshold

reckart:
- Cap the score at 0 - do not return lower values.


inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/StringMatchingRecommender.java
```diff
@@ -87,6 +87,8 @@ public void train(RecommenderContext aContext, List<CAS> aCasses)
         
         aContext.put(KEY_MODEL, dict);
         aContext.markAsReadyForPrediction();
+        
+        log.info("Learned dictionary model with {} entries", dict.size());
```
Rentier: Debug?

Issue #756, opened 2018-12-07
Exporting recommender uses wrong field name for states ignored during training


Rentier:
Exporting recommender uses wrong field name for states ignored during training:

```
    @JsonProperty("statesForTraining")
    private Set<AnnotationDocumentState> statesForTraining;
```

needs to be

```
    @JsonProperty("statesIgnoredForTraining")
    private Set<AnnotationDocumentState> statesIgnoredForTraining;
```

PR #757, opened 2018-12-07
#756 - Exporting recommender uses wrong field names

Rentier:
- Fix field name


Rentier:
@reckart Can we merge this now?

Issue #758, opened 2018-12-08
OpenNLP NER fails on overlapping annotations


reckart:
**Describe the bug**
The OpenNLP NER recommender throws an exception if there are overlapping annotations on its target layer.

**To Reproduce**
1. Create OpenNLP NER recommender on named entity layer
2. Create two overlapping annotations (maybe create a few more non-overlapping to avoid the recommender not running due to lack of training data
3. Observe exception in log

**Expected behavior**
Well, short of actually supporting overlapping annotations, it would be a first step to keep only the first of the overlapping annotations and to drop the others.

**Please complete the following information:**
 - Version and build ID: f4f1c1612f3f054c5418de6e87adf06343017d8e

PR #759, opened 2018-12-08
#758 - OpenNLP NER fails on overlapping annotations

reckart:
- Remember the end of the last annotation and do not accept any new annotation that starts earlier than that

### DIFF
--- a/inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommender.java
+++ b/inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommender.java
@@ -227,13 +227,21 @@ public class OpenNlpNerRecommender
         List<AnnotationFS> annotations = selectCovered(annotationType, aSentence);
         int numberOfAnnotations = annotations.size();
         List<Span> result = new ArrayList<>();
+        int maxSeen = 0;
         for (int i = 0; i < numberOfAnnotations; i++) {
             AnnotationFS annotation = annotations.get(i);
             int begin = idxToken.get(idxTokenOffset.get(annotation.getBegin()));
             int end = idxToken.get(idxTokenOffset.get(annotation.getEnd()));
             String label = annotation.getFeatureValueAsString(feature);
+            
+            if (begin < maxSeen) {
+                LOG.warn("Skipping overlapping annotation: [{}-{}, {}]", begin, end + 1, label);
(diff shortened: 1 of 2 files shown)


inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommender.java
```diff
@@ -227,13 +227,21 @@ private NameSample createNameSample(CAS aCas, AnnotationFS aSentence,
...
         int numberOfAnnotations = annotations.size();
         List<Span> result = new ArrayList<>();
+        int maxSeen = 0;
         for (int i = 0; i < numberOfAnnotations; i++) {
             AnnotationFS annotation = annotations.get(i);
             int begin = idxToken.get(idxTokenOffset.get(annotation.getBegin()));
             int end = idxToken.get(idxTokenOffset.get(annotation.getEnd()));
             String label = annotation.getFeatureValueAsString(feature);
+            
+            if (begin < maxSeen) {
```
Rentier: It would be nice if you could comment this code, I need to think alot why that makes sense.

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommender.java
```diff
@@ -227,13 +227,21 @@ private NameSample createNameSample(CAS aCas, AnnotationFS aSentence,
         List<AnnotationFS> annotations = selectCovered(annotationType, aSentence);
         int numberOfAnnotations = annotations.size();
         List<Span> result = new ArrayList<>();
+        int maxSeen = 0;
```
Rentier: `maxSeen` sounds more like a count than an index. I would call it something like `largestEndSoFar` or so

reckart: Variable has been renamed

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommender.java
reckart: Code has been commented.

Issue #761, opened 2018-12-09
Option to explicitly enable PDF/HTML editor support


reckart:
**Describe the refactoring action**
While the PDF and HTML annotation editor aren't quite done, they should be disabled by default.

**Expected benefit**
Remove less-then-alpha functionality from the UI.

PR #762, opened 2018-12-09
#761 - Option to explicitly enable PDF/HTML editor support

reckart:
- Disabled PDF editor by default - can be activated by adding `ui.pdf.enabled=true` to the settings.properties file
- Disabled HTML editor by default - can be activated by adding `ui.html.enabled=true` to the settings.properties file


wpmp:
In which file do I have to set this new properties?

reckart:
You can either explicitly specify a location where INCEpTION puts its data and looks for the `settings.properties` file using `-Dinception.home=/path/to/some/folder`. If you do not specify this system property, INCEpTION uses `~/.inception` by default, so you could place the `settings.properties` file there.

PR #745, opened 2018-12-04
Feature/372 redesign layout of search page

az79nefy:
* Redesign of the external search page layout (resolves #372 )
* Import document by button + import status changes dynamically (resolves #374 )
* Document opens as new window instead of popup (resolves #375 )

How to test: 
- Access page through the "Search" menu button and make a search query
- Open the text of a document
- Import a document

PR #765, opened 2018-12-10
#764 basic results reranking for external search

### LINKED ISSUE #764: Basic results reranking for external search
We want to be able to re-rank and/or post process the results obtained through external search. 

To investigate how to implement this, we consider the following case:

* search remote repository for documents using a query consisting of one or more words
* retrieve the matching documents
* split the documents into sentences
* retain all sentences containing the query words
* score the remaining sentences
* aggregate the scores to the document level
* display the results according to the aggregate score

- [ ] Implement this as a basic unit test
- [ ] Integrate the pipeline into the application so it can be tried out live

reckart:
**What's in the PR**
* Extract individual sentences from external search results
* Score sentences with respect to suitability for dictonary examples
* Cluster sentences based on similarity
* Display sentences in a separate view of the external search page
* resolves #764 

**How to test manually**
* Execute a query in external search with "show best sentences" view. This is done by appending " sentenceHighlights:true" to the end of the desired query.
* Observe the highlights of returned search results (should be sentences).

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


reckart:
@az79nefy this PR is based on #745. Once #745 is merged into master and we then merge master into this branch, the overlapping commits should disappear from this PR... or not, but it won't matter then either ;)

az79nefy:
@reckart I implemented this features by appending " sentenceHighlights:true" to the end of the desired query. Do you want a different way to toggle the "show best sentences" highlighting?

jcklie:
@reckart @az79nefy What is the state of this PR?

reckart:
I need to have a close look at it... so I'm the blocker here.

reckart:
It looks like the basic structure of the enhancement is there, but there still seems to be a long way to go.

One of the first things that needs to be done is to improve the speed of the clustering because one can hardly test/debug right now.

The next thing to consider after that is how to deal with the fact that the `ExternalSearchResult` is currently designed to represent a single document - but when we cluster by sub-document units, then this is questioned. I have changed the code such that the title/documentId of the result is obtained from the "best" representative of a cluster. That means when pressing "import", the document with the "best mention" will be imported. The question is whether that is a good approach or whether we would want to import the "n-best" documents for each cluster.

PR #763, opened 2018-12-10
#746 - Creating two recommenders with same configuration throws

reckart:
- Add a proper validator for the name field which checks if a recommender by the same name already exists and provides a suitable error message in this case
- Add an option to auto-generate the recommender name to avoid forcing the user into inventing names all the time

**How to test**

* try creating new recommenders and editing existing recommenders with and without auto-generating the name
* try changing the recommender configuration with auto-generation enabled/disabled to see if/when the name is updated based on the configuration
* try creating recommenders with emtpy names and duplicate names

PR #766, opened 2018-12-10
Bugfix/746 creating two recommenders with same configuration throws 0 7 x

reckart:
**What's in the PR**
* Same as #763 but against master (i.e. with conflicts resolved)

**How to test manually**
* See #763

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
> I created two with the same name (embedded db) and it crashes:

Should be fixed now in both PRs. I guess I mostly tested with modifying existing recommenders and forgot to try the case where one really creates a new one...

Rentier:
Sry. I still have the error when having two recommender, attempting to change the name of one to the name of the second and then switching between the recommenders does not update the name anymore.

Rentier:
Works now

Issue #811, opened 2019-01-02
Search Bar for Knowledge Base Page


MdeBoer95:
**Is your feature request related to a problem? Please describe.**
Finding a specific concept/instance in the concept tree can be cumbersome especially if the parent concept or the type of an instance is unknown

**Describe the solution you'd like**
A search bar at the top of the page should make it easier to find a specific concept/instance

PR #818, opened 2019-01-03
#811 - Search bar for knowledge base page

MdeBoer95:
**What's in the PR**
* Add a search bar to the KB Page that makes it easier to find concepts and instances in the knowledge base

**How to test manually**
* Use the search bar on the KB Page, see if it works.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


MdeBoer95:
@reckart searching for concepts and instances should work. I am not sure what would be the best way to include properties. Right know the ```getEntitiesInScope(...)``` method uses a ```ConceptFeatureValueType``` as argument to restrict the type of resource that is searched for. ```ConceptFeatureValueType``` can be ```CONCEPTS```,```INSTANCES``` or ```ANY_OBJECT```. I could just add ```PROPERTY``` to the enum but that wouldn't make much sense for the ```ConceptFeature``` (and it would also require to add something like ```CONCEPTorINSTANCE```) so it doesn't seem like the best option. I could also create a ```ResourceType``` enum with the values ```CONCEPT```, ```INSTANCE```, ```PROPERTY```and then pass a ```Collection<ResourceType>``` to ```getEntitiesInScope(...)``` but that would require to re-factor the concept feature accordingly. Any suggestion or other idea?

reckart:
It's probably ok to search only over concepts and instances for the moment since the number of properties is comparatively small. Let's do it step by step.

MdeBoer95:
>Is the code using the full-text KB search...

No, but I will add it.

MdeBoer95:
Should I use the same layout for displaying the descriptions of the search results that is used in the AnnotationPage ? If so, where can I find the html code for this?

MdeBoer95:
@reckart I might add some smaller changes but you can start to review it, if you want.

MdeBoer95:
@Rentier is there an error when you try with the Wine-KB? I tried exporting and importing a project with Wine-KB and it worked.
I think the issue with UKP Wikidata is not related to this PR. It also happens on master. Sometimes it times out sometimes not.

Rentier:
I do not see an error, I just do not see concepts.

MdeBoer95:
Can you send me the project?

reckart:
Some items that appear in the dropdown work, when I click on some others, nothing happens (testing with lemonUby Verbnet in a local KB). E.g. "Lexical property" loads, but "Lexical entry" does not.

When the search string ends with a space, an exception is thrown (local KB).

When the search string contains * or ?, exceptions may also be thrown (local KB).

I get scrollbars. Looks like the new input field might break the layout.

MdeBoer95:
I also tested with lemonUby and it works for me. Both "Lexical property" and "Lexical entry" load properly.

I think it's ok that this error is thrown, we should just propargate it to the UI to tell the User that his query is not valid.

What exactly do you mean with "I get scrollbars"?

MdeBoer95:
I just saw that the exception is already caught in the sail-lucene code. So I don't know how to propagate the error to the UI.

MdeBoer95:
I would say it is okay to remove them from the input, since most users will probably just type the beginning of the instance name if they want to search for it and not use any wildcards.

reckart:
Well, we can try it and if we come up with a better idea, we change it ;)

MdeBoer95:
If space is typed as first character in the field, the ```String input``` in the code is ```null```. Do we want to consider all entities in that case (i.e. set input string to ```""```) or none (i.e. return an empty list)?

reckart:
Whitespace at the beginning or end of a query should be trimmed before the query is considered further.

MdeBoer95:
I mean if I type only one space and nothing else the input is ```null```. So there is nothing to trim.

reckart:
Ok, then I'd say "no input means nothing to do". KBs tend to have a large amount of items. Listing them all would be infeasible. Actually, do you even get a server-side notification on "no input"? If yes, what we could do is display a list of the 10 most recent items which the user has selected in previous queries.

MdeBoer95:
There is no server-side notification at the beginning when the autocomplete field is selected. Only when a user types only spaces, or  when he types for example one character and removes it so that the field is empty again it sends a notification where the input is ```null```.

PR #820, opened 2019-01-03
#792 - Open imported doc in external search

### LINKED ISSUE #792: A quick way to go to the document annotation after importing from external search
**The problem**
Most likely after importing documents from external search page, users will want to start annotating on these documents. However, it takes several steps to get start annotating; 1. Click on projects, 2. select the project, 3. finding the document just imported from a possibly long list, 4. click on the document, 5. click 'Open'. 

It would be nice to have a quicker way to get to a document just imported directly from the external search page, perhaps with just a single click.

**Possible solution**
Replace the "import" buttons with an "open" button if that document has been imported, which opens it up for annotation when clicked.

shinyichen:
Pull request for feature request #792 A quick way to go to document annotation after importing from external search.

**What's in the PR**
* Added an `Open` button in the `SearchPage` `ResultRowView` that becomes visible when a document has been imported
* The `Open` button opens the document in `AnnotationPage`
* Closes #792

**How to test manually**
*Part 1.*
1. Select a project and go to the external search page.
2. Type in a search keyword and wait for the results.
3. From the results, import a document.
4. The `Import` button should disappear and the `Open` button should appear.
5. Click on the `Open` button, the page should redirect to the Annotation page with the imported document open. Check that the title matches the one imported.

*Part 2.*
1. Go back to external search page.
2. Type in the same search keyword as Part 1 and wait for results.
3. Check that the imported document from Part 1 has the `Open` button visible.
4. Click on the `Open` button to check that it opens the document.

**Automatic testing**
None


inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
```diff
@@ -334,6 +336,15 @@ public ResultRowView(String id, long rowNumber, IModel<ExternalSearchResult> mod
...
                         !documentService.existsSourceDocument(project, documentTitle))));
+            add(new LambdaAjaxLink("openLink", _target -> {
+                PageParameters pageParameters = new PageParameters()
+                    .add(WebAnnoConst.PAGE_PARAM_PROJECT_ID, project.getId())
+                    .add(WebAnnoConst.PAGE_PARAM_DOCUMENT_ID,
+                        documentService.getSourceDocument(project, documentTitle).getId())
+                    .add(WebAnnoConst.PAGE_PARAM_FOCUS, 1);
+                setResponsePage(AnnotationPage.class, pageParameters);
+            }).add(
+                visibleWhen(() -> documentService.existsSourceDocument(project, documentTitle))));
```
rgabbard: We call `documentService.existsSourceDocument` three times in this method. It is probably a cheap operation for most plausible `DocumentService` implementations, but there's no harm in just doing it once and saving the result.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
```diff
@@ -334,6 +336,15 @@ public ResultRowView(String id, long rowNumber, IModel<ExternalSearchResult> mod
             add(new LambdaAjaxLink("importLink", _target -> actionImportDocument(_target, result))
                     .add(visibleWhen(() -> 
                         !documentService.existsSourceDocument(project, documentTitle))));
+            add(new LambdaAjaxLink("openLink", _target -> {
+                PageParameters pageParameters = new PageParameters()
+                    .add(WebAnnoConst.PAGE_PARAM_PROJECT_ID, project.getId())
+                    .add(WebAnnoConst.PAGE_PARAM_DOCUMENT_ID,
+                        documentService.getSourceDocument(project, documentTitle).getId())
```
rgabbard: Does this return the same thing as `result.getDocumentId()`?

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
```diff
@@ -334,6 +336,15 @@ public ResultRowView(String id, long rowNumber, IModel<ExternalSearchResult> mod
...
+                    .add(WebAnnoConst.PAGE_PARAM_FOCUS, 1);
```
rgabbard: Can you add a comment explaining what `PAGE_PARAM_FOCUS` means? It is less obvious than the others.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
rgabbard: When does the `visibleWhen` check get repeated? Can you confirm this updates appropriately when the user imports a document?

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
reckart: The ID in the result is an ID used by the external document repository. The SourceDocument ID is internal to INCEpTION and generated only upon import. While using the repository ID as the document title works, it feels a bit fragile and is probably something that needs to be refactored later on. It might be a good idea to introduce an `externalId` column in the `SourceDocument` class which can be used to carry the original ID a document had in an external repository.

reckart: Probably would need two columns: 
* `source` to identify where the document came from (maybe a string to be flexible)
* `sourceId` to uniquely identify the document within the original source (best also a string)

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
shinyichen: Comment added.
This is the line to focus on when opening in the Annotation page. The annotation page adjust paging so that focused line is centered on the page.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
shinyichen: Followed the suggestion and call `documentService.existsSourceDocument` once.

reckart: The visibility is updated every time the component is re-rendered. As long as either the entire page is reloaded or if the component is added directly or indirectly to an `AjaxRequestTarget`, it is re-rendered. If a component needs to be become visible during an AJAX request, it is necessary that `setOutputMarkupPlaceholderTag(true)` was called on it so that the AJAX request finds the element in the browser-side HTML page and can re-render it.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
rgabbard: @shinyichen : From `AnnotationPage`'s source, it looks like it defaults to the first line anyway:

```
int focus = 0;
if (aFocusParameter != null) {
    focus = aFocusParameter.toInt(0);
}   
```

so does everything work fine if the `PAGE_PARAM_FOCUS` parameter is simply omitted?

Ideally, we'd scroll to wherever within the document the search term was found, but that can wait for a future PR.

shinyichen: I removed it.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/SearchPage.java
rgabbard: Created issue #824 for this refactoring

Issue #348, opened 2018-06-27
deleting features


ClaudiaSchulz:
It would be useful if it was possible to delete features from a project so they don't show up anymore at all (rather than just disabling them)

reckart:
This feature is partially covered by upstream PR https://github.com/webanno/webanno/pull/1096

PR #826, opened 2019-01-05
#348 - deleting features

reckart:
**What's in the PR**
- Refactored DB so that "feature" field of the Recommender becomes a foreign key reference to the AnnotationFeature table
- Placed a cascade-on-delete constraint which deletes the recommender when the corresponding feature is deleted
- The DB refactoring intentionally does not delete the deprecated column - this column should be dropped by a separate changeset in a later version/revision

**How to test manually**
* Create a custom span layer with a custom string feature
* Create a recommender for that that feature
* Delete the feature
* Check if the recommender is still there

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
+++ b/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
@@ -42,7 +42,6 @@ import org.apache.uima.jcas.JCas;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import de.tudarmstadt.ukp.clarin.webanno.api.AnnotationSchemaService;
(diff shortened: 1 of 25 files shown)


Rentier:
@reckart This has conflicts.

reckart:
@Rentier conflicts resolved

reckart:
@Rentier @Sibgha360 thanks for testing. I have changed the DB changset such that the not-null constraint is only added after the actual migration when it is guaranteed that no null values exist in the affected column.

I have tested using the following steps:

* delete local embedded DB
* switch to master branch and rebuild code
* start INCEpTION as standalone application
* create project and a recommender in the project
* stop INCEpTION
* switch to PR branch and rebuild code
* start INCEpTION as standalone application
* check that startup works on and that recommender looks good

I have also tested by resetting my local MySQL DB to the previous DB schema (note that the changeset does not delete the old string-valued column, it only renames it, so going back to the old scheme is feasible), deleting the `20190104-1` row from the `DATABASECHANGELOG`table, and starting the application using the PR branch. This worked for me as well.

Could you please confirm it also works for you now?

PR #834, opened 2019-01-06
#803 - Rendering of multi range annotations

### LINKED ISSUE #803: Rendering of multi range annotations
**Describe the bug**
Currently annotations with multiple ranges will not be rendered correctly in PDF editor. It will only show the annotation part in the first range, missing the rest of the ranges.

**To Reproduce**
1. Create an annotation over sentence boundary in brat editor
2. Change to PDF editor
3. Annotation will only cover parts of the first sentence

**Expected behavior**
Annotations with multiple ranges should be rendered in PDF editor correctly. If two ranges are off only by one or zero it should be rendered as one annotation, else it should be rendered as two annotations.

wpmp:
**What's in the PR**
* Handling of discontinued annotations that contain multiple ranges
* Closes #803 

**How to test manually**
* Create a discontinued annotation (over sentence boundary) in brat editor for a PDF document
* Change to PDF editor
* Annotation will span over sentence boundary

**Automatic testing**
* [X] PR includes unit tests

**Documentation**


wpmp:
@reckart sure!
I will change to start with an offset of 0.
However if there are multiple matches in the document text for the given annotation text I'd increase offset by 5 until there is only one match left.
If there is at the beginning already no match we won't be able to find it, so it is an error.
If it is one match left it is a match between PDFExtract text and documenttext from DKPro.
What do you think?

reckart:
I'm not sure I understand this matching algorithm. 

If you perform an exact substring search over a text and in every step increase the offset by 5, isn't there a high chance that you actually miss matches (like 80% of the time?)

What about the case when there is more than one match but only one of the matches is actually annotated?

Is the algorithm that you are using described somewhere?

Did you consider the use of an approximate/fuzzy matching algorithm or a string alignment (e.g. Bitap) algorithm (e.g. Viterbi).

reckart:
> About the used algorithm: Can I continue with the exact string search for now as a baseline and maybe check later if or which other method for alignment is more suitable?

Going for a baseline first is sensible. The known problems with the baseline should be recorded (in documentation / issues) along with pointers to better alternatives if available. Reserving some time to follow up on at least one of these alternatives is advisable.

wpmp:
@reckart ok, nice, I'll do so.
If you want to review the code you can go for it. :)

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
+        for (VRange range : aVspan.getRanges())
+        {
+            if (range.getBegin() < vSpanBegin) vSpanBegin = range.getBegin();
+            if (range.getEnd() > vSpanEnd) vSpanEnd = range.getEnd();
         }
+
+        // use an offset before and after string to find a unique text snippet for an annotation
+        // begin with 0 and increase by 1 until a unique text is found
+        // offset is limited to length of a quarter of the text
+        for (int offset = 0; offset < docTextLen / 4 + 1; offset++)
```
reckart: Why is the offset limited to 25% of the text?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
+                return new Span(aVspan.getVid().getId(), firstLine.getPage(), aColor,
+                    annotatedText.substring(annotationBegin - index, annotationEnd - index + 1),
+                    firstLine.getPosition(), lastLine.getPosition());
+            }
+        }
+        LOG.error("Could not map existing annotation with id " + aVspan.getVid().toString());
+        return null;
+    }
+
+    private static List<Integer> getAllIndices(String searchIn, String searchFor)
```
reckart: What does this method do?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
+            String annotatedText = aDocumentText.substring(offsetBegin, offsetEnd).replaceAll("\\s", "");
+            String offsetBeforeText = aDocumentText.substring(offsetBegin, vSpanBegin).replaceAll("\\s", "");
+            String offsetAfterText = aDocumentText.substring(vSpanEnd, offsetEnd).replaceAll("\\s", "");
+
+            List<Integer> indices =
+                getAllIndices(aPdfExtractFile.getStringContent(), annotatedText);
+            if (indices.size() == 0) {
+                // if there are no matches end search as adding more characters won't help anymore
+                break;
+            } else if (indices.size() == 1) {
```
reckart: What if the size of indices is larger than 1?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
+                int index = indices.get(0);
+                // get begin and end position of the original annotationText within PDFExtract text
+                int annotationBegin = index + offsetBeforeText.length();
+                int annotationEnd = index + annotatedText.length() - offsetAfterText.length() - 1;
+                // get according PDFExtract file lines for begin and end of annotation
+                PdfExtractLine firstLine = aPdfExtractFile.getStringPdfExtractLine(annotationBegin);
+                PdfExtractLine lastLine = aPdfExtractFile.getStringPdfExtractLine(annotationEnd);
+                return new Span(aVspan.getVid().getId(), firstLine.getPage(), aColor,
+                    annotatedText.substring(annotationBegin - index, annotationEnd - index + 1),
```
reckart: The annotatedText doesn't contain any spaces anymore, right? What is the excerpt that you create from the annotatedText here used for?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
+                    firstLine.getPosition(), lastLine.getPosition());
+            }
+        }
+        LOG.error("Could not map existing annotation with id " + aVspan.getVid().toString());
```
reckart: I kind of assume that the case indices.size() > 1 means that there were multiple places where the annotation could match. It would be good to give different messages depending on whether the annotation could not be rendered because there were no matches and because there were multiple matches. The message should be displayed in the UI, not only in the error log - otherwise the user has no idea what is happening.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -108,35 +110,62 @@ public static PdfAnnoModel render(AnnotatorState aState, VDocument aVDoc, String
...
-            return new Span(aVspan.getVid().getId(), first.getPage(), aColor,
-                cleanAnnotatedText.substring(start - index, end - index).replaceAll("\\s", ""),
-                first.getPosition(), last.getPosition());
+        int docTextLen = aDocumentText.length();
+        int vSpanBegin = aDocumentText.length();
+        int vSpanEnd = -1;
+        // search for begin of the first range and end of the last range and use those
+        for (VRange range : aVspan.getRanges())
+        {
+            if (range.getBegin() < vSpanBegin) vSpanBegin = range.getBegin();
```
reckart: What is the purpose of these if clauses?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: I thought it wouldn't be possible to find a unique mapping if there were two or more possible candidates found when offset reaches 25% pre and post annotated text however I just figured out negative examples.
It was a misconception of mine.
I will change the limit to document text length.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: It searches for all starting indices of a String `searchFor` in the String `searchIn`.
Should I add Javadoc?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Search is continued by increasing offset because there is no unique match found but multiple possible matches.
With increased offset chances are better to find a unique match.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Actually, you don't have to find all indexes, right? There are only three cases you currently handle:
* no match
* exactly one match
* more than one match (once you find the second match, the search can be aborted because it is already an error case)

You could for example change the method such that it throws an exception with a suitable error message in the no-match and more-than-one-match case.

reckart: Adding JavaDoc would be good and renaming the method might also help (e.g. `findMatch`).

wpmp: Yes, you are right, I'll do it like that.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Bad variable naming of mine... :/
Actually the variable `annotatedText` also already contains the offset text, which then needs to be removed.
I will change it to only contain the annotatedText.

reckart: Ok, but what is it used for? I mean, the text doesn't contain any spaces... what is it good for without the spaces?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Do you have unit tests for the matching algorithm?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Finding the lowest begin and the highest end of the ranges.
Multiple ranges for one annotation in INCEpTION only occur when annotation crosses sentence boundary, right?
So what we do here is basically search for the begin and end of all ranges together so we can create one annotation for it.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Yes that's right.
If I'm not mistaken we don't need an error message for multiple matches.
If there are multiple matches we will find a unique match at latest when we search for annotation with whole document as context (with changes discussed earlier about offset limit to document text length).

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Whops - I thought I had cancelled this comment instead of submitting it. It took me a moment to understand what the code does and after that I also understood the line comment above the for loop. Maybe you would like to "simplify" the code a bit to something like (no idea if you think this is simpler):

```
        int begin = span.getRanges().stream()
            .mapToInt(range -> Math.min(range.getBegin(), range.getEnd()))
            .min().getAsInt();

```

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Ah... I think *now* I understand why earlier you increased this one value by 5 in every step. That wasn't a search offset but a context window size, wasn't it? I suppose increasing the window size by an increment > 1 could indeed make sense.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: It is for producing the Anno file for PDFAnno and the format for this file is like

```
[[spans]]
id = "1"
page = 1
label = "label"
text = "someannotatedtext"
textrange = [22,38]
```
It doesn't contain spaces, as it uses PDFExtract output without spaces.
It seems PDFAnno is not aware of spaces at all, and it does not need them because it knows the position of the characters from the PDFExtract output.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Sure, seems good :)

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: Not yet implemented.
That's something I wanted to talk to you but haven't done yet.
I wanted to write tests for the `render` method in `PdfAnnoRenderer` class.
I had a look at brat tests and wanted to do it similar however I encountered some problems which I was not able to resolve and therefore didn't include the tests for now.
Maybe we can discuss in gitter, but I need to have a look at this again.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: So does it use the text? Since it has the range, the text seems redundant, doesn't it?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Ping me on Gitter when you wish to discuss.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: Make it a constant with a speaking name like `WINDOW_SIZE_INCREMENT` - sure, 5 sounds good. Be sure to handle cases where the match is near the document boundaries and where you actually may not find the full number of chars before/after (or both).

wpmp: Ok, I will also rename `offset` to e.g. `windowSize`.
I already take care of boundaries when setting `offsetBegin` and `offsetEnd`. :)

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: I also wondered why the text is included there.
But I think it is useful when exporting annotation froms PDFAnno.
In our case it is useless I guess.
I didn't check, but at some points while developing it seemed that the content of `text` field is not relevant to PDFAnno and it could contain wrong data but still find the correct annotation position.
I am not sure if it is required at all and would cause errors if the field is completely missing.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: @reckart Should a unit test for the `render` method be added in this PR?

reckart: Would be good.

wpmp: @reckart seems like the issue for the test disappeared.. don't know exactly why this is the case, however it is gone.
I copied over `tcf04-karin-wl.xml` from Webanno to test resources folder.
Is this ok or should I somehow import it from the webanno-brat package?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
wpmp: @reckart Should I check out if PDFAnno can import the anno file without `text` fields or can I resolve this conversation?
Even if it would work without it I'd rather keep it in because it is the actual format.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: You can copy it, no problem.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
reckart: I don't have a strong opinion on it. If you are curious, test it - otherwise resolve.

wpmp: Seems to work without the text, however when pressing export button the text is missing.
I will leave it in for now but keep it in mind.
Maybe this can be useful for performance improvement or other stuf later on if required.

reckart: Add a comment to the source code explaining why this is added / what happens if it is omitted.

wpmp: Added it to the Span class, containing the `text` field and also producing the anno file span string.

wpmp:
@reckart I think this PR is now ready for another review :)

Issue #833, opened 2019-01-06
Error Page when saving Concept Feature traits


MdeBoer95:
**Describe the bug**
When the save button for the concept feature traits is clicked a NullPointerException is thrown and the ErrorPage is shown.

**To Reproduce**
1. Go to Layer configurations and select NamedEntity layer. Try to save configurations for the identifier feature.
2. See error

**Expected behavior**
The feature traits should be saved without any error

**Please complete the following information:**
 - Version and build ID: [ Technische Universität Darmstadt -- Computer Science Department -- INCEpTION 𝜶 -- 0.7.0-SNAPSHOT (, build )  ]
 - OS: [Linux]
 - Browser: [chrome]

MdeBoer95:
https://github.com/inception-project/inception/blob/64811fa301e3838b6e828166e078c8ca55079ce3/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureTraitsEditor.java#L69-L94

@reckart When I look at it in the debugger the model object of ```feature```(l.69) is ```null``` in ```onSubmit()``` but not when when the ConceptFeatureTraitsEditor is created. So the value is set properly but "suddenly" ```null``` when  ```onSubmit()``` is called.
Any idea why that happens?

reckart:
@MdeBoer95 no idea. These fields can be marked with `final` and the compiler doesn't complain - so if they start out with a non-null value, that value should never change. Try making the fields final and add a `Validate.notNull(...)` to ensure the field doesn't start out non-null.

MdeBoer95:
Still not sure why exactly it happens but I have a workaround in PR #835

reckart:
I have just hit this problem with a `LinkFeature` - so it is a general problem, not one that is to be fixed in the ConceptFeature only.

```
java.lang.NullPointerException: null
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.feature.SlotFeatureSupport.writeTraits(SlotFeatureSupport.java:284) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.feature.editor.LinkFeatureTraitsEditor.writeTraits(LinkFeatureTraitsEditor.java:169) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.feature.editor.LinkFeatureTraitsEditor.access$0(LinkFeatureTraitsEditor.java:161) ~[classes/:?]
[...]

PR #835, opened 2019-01-06
#833 - Error Page when saving Concept Feature traits

MdeBoer95:
**What's in the PR**
* Fix bug described in #833 

**How to test manually**
* save traits for concept feature, see if bug is fixed

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
This PR obtains the value of the feature model and then uses that value in the editor. That means if the feature in the model changes, the editor won't notice it. When we pass a model to a component, we would expect the component to always obtain the current value of the model whenever it is needed (that was the previous code). 

So if the value of the model (not the model itself) becomes `null` and this causes an error, then I would guess that there is a bug somewhere else - not in the editor component.

reckart:
See `FeatureDetailForm` lines 135 and following: If the feature type has changed, we need to set up a new traits editor - that means if the feature type does not change when switching between features, then the editor is reused - only the model values are updated. With your workaround, that would break. But at the moment, it doesn't explain to me under which circumstances the model value can become `null` when the editor is still visible/used.

reckart:
Rejecting workaround. As mentioned in #833, this happens with other types of features as well and thus needs to be fixed at a different level/location.

Issue #825, opened 2019-01-04
Split-screen external search and annotation view


rgabbard:
**Is your feature request related to a problem? Please describe.**

The current workflow for external search splits searching and annotation into two phases: first you search for documents and import them, then you go open them for annotation.  This is inconvenient for workflows which depend on a faster back-and-forth between search and annotation.

**Describe the solution you'd like**

I would like to have a two-paned view, where a narrow left-hand pane had the search term box and the right-hand pane had the usual annotation interface open to the selected document. Ideally, the search query would be communicated to the annotation view in order to automatically scroll to the first matching portion of the document.

![img_20190104_130536 1](https://user-images.githubusercontent.com/7139661/50703430-1e9c4100-1022-11e9-8b49-7fd026d33751.jpg)

Details:
* presumably a document would be automatically imported into the document repository when its search result was clicked

rgabbard:
(if we could implement this, it would remove the need for #792 and #793 as far as our particular project was concerned)

shinyichen:
Would this page only handle external searches, leaving the internal search and annotation page the way it is? If so I think it sounds feasible.

I like this idea. Just throwing out some thoughts that came to mind:
1. After a search and you start to annotate, which I am not sure how long that usually takes, you probably want to maintain this list so you can come back to. Say if you navigate away from search, want to perform another search or want to come back to work on it later. Although you could just perform the same search again (from my observation it seems the results and orderings are always the same with the search), this workflow seems odd. I'd like to come back to the same set of documents, but I don't want to assume a search always returns the same results in same order. Or we don't really care about that?
2. Would it be confusing to have two places to annotate (from external search page and from annotation page).

rgabbard:
@shinyichen : As far as ISI is concerned, we only care about external search.  Someone else might want to extend it to internal search in the future.

re: 1 - I think the idea is that the search result list on the left pane is persistent - so you click one document, annotate it in the right pane, click to the next document in the left pane, etc.

re: 2 - I imagine rather than replacing the existing interfaces (which work well for many applications), we would want to add some sort of new option for the hybrid interface

rgabbard:
(I should note this is the sort of interface we used when we implemented the original curated training annotation tool)

reckart:
> re: 2 - I imagine rather than replacing the existing interfaces (which work well for many applications), we would want to add some sort of new option for the hybrid interface

With "existing" you mean the full-screen search pane?

[...]

PR #846, opened 2019-01-11
#825 external search bar - keyword highlight

shinyichen:
#825 External Search Side bar - Keyword Highlight

**What's in the PR**
This is an additional feature in the #825 to add keyword highlight to annotation page

**How to test manually**
1. Select a project and go to Annotation section from the project dashboard.
2. From the sidebar, select external search.
3. Do a search in the external search.
4. Click on a result.
5. When the annotation appears, there should be keywords highlighting in the annotation page.
6. The keywords highlighted should match the same sentence and same keyword highlight in the result row.

**Automatic testing**
None


rgabbard:
Can you change the base for this MR to `isi-vista:feature/825-external-search-sidebar` so it is easier to see this differences specific to this issue?

shinyichen:
@reckart After I pulled master branch into this branch, there is a problem with `DocumentImporterImpl.importDocumentFromDocumentRepository` where `externalSearchService` is null. `DocumentImporterImpl` was changed by #908

reckart:
@shinyichen Problem was that you injected the `DocumentImporter` not by interface but by implementation class into the external search sidebar. Changing the field type from `DocumentImporterImpl` to `DocumentImporter` fixed the problem.

shinyichen:
Thank you finding that. @reckart

shinyichen:
There are some unused declared dependencies in elastic search module causing build to fail. Are these ok to delete?

screen shot 2019-02-01 at 2 06 18 pm

reckart:
@shinyichen I guess yes. I deleted a test class which was causing me problems and which was never really used. This must have freed up these depencencies.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
```diff
@@ -20,27 +20,38 @@
...
 import de.tudarmstadt.ukp.inception.externalsearch.elastic.model.ElasticSearchHit;
 import de.tudarmstadt.ukp.inception.externalsearch.elastic.model.ElasticSearchResult;
 import de.tudarmstadt.ukp.inception.externalsearch.elastic.traits.ElasticSearchProviderTraits;
+import de.tudarmstadt.ukp.inception.support.annotation.OffsetSpan;
 
 public class ElasticSearchProvider
     implements ExternalSearchProvider
 {
 
+    private static final String highlight_start_tag = "<em>";
```
reckart: Constant names should be all capitals.

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
```diff
@@ -0,0 +1,46 @@
...
+package de.tudarmstadt.ukp.inception.support.annotation;
+
+import java.io.Serializable;
+
+public class OffsetSpan implements Serializable
+{
+
+    private static final long serialVersionUID = -2532025576874280981L;
+
+    private int start;
```
reckart: Since there are not setters, these fields can be final.

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private int endExclusive;
+
+    public OffsetSpan(int aStart, int aEndExclusive) {
+        start = aStart;
+        endExclusive = aEndExclusive;
+    }
+
+    public int getStartOffset()
```
reckart: We usually call these `getBegin()`.

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return start;
+    }
+
+    public int getEndOffsetExclusive()
```
reckart: We usually call these `getEnd()`. Ranges in UIMA are always half-open `[begin,end)`. If you want to make that explicit, I'd suggest you add it as JavaDoc instead of using a verbose name.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/sidebar/ExternalSearchAnnotationSidebar.java
```diff
@@ -424,5 +461,15 @@ public void setCurrentPage(long aCurrentPage)
         {
             currentPage = aCurrentPage;
         }
+
+        public ExternalSearchResult getSelectedResult()
+        {
+            return selectedResult;
+        }
+
+        public void setSelectedResult(ExternalSearchResult selectedResult)
```
reckart: `aSelectedResult` and no `this`.

inception-external-search-core/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/ExternalSearchHighlight.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final long serialVersionUID = 4044755133816131842L;
+
+    private String highlight = null;
+
+    private List<OffsetSpan> offsets;
+
+    public ExternalSearchHighlight(List<OffsetSpan> aOffsets) throws IllegalArgumentException
+    {
+        if (aOffsets == null || aOffsets.isEmpty()) {
```
reckart: Could use `org.apache.commons.lang3.Validate.notEmpty(T, String, Object...)` to check the argument.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
```diff
@@ -126,9 +137,49 @@ public boolean isConnected()
...
+                    String highlight_clean = highlight.replace(highlight_start_tag, "")
+                        .replace(highlight_end_tag, "");
+
+                    // find the matching highlight offset in the original text
+                    int highlight_start_index = originalText.indexOf(highlight_clean);
+
+                    // find offset to all keywords in the highlight
+                    // they are enclosed in <em> </em> tags in the highlight
+                    String highlightTemp = highlight;
+                    ArrayList<OffsetSpan> offsets = new ArrayList<>();
```
reckart: Variable type should be `List<OffsetSpan>`.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
shinyichen: Fixed.

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
shinyichen: Updated to final

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
shinyichen: Renamed.

inception-support/src/main/java/de/tudarmstadt/ukp/inception/support/annotation/OffsetSpan.java
shinyichen: Renamed.

inception-ui-external-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/externalsearch/sidebar/ExternalSearchAnnotationSidebar.java
shinyichen: Updated.

inception-external-search-core/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/ExternalSearchHighlight.java
shinyichen: Changed. Now using `commons.lang3.Validate.notEmpty`.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
shinyichen: Changed to `List`

reckart:
@shinyichen @rgabbard The highlighting didn't work for me at all. 59eb12d fixes the problem for me. Can you please check if it also still works for you? Also, I made some generate improvements to the sidebar code in 351baf9.

If you are good with the changes, we can merge this.

rgabbard:
@reckart : Hm! In my experimenting just now selecting a document in the sidebar did not switch to it in the annotation view.  Are you seeing the same behavior?  I'll check into it on Monday.

reckart:
@rgabbard in general it worked for me, although I also had the feeling it sometimes might have gotten stuck. Try reloading the browser (F5) and see if it works then. Mind that if the document is already open, nothing is switched. I noticed switches, but I did not notice a jump to the first match - would I think would be convenient.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
```diff
@@ -126,9 +137,49 @@ public boolean isConnected()
...
+                        int start = highlight_start_index +
+                            highlightTemp.indexOf(HIGHLIGHT_START_TAG);
+                        highlightTemp = highlightTemp.replaceFirst(HIGHLIGHT_START_TAG, "");
+                        int end = highlight_start_index +
+                            highlightTemp.indexOf(HIGHLIGHT_END_TAG);
+                        highlightTemp = highlightTemp.replaceFirst(HIGHLIGHT_END_TAG, "");
+                        offsets.add(new OffsetSpan(start, end));
+                    }
+
+                    if (!offsets.isEmpty()) {
```
reckart: @rgabbard @shinyichen Here you fully skip highlights for which you cannot find a document position. I think this is too aggressive. I am planning to change this in https://github.com/inception-project/inception/pull/949 such that it is allowed to have `ExternalSearchHighlight`s without offset information. They can still render as highlights on the external search page and in the external search sidebar, even if the matches cannot be highlighted in the actual text. The reason is that there may be cases (e.g. the PubAnnotation connector) where I would like to show the highlights in the search results but at the time, I don't actually have access to the full text and therefore cannot calculate the offsets.

rgabbard: @reckart : Your change seems reasonable to me.

Issue #847, opened 2019-01-11
Fix/improve default KB profiles


reckart:
**Describe the bug**
Some of the settings in the default KB profiles are incorrect, e.g. when using the local FTS mode for remote KB which do not support it. Other KB profiles can be improved, e.g. by adding root concepts.

**To Reproduce**
* E.g. try using FTS search on the GND profile
* The GND profile could use an explicit root concept definition to speed up access

**Expected behavior**
KB profiles should deliver the best-possible user experience out of the box.

**Please complete the following information:**
 - Version and build ID: 4f601943fc586a9c57e874c0ecd7ba6c4f378f25

PR #851, opened 2019-01-12
#847 - Fix/improve default KB profiles

reckart:
**What's in the PR**
- Disable FTS for ZBW endpoints
- Define explicit root for GND and YAGO
- Improve mapping for YAGO
- Avoid line-breaks in the concept tree
- Allow owl:Thing to be used as a root concept
- Handle decimal values using the numeric editor

**How to test manually**
* Create KBs from the updated profiles
* Try browsing on the KB page
* Try linking an entity mention to a concept from them

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Forcing this in as it is blocking the release (YAGO tests giving problems during release because root-concept query times out).

Issue #860, opened 2019-01-14
Tasks are added even if same kind is already in queue


Rentier:
**Describe the bug**
When e.g. using recommenders, new training tasks are scheduled even though there is already one in the queue.

**To Reproduce**
1. Create a project with recommenders that need some time to train
2. Quickly annotate things
3. The queue of the task scheduler runs full

**Expected behavior**
Tasks are only added if there is not already one of the same type in the queue

Issue #780, opened 2018-12-13
Wicket `@SpringBean` injection generates illegal access warning on recent JDKs


reckart:
**Describe the bug**
Wicket `@SpringBean` injection generates illegal access warning on recent JDKs.

**To Reproduce**
1. Add `--illegal-access=debug` to the VM arguments
2. Run INCEpTION on a recent JDK (9 or higher)
3. Access any page which uses bean injection, e.g. the annotation page
4. See message in log
```
WARNING: Illegal reflective access by net.sf.cglib.core.ReflectUtils$1 (file:.../cglib-3.2.6.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
	at net.sf.cglib.core.ReflectUtils$1.run(ReflectUtils.java:61)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at net.sf.cglib.core.ReflectUtils.<clinit>(ReflectUtils.java:52)
	at net.sf.cglib.core.KeyFactory$Generator.generateClass(KeyFactory.java:243)
	at net.sf.cglib.core.DefaultGeneratorStrategy.generate(DefaultGeneratorStrategy.java:25)
	at net.sf.cglib.core.AbstractClassGenerator.generate(AbstractClassGenerator.java:329)
	at net.sf.cglib.core.AbstractClassGenerator$ClassLoaderData$3.apply(AbstractClassGenerator.java:93)
	at net.sf.cglib.core.AbstractClassGenerator$ClassLoaderData$3.apply(AbstractClassGenerator.java:91)
	at net.sf.cglib.core.internal.LoadingCache$2.call(LoadingCache.java:54)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at net.sf.cglib.core.internal.LoadingCache.createEntry(LoadingCache.java:61)
	at net.sf.cglib.core.internal.LoadingCache.get(LoadingCache.java:34)
	at net.sf.cglib.core.AbstractClassGenerator$ClassLoaderData.get(AbstractClassGenerator.java:116)
	at net.sf.cglib.core.AbstractClassGenerator.create(AbstractClassGenerator.java:291)
	at net.sf.cglib.core.KeyFactory$Generator.create(KeyFactory.java:221)
	at net.sf.cglib.core.KeyFactory.create(KeyFactory.java:174)
	at net.sf.cglib.core.KeyFactory.create(KeyFactory.java:153)
[...]

PR #862, opened 2019-01-14
#860 - Tasks are added even if same kind is already in queue

Rentier:
**What's in the PR**
- Do not enqueue tasks if task of the same kind is already queued
- Add Spring configuration for scheduler properties

**How to test manually**
0. Maybe reduce the value for `scheduler.threads` to 1 or 2
1. Create a project with a slow running recommender
2. Annotate things very quickly
3. Look in the log whether you see `Task already in queue`

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-app-webapp/src/main/resources/application.properties
+++ b/inception-app-webapp/src/main/resources/application.properties
@@ -10,8 +10,8 @@ backup.keep.time=0
 backup.interval=0
 backup.keep.number=0
 
-scheduler.threads=8
-scheduler.queuesize=200
+scheduler.threads=4
+scheduler.queuesize=100
 
 debug.casDoctor.checks=
 debug.casDoctor.repairs=
--- a/inception-app-webapp/src/main/resources/application.properties
+++ b/inception-app-webapp/src/main/resources/application.properties
@@ -10,6 +10,9 @@ backup.keep.time=0
 backup.interval=0
 backup.keep.number=0
 
+scheduler.threads=8
+scheduler.queuesize=200
+
 debug.casDoctor.checks=
 debug.casDoctor.repairs=
 debug.casDoctor.fatal=false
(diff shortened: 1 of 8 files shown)


Rentier:
This also needs to be merged into master I think.

inception-app-webapp/src/main/resources/application.properties
```diff
@@ -10,6 +10,9 @@ backup.keep.time=0
 backup.interval=0
 backup.keep.number=0
 
+scheduler.threads=4
```
reckart: Should be mentioned in the documentation.

inception-app-webapp/src/main/resources/application.properties
```diff
@@ -10,6 +10,9 @@ backup.keep.time=0
...
+scheduler.queuesize=100
```
reckart: `queueSize`?

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
```diff
@@ -36,19 +37,18 @@
...
-
     private final ApplicationContext applicationContext;
     private final ThreadPoolExecutor executor;
 
     private final List<Task> runningTasks;
 
     @Autowired
-    public SchedulingService(ApplicationContext aApplicationContext)
+    public SchedulingService(ApplicationContext aApplicationContext,
+                             @Value("${scheduler.threads:4}") int numberOfThreads,
```
reckart: Hm... I'm always using a dedicated class to contain the configuration (e.g. `KnowledgeBaseProperties`)... I tend to find that more "scalable" when adding more settings in the future. Due to #780 I am also now switching all of these classes to an interface/impl pair.

reckart:
I usually merge the maintenance branch into the master branch after integrating a PR into the maintenance branch. Otherwise we'd need to do two PRs and two reviews...?

Rentier:
We just need one, I am just saying that we just should not forget it.

Rentier:
Sry for the closing,

PR #869, opened 2019-01-15
#838 - Creation of spans in PDF editor

### LINKED ISSUE #838: Creation of spans in PDF editor
In PDF editor it is not possible to add span annotations.

When selecting text in the document in PDF editor a span annotation should be created.
If it is not possible the user should be informed about the error.

wpmp:
**What's in the PR**
Possibility to create span annotations within PDF editor view.
Closes #838 

**How to test manually**
1. Open a PDF file in PDF editor view
2. Select a layer
3. Select some text in PDF
4. Annotation should appear after a short time
5. Annotation should also appear when changeing to brat editor

**Automatic testing**
* [X] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
+++ b/inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
@@ -125,7 +125,8 @@ public class PdfAnnoPanel
                 try
                 {
                     String pdftext = PDFExtractor.processFileToString(pdfFile, false);
-                    PdfAnnoModel pdfAnnoModel = aPdfAnnotationEditor.renderPdfAnnoModel(pdftext);
+                    PdfAnnoModel pdfAnnoModel =
+                        aPdfAnnotationEditor.renderPdfAnnoModel(aTarget, pdftext);
                     String script = getAnnotationsJS(pdfAnnoModel);
                     aTarget.appendJavaScript(script);
                 }
@@ -135,8 +136,8 @@ public class PdfAnnoPanel
                         + "with PDFExtractor.", e);
                     error("Unable to get PDF text for " + pdfFile.getName()
                         + "with PDFExtractor.");
+                    aTarget.addChildren(getPage(), IFeedback.class);
                 }
-                aTarget.addChildren(getPage(), IFeedback.class);
             }
         });
 
@@ -154,7 +155,8 @@ public class PdfAnnoPanel
                     boolean success =
                         aPdfAnnotationEditor.createSpanAnnotation(aTarget, getRequest(), pdftext);
                     if (success) {
-                        PdfAnnoModel model = aPdfAnnotationEditor.renderPdfAnnoModel(pdftext);
+                        PdfAnnoModel model =
+                            aPdfAnnotationEditor.renderPdfAnnoModel(aTarget, pdftext);
                         String script = getAnnotationsJS(model);
                         aTarget.appendJavaScript(script);
                     }
@@ -163,8 +165,8 @@ public class PdfAnnoPanel
                 {
                     log.error("Unable to get PDF text for [{}]", pdfFile.getName()
                         + "with PDFExtractor.", e);
+                    aTarget.addChildren(getPage(), IFeedback.class);
                 }
-                aTarget.addChildren(getPage(), IFeedback.class);
             }
         });
 
--- a/inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
+++ b/inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
@@ -105,7 +105,7 @@ public class PdfAnnoPanel
                 }
                 catch (IOException e)
                 {
-                    log.error("Unable to get PDF text for " + pdfFile.getName()
+                    log.error("Unable to get PDF text for [{}]", pdfFile.getName()
                         + "with PDFExtractor.", e);
                 }
             }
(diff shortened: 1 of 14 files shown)


reckart:
@wpmp I have tried in Safari and Chrome, but when I select text - nothing happens. Well, the I see the text being marked with the mouse, but I don't see an annotation appear after a short while. I note no errors in the JS console either.

reckart:
Looks pretty cool already. Some issues:

* Annotation recommendations appear at the wrong positions depending on the "page" selected in the action bar of the annotation page and not on the entire document depending on the "window size" configured in the annotation preferences
* Not possible to select another annotation - annotations only load if they are created
* Consequently, deleting annotations other than the one that was just created doesn't work because in order to delete an annotation, it needs to be possible to select it (to open it  in the right sidebar).
* Error message about annotations that cannot be matched appear only when the next successful annotation is made - message should not only include the ID but also the text covered by the annotation
* Creating an annotation triggers two subsequent AJAX calls and this makes the feedback messages  from the first AJAX call (e.g. success message) disappear when the second AJAX call completes

wpmp:
>     * Annotation recommendations appear at the wrong positions depending on the "page" selected in the action bar of the annotation page and not on the entire document depending on the "window size" configured in the annotation preferences

I would tackle this in another issue / PR

>     * Not possible to select another annotation - annotations only load if they are created

Also new issue / PR

>     * Consequently, deleting annotations other than the one that was just created doesn't work because in order to delete an annotation, it needs to be possible to select it (to open it  in the right sidebar).

Along selection or new issue / PR

>     * Error message about annotations that cannot be matched appear only when the next successful annotation is made - message should not only include the ID but also the text covered by the annotation

Should be fixed now

>     * Creating an annotation triggers two subsequent AJAX calls and this makes the feedback messages  from the first AJAX call (e.g. success message) disappear when the second AJAX call completes

Should also be fixed now

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -157,11 +158,11 @@ private static Span convertToPdfAnnoSpan(VSpan aVSpan, String aColor, String aDo
             } catch (NoMatchFoundException e) {
                 // if no match is found stop search here. increasing context won't help
                 LOG.error("Could not find a match for existing annotation with id "
-                    + aVSpan.getVid().toString());
+                    + aVSpan.getVid().toString() + " and text \"" + annotatedText + "\"");
```
reckart: When logging via SLF4J, use placeholders instead of String concatenation:

```
LOG.error("Could not find a match for existing annotation with id [{}] and text [{}]", aVSpan.getVid(), annotatedText);
```

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoRenderer.java
```diff
@@ -157,11 +158,11 @@ private static Span convertToPdfAnnoSpan(VSpan aVSpan, String aColor, String aDo
...
                 break;
             }
         }
-        return null;
+        return new Span(aVSpan.getVid().getId(), -1, null, annotatedText, -1, -1);
```
reckart: Why use only the ID part of the VID and not the entire VID? The ID alone is not guaranteed to be unique. E.g. an annotation might have the same ID as a suggestion.

wpmp: Oh, I wasn't aware of that.
Do I get the entire VID via `toString` method?

wpmp:
@reckart whoops, I messed up the last commit message :/
What can I do to fix it?
I can only think of `git commit --amend` and a force push but I don't think that you want this?

reckart:
You can force push to PR branches - that's ok.

wpmp:
Done.
PR should be fine for another review when you are ready

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
```diff
@@ -124,26 +126,45 @@ protected void respond(AjaxRequestTarget aTarget)
...
                 catch (IOException e)
                 {
-                    log.error("Unable to get PDF text for " + pdfFile.getName()
+                    log.error("Unable to get PDF text for [{}]", pdfFile.getName()
                         + "with PDFExtractor.", e);
                     error("Unable to get PDF text for " + pdfFile.getName()
                         + "with PDFExtractor.");
-                    aTarget.addChildren(getPage(), IFeedback.class);
                 }
+                aTarget.addChildren(getPage(), IFeedback.class);
```
reckart: I am observing two Wicket AJAX requests when making an annotation. I do not know where the second request comes from because the rendered annotations already appear to be contained in the first one. Maybe you know?

Anyway, the second request contains the feedback panel which causes success/error messages to disappear prematurely. This can be alleviated by adding the feedback panel to the AJAX target only in the `catch` part of the the two exception handling blocks when you actually add a feedback message instead of adding in unconditionally in every request.

wpmp: It seems that those second request does not occur if the next annotation is in the same sentence as the previous one was created in.
If it is two different sentences the request occurs.
I have no clue where it is coming from.
Maybe because in brat this sentence needs to be rerendered or so?

I cannot move it into the `catch` block because in used methods `renderPdfAnnoModel` and `createSpanAnnotation` of `PdfAnnoEditor` `error` method is also used where no exceptions are thrown.

reckart: Well, the error method is probably not called unconditionally there either, right? So you can repeat there adding the feedback panel to the target. It is no problem if it is added in multiple places.

Issue #870, opened 2019-01-16
IRI info for IRI-valued statements


reckart:
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to access the IRI of an IRI-valued statement in the KB without having to click on the statement and to load the target (in particular since there is no convenient way to "go back").

**Describe the solution you'd like**
We already have the black "i" symbol which opens a tooltip with the IRI for the property labels in the statement groups. The same symbol could be added to IRI-valued statements.

PR #871, opened 2019-01-16
#870 - IRI info for IRI-valued statements

reckart:
**What's in the PR**
- Factor IRI badge out into a re-usable class
- Add IRI badge to IRI-value statements in presentation mode
- Move IRI badge to the front of property labels in the statement group so they are all nicely aligned
- Optimized the loading of the KBObject in the IRI Value presenter to avoid unnecessary KB queries

**Note:** Made this a 0.7.1 feature since it is small and very useful.

**How to test manually**
- Open KB
- Look for IRI-valued statments
- Click on IRI badge to see if IRI tooltip appears and shows the expected value

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/IriInfoBadge.html
+++ b/inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/IriInfoBadge.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<!--
+ - Copyright 2019
+ - Ubiquitous Knowledge Processing (UKP) Lab
+ - Technische Universität Darmstadt
+ - 
+ - Licensed under the Apache License, Version 2.0 (the "License");
+ - you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


reckart:
Review anybody? I'm afraid that having provided the PR, I cannot review it myself.

MdeBoer95:
Just out of curiosity. Is there any benefit from checking against ```null``` compared to checking if an Optional value ```isPresent()```? (see changes in ```IRIValuePresenter```)

reckart:
> Just out of curiosity. Is there any benefit from checking against null compared to checking if an Optional value isPresent()? (see changes in IRIValuePresenter)

Puh, maybe not. When working at the Wicket layer, I'm used to working with serializable objects (Optional is not serializable) and with the fact that wicket components/models can usually deal with `null` but have no idea how to treat an `Optional`. I'd have to review in detail again to figure out whether in this particular case it would be possible to use `Optional` and to which extent.

reckart:
@MdeBoer95  Thanks for the review :)

Issue #874, opened 2019-01-17
Not removing recommendations when stacked entities are allowed


qiuwei:
**Is your feature request related to a problem? Please describe.**
When stacked entities are allowed in the projection configs, do not remove overlapping recommendations when one is accepted. 

**Describe the solution you'd like**
In this case, it might better default to keeping all of the recommendations or at least make it configurable.

PR #876, opened 2019-01-17
#874 - Not removing recommendations when stacked entities are allowed

reckart:
**What's in the PR**
- When stacking is enabled, only hide recommendations that have the same labels as already existing annotations overlapping with the suggestion. Alternative labels are still suggested.

**How to test manually**
* enable stacking on the named entity layer
* create a string matching recommender on the value feature
* add annotations to a document such that some of the annotations have a different value but annotate the same word

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@qiuwei I wonder - could you test-drive this PR to check if it fits your needs?

Issue #877, opened 2019-01-17
Recommendations not generated when opening a document


reckart:
**Describe the bug**
When opening a document, recommendations do not appear. This was always the case because the predictions may take a moment, but a reload of the page (F5 or open a sidebar) would pick them up. This is no longer the case.

**To Reproduce**
1. Configure a project with a recommender (string matching is sufficient)
2. Create a few annotations such that the recommender generates suggestions
3. Restart application
4. Open the document again

**Expected behavior**
Once the background training is complete, pressing F5 should cause a re-render which picks up any suggestions that have been created.

**Screenshots**
This is what I see in the logs:

*Open document from dashboard*
```
2019-01-17 11:04:13 INFO [admin] AnnotationPage - BEGIN LOAD_DOCUMENT_ACTION at focus 0
2019-01-17 11:04:13 DEBUG [admin] CasStorageServiceImpl - Reading annotation document [Wikipedia-Obama.txt] (126) for user [admin] in project [A Recommendation playground (NER)] (40)
2019-01-17 11:04:15 DEBUG [admin] CasStorageServiceImpl - Loaded CAS [126,admin] from disk and stored in cache
2019-01-17 11:04:16 INFO [admin] AnnotationSchemaServiceImpl - Upgraded CAS of user [admin] for document [Wikipedia-Obama.txt](126) in project [A Recommendation playground (NER)](40)
2019-01-17 11:04:16 DEBUG [admin] CasStorageServiceImpl - Preparing to update annotations for user [admin] on document [Wikipedia-Obama.txt](126) in project [A Recommendation playground (NER)](40)
2019-01-17 11:04:16 INFO [admin] CasStorageServiceImpl - Updated annotations for user [admin] on document [Wikipedia-Obama.txt](126) in project [A Recommendation playground (NER)](40)
2019-01-17 11:04:16 INFO [admin] CasStorageServiceImpl - Removed surplus history file [admin.ser.1547717076923.bak] of user [admin] for document [Wikipedia-Obama.txt](126) in project [A Recommendation playground (NER)](40)
2019-01-17 11:04:16 INFO [admin] EventRepositoryImpl - LoggedEvent [created=Thu Jan 17 11:04:16 CET 2019, event=AfterAnnotationUpdateEvent, user=admin, project=40, document=126, annotator=admin]
2019-01-17 11:04:16 INFO [SYSTEM] SelectionTask - [admin][[Named entity@value] String Matcher]: Evaluating...
[...]

PR #881, opened 2019-01-17
#877 - Recommendations not generated when opening a document

Rentier:
**What's in the PR**
* Start TrainingTask instead of PredictionTask

**How to test manually**
1. Configure a project with a recommender (string matching is sufficient, do not enable "always active")
2. Create a few annotations such that the recommender generates suggestions
3. Restart application
4. Open the document again
5. See that there are annotations

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
@@ -173,7 +173,7 @@ public class SelectionTask
             recommendationService.setActiveRecommenders(user, layer, activeRecommenders);
         }
 
-        schedulingService.enqueue(new PredictionTask(user, getProject()));
+        schedulingService.enqueue(new TrainingTask(user, getProject()));
     }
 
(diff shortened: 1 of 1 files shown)


Rentier:
I do not exactly know why that fixes it. I assume that the prediction that is triggered by the selection task is too fast so that no training is done before, therefore not marking the contexts ready.

reckart:
The training/prediction happen internally in the selection task. Would it help if the selection task used a private internal context instead of accessing the "normal" one?

Rentier:
The selectiontask should not need a context, as we never save the model for evaluation.

reckart:
Why is the prediction triggered by the selection task?

Rentier:
I do not know. Although I think that it is reasonable that the selection starts the training, so that we know that enabling/disabling is fresh when retraining.

reckart:
The training should be triggered when
* the user opens the document (initialization)
* the user adds/updates/deletes an annotation (data changes)

The prediction should be triggered when
* the training is complete

The evaluation should be triggered when
* every n-th time (including the first time) the training is triggered a selection needs to be performed first

Issue #620, opened 2018-11-05
Exclude PDFAnno UI


wpmp:
**Is your feature request related to a problem? Please describe.**
PDFAnno needs to be adapted to the Inception UI in Inception PDF editor.

**Describe the solution you'd like**
Elements in the PDFAnno index.html that are not required are removed or hidden.
This means that only the PDF.js panel should be visible within Inception PDF editor.
Also buttons within PDF.js panel are removed which provide file operation functionality, such as `Open file`, etc.

PR #882, opened 2019-01-17
#620 - Exclude PDFAnno UI

wpmp:
**What's in the PR**
* add ui.pdf.debug config property
* PDF editor now shows only the PDF div if debug is set to false, else it shows complete PDFAnno UI

**How to test manually**
1. Set ui.pdf.debug to true or false (if not set it is false)
2. Open a PDF document in PDF editor
3. PDFAnno UI should be set when ui.pdf.debug=true, else only PDF div should be visible

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Looks good. In a follow-up issue we should also remove some of the pdf.js buttons, e.g. the button to upload a PDF, to rotate, the search also doesn't work for me, etc.

Issue #703, opened 2018-11-23
Link annotations to images via URL


reckart:
**Is your feature request related to a problem? Please describe.**
When annotating scanned/OCRed text, it would be quite useful if the original scan could be shown on screen.

**Describe the solution you'd like**
- [x] a new *image feature* type where the feature value is a URL that points to an image
- [x] a new *image sidebar* which goes through the annotations currently visible on the screen and if any of them has an image feature, it loads that image and displays it. Clicking on an image should focus the annotation(s) that reference it

**Additional context**
Note, this is not about creating annotations on images, but rather about linking annotations to images entire images represented by a URL, not areas within an image. The URL might encodes an image region via IIIF, but to INCEpTION it's just an image.

mromanello:
Hi @reckart 

I've implemented a small Uima/XMI exporter for our data, and I could produce soon some data with this new `image_link` layer for you to test it (as soon as the feature is implemented in INCEpTION, of course).

I've created and export an "Images" layer that include an `image_link` feature (see snippet below). Do you think that could work? For our data I'd put in there IIIF links to the image region the annotated text is about.  

```json
[ {
  "name" : "webanno.custom.Images",
  "features" : [ {
    "name" : "image_link",
    "tag_set" : null,
    "uiName" : "image_link",
    "type" : "uima.cas.String",
    "enabled" : true,
    "visible" : true,
    "include_in_hover" : false,
    "required" : false,
    "remember" : false,
    "hideUnconstraintFeature" : false,
    "description" : "URL of the image to which the annotated text span is related.",
    "project_name" : "impresso",
    "multi_value_mode" : "NONE",
    "link_mode" : "NONE",
    "link_type_name" : null,
    "link_type_role_feature_name" : null,
    "link_type_target_feature_name" : null,
    "traits" : null
  } ],
  "uiName" : "Images",
  "type" : "span",
  "description" : null,
  "enabled" : true,
  "built_in" : false,
  "readonly" : false,
  "attach_type" : null,
  "attach_feature" : null,
  "allow_stacking" : false,
  "cross_sentence" : true,
  "show_hover" : true,
  "anchoring_mode" : "SENTENCES",
  "lock_to_token_offset" : false,
  "multiple_tokens" : false,
  "project_name" : "impresso",
  "linked_list_behavior" : false,
  "on_click_javascript_action" : null
} ]
```

reckart:
Ok, here you go with a first prototype :)

[...]

PR #883, opened 2019-01-17
#703 - Link annotations to images via URL

reckart:
**What's in the PR**
- Add new feature type for image URLs
- Add sidebar which shows all the images refered to by image features on the current page

**How to test manually**
* Create a new span layer
* On the layer, create a feature of type "Image URL"
* Go to the annotation page
* Open the images sidebar
* Create an annotation and set the feature value to the URL of an image (e.g. a PNG, JPG, etc.)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
This feature should be working ok enough not to require a switch to disable it - but I expect there will be improvements in the future.

Issue #910, opened 2019-02-01
Whitespace can confuse StringMatchingRecommender


reckart:
**Describe the bug**
If the StringMatchingRecommender learns that `John Smith` is a person, it won't predict that `John\tSmith` is a person as well.

**To Reproduce**
1. Import `John Smith John\tSmith` and test.

**Expected behavior**
The kind of whitespace which separates tokens should not confuse the recommender.

**Please complete the following information:**
 - Version and build ID: 4a9f918d75a0ac64af72dbc61b0398c61860f8e2

Rentier:
Dont we normalize whitespace somewhere?

reckart:
No at the level of the document text that is in the CAS. The CAS contains whatever whitespace the original document contained (and which passed through the DKPro Core Reader). We do a bit of normalization e.g. when sending text to brat.

reckart:
Note, for other recommenders, this shouldn't be a problem because they operate on Tokens and Tokens normally don't include whitespace. But the StringMatchingRecommender operates directly on the CAS document text.

Rentier:
What happens if we would introduce the sentence level recommender in #590 ?

PR #911, opened 2019-02-01
#910 - Whitespace can confuse StringMatchingRecommender

reckart:
**What's in the PR**
- Allow the Trie to internally sanitize keys
- Added WhitespaceNormalizingSanitizer
- Added unit tests for Trie with and without sanitizing

**How to test manually**
* See issue description

**Automatic testing**
* [x] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/Trie.java
+++ b/inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/Trie.java
@@ -83,7 +83,7 @@ public class Trie<V>
     }
 
     /**
-     * @param key
+     * @param aKey
      *            the key.
      * @param value
      *            the value.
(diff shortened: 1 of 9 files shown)


reckart:
@inception-project/inception-reviewers Review anybody?

inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/Trie.java
```diff
@@ -379,6 +419,30 @@ private void keys(final Character c, final Node n, final StringBuilder b,
         }
     }
 
+    public static interface KeySanitizerFactory
```
Rentier: I do not understand why you need a factory here. Also, I would move this static class to its own file. For me, nested classes/interfaces are only used inside a file, but you implement it outside.

reckart: I need a factory because the actual sanitizer is stateful. It is a character stream processor which needs to have a memory about the last character(s) it has seen. Thus, I cannot re-use a sanitizer instance.

PR #925, opened 2019-02-05
#277 Import/export external search settings

### LINKED ISSUE #277: Import/export external search settings when importing/exporting project

az79nefy:
**What's in the PR**
* import and export functionality for a project covers external document repositories
* resolves #277 

**How to test manually**
* create a project with external document repositories
* export project and re-import into new project
* check whether external document repositories in new project match the original project

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


inception-external-search-core/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/exporter/DocumentRepositoryExporter.java
```diff
@@ -0,0 +1,109 @@
...
+    private final ExternalSearchService externalSearchService;
+    
+    @Autowired
+    public DocumentRepositoryExporter(ExternalSearchService aExternalSearchService)
+    {
+        externalSearchService = aExternalSearchService;
+    }
+    
+    @Override
+    public List<Class<? extends ProjectExporter>> getImportDependencies()
```
reckart: I don't think this imported/exporter depends on the LayerExporter.

az79nefy: Does this exporter have any dependencies? I could just use the default ``getImportDependencies`` and ``getExportDependencies`` methods provided by the ProjectExporter interface and return an empty list.

az79nefy:
Encountering weird bug: When importing back an exported document, the document repositories in the imported project are correct. But the project that was exported deletes its document repositories, once the same project is imported again.

az79nefy:
Solved bug:

Assigning document repositories an existing ID (even across projects) leads to the deletion of the former document repository.
@reckart I solved this with random number generation for new ID, but this is not fail safe. How do we guarantee unique IDs?

reckart:
When creating/importing a document repository, the ID must be null - the database automatically assigns an ID. If you check the other importers, you'll note that none of them sets the ID field on import and most don't even save the ID field on export.

Issue #924, opened 2019-02-05
Improve debugging of scheduler


Rentier:
**Is your feature request related to a problem? Please describe.**
It is difficult to reason about why tasks are scheduled. This makes it hard to debug.

**Describe the solution you'd like**
Add a "source" or "origin" flag to `SchedulingService::enqueue"

PR #927, opened 2019-02-05
#924 - Improve debugging of scheduler

Rentier:
**What's in the PR**
- Add 'origin' to enqueue

**How to test manually**
- Set logging to debug
- Annotate stuff
- See in the log that tasks are started for reasons

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
@@ -173,7 +173,8 @@ public class SelectionTask
             recommendationService.setActiveRecommenders(user, layer, activeRecommenders);
         }
 
-        schedulingService.enqueue(new TrainingTask(user, getProject()));
+        schedulingService.enqueue(new TrainingTask(user, getProject()),
+                "Started by SelectionTask after activating recommenders");
     }
 
     private List<CAS> readCasses(Project aProject, String aUserName)
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
@@ -64,9 +64,9 @@ public class SelectionTask
     private @Autowired ApplicationEventPublisher appEventPublisher;
     private @Autowired SchedulingService schedulingService;
 
-    public SelectionTask(Project aProject, User aUser)
+    public SelectionTask(Project aProject, User aUser, String aTrigger)
     {
-        super(aUser, aProject);
+        super(aUser, aProject, aTrigger);
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
```diff
@@ -173,7 +173,8 @@ public void run()
             recommendationService.setActiveRecommenders(user, layer, activeRecommenders);
         }
 
-        schedulingService.enqueue(new TrainingTask(user, getProject()));
+        schedulingService.enqueue(new TrainingTask(user, getProject()),
+                "Started by SelectionTask after activating recommenders");
```
reckart: "Started by" is redundant with the "Triggered by" prefix that is always added when logging.

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
```diff
@@ -85,13 +85,15 @@ private void afterExecute(Runnable aRunnable, Throwable aThrowable)
         return result;
     }
 
-    public void enqueue(Task aTask)
+    public synchronized void enqueue(Task aTask, String aOrigin)
```
reckart: `aOrigin` -> `aReason`?

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
```diff
@@ -85,13 +85,15 @@ private void afterExecute(Runnable aRunnable, Throwable aThrowable)
...
     {
         if (getScheduledTasks().contains(aTask)) {
             log.debug("Task already in queue: {}", aTask);
             return;
         }
 
+        log.debug("Enqueuing task [{}] - Source: [{}]", aTask, aOrigin);
```
reckart: I think it would be good if the task would remember why it was triggered so we can even log this info when the task completes (asynchronously).

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
Rentier: I do not like reason too much, as it implies causality

reckart: I think causality is given. We wouldn't start a task if there wasn't a reason for it?

Rentier: I understand that your wording might stem from Exceptions

reckart: If you don't like reason, how about `aTrigger`? ;)

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
Rentier: Done

inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/SchedulingService.java
Rentier: Done

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/SelectionTask.java
Rentier: Done

Issue #885, opened 2019-01-17
Clean PDF.js UI


wpmp:
**Is your feature request related to a problem? Please describe.**
In the PDF editor the UI of PDF.js is shown.
There are some buttons that are not working in the PDF editor or should not be used.

**Describe the solution you'd like**
Buttons that are unusable or not required should be removed

wpmp:
@reckart which elements should be hidden?
I'd say we should hide `search`, `presentation mode`, `open file`, `current view`, `rotate clockwise` and `rotate counterclockwise`.
All of those either do not work, cause misbehaviour in INCEpTION / PDFAnno or have no use in INCEpTION.
What do you think?

PR #948, opened 2019-02-13
#885 - Clean PDF.js UI

wpmp:
**What's in the PR**
* remove search, presentation mode, open file, current view, rotate clockwise and rotate counterclockwise buttons from PDF.js UI

**How to test manually**
* Open a PDF document with the PDF editor
* The removed buttons should not be visible anymore

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I would also suggest to hide "print" and "download" - because the user might expect for both that the print/download includes the annotations, but they won't.

wpmp:
Good point

Issue #939, opened 2019-02-11
PDF editor loads wrong PDF


wpmp:
**Describe the bug**
When opening a PDF file it is possible that a wrong PDF file is loaded instead.
This happens when the PDF file is loaded over an URL that is cached by the browser.

**To Reproduce**
1. Start INCEpTION
2. Open a PDF file
3. Restart INCEpTION
4. Open some other PDF file
5. Wrong PDF file is loaded

**Expected behavior**
The PDF file that is selected should be loaded and opened.

PR #952, opened 2019-02-14
#939 - PDF editor loads wrong PDF

wpmp:
**What's in the PR**
* remove caching on GET request for PDF file and pdftxt which fixes loading wrong PDF files

**How to test manually**
* open a PDF file
* restart INCEpTION
* open another PDF file

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I tried it in Safari, but didn't seem to work for me.

wpmp:
@reckart well, that's unfortunate.
I could maybe add a random query parameter to the GET request, like the current time.
That should work, what do you think?

reckart:
Adding a random query parameter should work.

wpmp:
@reckart I changed it, hope it also works for safari now

Issue #947, opened 2019-02-13
Removing recommendations in PDF editor


wpmp:
**Is your feature request related to a problem? Please describe.**
In brat editor recommendations are removed when double clicking them.
This behavior should also added to the PDF editor.

**Describe the solution you'd like**
On double clicking a recommendation it gets removed.
Single clicking the recommendation still creates an actual annotation.

PR #951, opened 2019-02-14
#880 - Selection of relations in PDF editor

wpmp:
**What's in the PR**
* support for selecting relation annotations in PDF editor
* fixes tests for relation selection
* adding a missing extension marker in PDFAnno source

**How to test manually**
1. open a PDF file in PDF editor
2. create two spans
3. create a relation between those spans by drag and drop
4. select this relation

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**
* [ ] PR updates documentation

PR #960, opened 2019-02-19
#947 - Removing recommendations in PDF editor

wpmp:
**What's in the PR**
* support for removing recommendations by double clicking span annotation knob

**How to test manually**
1. add a recommender
2. open a PDF document in PDF editor
3. double click the annotation knob of a recommendation

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


wpmp:
This PR is based on #951

reckart:
@wpmp is this really WIP or is it already good to review?

wpmp:
@reckart it is ready for review, just kept WIP because of the merging and branches stuff

wpmp:
@reckart so, recommendations have its own color, right?
If that's true I could identify them in PDFAnno by its color.
Maybe I could change PDFAnno to render the dot to be of a different color, so all "normal" annotations would be blue and the recommendations would be like grey or so.
What do you think?

reckart:
Yes, they have a different color - but that is not the point.

Let's say you have two suggestions for a given span, e.g. it could be a "LOC" or a "PER" - so there will be two blue dots. I cannot tell which one I'm accepting/rejecting before actually doing it.

wpmp:
Ok, now I understand.
I guess I get the type of a suggestion by its layer?

reckart:
See `de.tudarmstadt.ukp.clarin.webanno.api.annotation.util.TypeUtil.getUiLabelText(TypeAdapter, Map<String, String>)` and its usages.

wpmp:
I created an issue #981

Issue #940, opened 2019-02-11
External search sidebar does not work properly if annotation interface accessed via top-level external search


rgabbard:
**Describe the bug**
If the user first accesses the "Annotate" top-level menu item and then opens the the external search sidebar (introduced in https://github.com/inception-project/inception/pull/846), it works correctly.  However, if the user first accesses the "Search" top-level item and then shifts to the annotation view by opening one of the results, then whatever document is initially loaded will be "stuck" in the interface, even if the user attempts to switch to another document via the external search interface.

**To Reproduce**
1. Open a project configured for external search.
2. Click "Search" on the left
3. Execute a search for a keyword sure to produce some results.
4. Click "Import" for the first result.
5. Click "Open" for the first result.
6. Open the external search sidebar by clicking the globe with an arrow.  
7. Do another search.
8. Click the first result.

Observe the document open in the annotation pane does not change.

**Expected behavior**

The document open in the annotation pane does not change to the selected document.

**Please complete the following information:**
 - Version and build ID:  INCEpTION -- 0.8.0-SNAPSHOT (, build ), https://github.com/inception-project/inception/pull/846 branch
 - OS: OS X
 - Browser: Chrome

rgabbard:
(unfortunately, we at ISI probably won't be able to pursue this one, since it isn't necessary to fix it to accomplish the goal we are funded for)

reckart:
I guess the problem are the duplicate URI parameters that appear when opening a document via the external search page: 

```
http://localhost:8080/inception-app-webapp/annotation.html?12&p=94&d=1338#!p=94&d=1338&f=1
```

PR #965, opened 2019-02-21
#940 - External search sidebar does not work properly if annotation interface accessed via top-level external search

reckart:
**What's in the PR**
- Use fragment parameters instead of page parameters when opening a document from the external search page to avoid page parameters overriding fragment parameters later on

**How to test manually**
* See issue description

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@az79nefy Maybe you could have a look at it?

Issue #968, opened 2019-02-21
Random score option in external search


shinyichen:
**The Problem**
The current external search (ElasticSeach) returns results in order of the best match. We would like to have the option to use the consistent random ordering when searching.

**Solution**
Have the random order option when configuring a new document repository. Use the [`random_score`](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html#function-random) in ElasticSearch query. 

**Alternatives**
Have the random order checkbox in the search form.

reckart:
With the fixed random seed currently used, all users will get the same results for the same query. If you wanted to have different users get different results (but still consistently across sessions/searches), it could be considered to use a hash of the user ID as the seed. I don't know what your specific motivation is for randomization.

PR #969, opened 2019-02-22
Random score external search

shinyichen:
**What's in the PR**
* PR for feature request #968 
* Added random order option to the elastic search config form
* Updated query to use random_score if random order is true

**How to test manually**
We can test this by going to the project configuration for elastic search and check the `random ordering` box. Then do search in External Search. The results should be in different order than ordered search. Do the same search term again and see that the results from random ordering is always consistent.


reckart:
@shinyichen is it ready for review?

shinyichen:
Yes @reckart. Adding @rgabbard to the loop.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
```diff
@@ -92,15 +94,24 @@ public boolean isConnected()
...
         
         // Set headers
         HttpHeaders headers = new HttpHeaders();
         headers.setContentType(MediaType.APPLICATION_JSON);
 
         
         // Set query
-        String query = "{\"size\":%d,\"query\":{\"match\":"
+        String query;
+        if (randomOrder) {
```
rgabbard: Can we build this up in a structured way and convert it to JSON? Right now there is significant repetition of the JSON request content between the two branches.

rgabbard: Also, a comment to document what is going on here would be nice.  You might want to give a name like `ARBITRARY_FIXED_SEED` to `5` to make clear there is nothing magic about it.

reckart: Maybe the seed should be a trait property too?

rgabbard: It could be - it looks like it is easy to add properties, so let's go ahead and do that @shinyichen

shinyichen: Would you let use edit the seed in the config?

rgabbard: There's probably little harm in it, except perhaps confusion.  Is it easy to provide some sort of help string for a config setting?

shinyichen: Tooltip?

shinyichen: I can look up how it's done.

shinyichen:
Ready for another round. Seed is configurable (mouse over the text field to see the tooltip). Query is build from a Json builder.

inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
```diff
@@ -92,100 +105,139 @@ public boolean isConnected()
...
+        ObjectNode emptyNode = mapper.createObjectNode();
+        highlightNode.putPOJO("fields", mapper.createObjectNode()
+            .putPOJO("doc.text", emptyNode));
+        bodyNode.putPOJO("highlight", highlightNode);
 
         // Set body
-        String body = String.format(query, resultSize, aQuery);
+        String body = null;
+        try {
+            body = mapper.writeValueAsString(bodyNode);
```
rgabbard: Why does the `try` block need to continue past this line in order to catch JSON errors?

shinyichen: Fixed.

rgabbard:
@reckart : This is ready for your review.

reckart:
I made some improvements to the UI code and code structure. 

@rgabbard @shinyichen when random results are turned on, the search page still shows a score. However, a user paying attention might find that the score doesn't relate too much to the actual quality of the result because the score is random (right?). Do you think the the score should not be added to the `ExternalSearchResult` if the ordering is random?

reckart:
Is there a particular reason that the random seed is limited to the range 0-1000?

reckart:
See also: https://github.com/inception-project/inception/issues/987

rgabbard:
@shinyichen : can you expand the range of the seed to whatever ElasticSearch accepts? (probably a `long`?)

I agree we shouldn't show the score when the ordering is random.  Can we shift suppressing the scores to a separate issue? (we are anxious to get this merged because we have annotators starting soon)

reckart:
@reckart I have made the changes to skip the score in random ordering mode. So what's left is the question about the random seed. Otherwise, it's good to merge from my perspective.

Issue #958, opened 2019-02-18
KB instance not properly filtered when annotating


mromanello:
**Describe the bug**

I have a KB connected to an annotation layer. For a given feature (`work_uri`) I specify its scope, in this case it can take as a value only instances of class `work`.
![image](https://user-images.githubusercontent.com/310693/52950555-ce81ff80-337f-11e9-83f9-1a5f909a9c8b.png)

Nevertheless, when annotating the dropdown menu for this feature is populated with strings attached to other instance types (e.g. `title`). 

**Expected behavior**

The dropdown for `work_uri` feature should contain only strings that in the KB are associated to instances of `work`.

**Please complete the following information:**
 - Version and build ID:0.7.2 (2019-01-30 17:46:49, build b490d4b8790c1c97df107762770f52dcd1b4fb8e)

----

- [x] Fix bug for the annotation page
- [x] Make sure the search field on the KB page also works well
- [x] Update documentation (developer and user documentation)

reckart:
I'm afraid that filtering and full-text search are (currently) mutually exclusive. The filtering should work if you disable the full-text search capability in the KB settings. Thus, it should work ok for comparatively small KBs but won't work well for large KBs.

mromanello:
Ok thanks. Actually, I could live with this behaviour ;-) the problem is that, if I disable the full-text search, it doesn't find any values at all. I wonder if it's normal? Or maybe I need to change some other settings?

![screen shot 2019-02-18 at 15 21 00](https://user-images.githubusercontent.com/310693/52956993-f332a300-3390-11e9-884e-873d439b44be.png)

reckart:
https://github.com/inception-project/inception/pull/912 might fix that problem.

reckart:
@mromanello could you test with the changes from #912 or tell me how to configure the KB so I can test with your settings?

mromanello:
> @mromanello could you test with the changes from #912 or tell me how to configure the KB so I can test with your settings?

hi @reckart sorry, I won't be able to test this soon-ish, but here are my KB settings:

![image](https://user-images.githubusercontent.com/310693/53181998-e7491a00-35f8-11e9-8077-6648e6ef846a.png)
![image](https://user-images.githubusercontent.com/310693/53182006-ec0dce00-35f8-11e9-8032-b783ff349699.png)

just ping me if you need more info.

reckart:
[...]

PR #978, opened 2019-02-26
#958 - kb instance not properly filtered when annotating

reckart:
**What's in the PR**

- Factor auto-complete field for KB items into a separate re-usable class
- Use this class in the ConceptFeatureEditor and on the KB page
- Fix the search on the KB page mainly by properly handling cases where withLabelMatchingExacltlyAnyOf or withLabelContainingAnyOf are called with an emtpy argument list
- Updated documentation (user + developer)

**How to test manually**
* Go to the KB page and search for some class or instance

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


MdeBoer95:
Works pretty well. Just a few smaller things.
- I found one strange case:
auswahl_008

- The autocomplete field shows entities, where at least 4 characters match.
So when I want to search for ```wine``` I have to type the whole word.
Is that the intended behaviour?

- I would suggest to show a badge for each item in the search results that indicates the type(Concept, Instance, Property). That makes it easier to find something on the first try if for example a concept and a property have the same name but with different casing. But that should be something for a separate Issue.
auswahl_009

reckart:
> "Region" -> "wine"
That looks indeed strange. I can reproduce it. I also note that when clicking on the result (or any also on the results I get when searching for "wine" in the first place) the respective item doesn't really load in the class/instance detail editor. I'd be inclined to open a separate issue/bug for this. WDYT?

> The autocomplete field shows entities, where at least 4 characters match.
  So when I want to search for `wine` I have to type the whole word.
  Is that the intended behaviour?

Yes. Virtuoso won't allow to do a wildcard search for words with less than four characters. So you can search for `wine` and `wine*` but not for `win*`. To be consistent across backends and because searches for short prefixes tend to yield many results and thus become slow, no wildcard search is attempted if the entered query is four characters or less. However, even for shorter queries, an exact-match query is performed, so if you enter `X` and there is an entry labelled `X`, you should find it.

> I would suggest to show a badge for each item in the search results that indicates the type(Concept, Instance, Property).

That is a good idea. The question is how much it slows down the process if we do this. But this needs to be found out. Could you please open a new issue for your idea?

MdeBoer95:
Yes, I will open corresponding issues for the bug and the visual enhancement.

MdeBoer95:
> Yes, I will open corresponding issues for the bug and the visual enhancement.

Created Issues #979 and #980.

reckart:
Thx

Issue #993, opened 2019-02-28
Display a preview of images in image feature editor


reckart:
**Is your feature request related to a problem? Please describe.**
The image feature editor currently only shows the image URL. 

**Describe the solution you'd like**
It would be nice to be able to see an actual preview of the image (thumbnail).

PR #996, opened 2019-02-28
#696 - Support document level annotations

reckart:
**What's in the PR**
- Added button to close annotation detail editor
- Properly reload the annotation list and update the labels when changes are made to annotations
- Added very basic documentation sections in user and developer documentation
- Removed hard-coded HTML ID which isn't really used and bad style anyway
- Add scroll bar to the annotation detail area
- Fixed exception when editing a concept feature at the document level
- Fixed exception when document level annotation has an image feature and image sidebar is opened

**How to test manually**
* Add `documentmetadata.enabled=true` to the `settings.properties` files
* Create 1-2 document-level annotation layers with different types of features
* Create annotations on these layers
* Access potentially related areas of the application (image sidebar, knowledge-base page, etc.)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


inception-layer-docmetadata/src/main/java/de/tudarmstadt/ukp/inception/ui/core/docanno/sidebar/DocumentMetadataAnnotationSelectionPanel.java
```diff
@@ -97,7 +101,10 @@ public DocumentMetadataAnnotationSelectionPanel(String aId, IModel<Project> aPro
         detailPanel = aDetails;
         selectedLayer = Model.of();
 
-        add(createAnnotationList());
+        annotationsContainer = new WebMarkupContainer("annotationsContainer");
```
jcklie: Dont we normally have constants for ids?

inception-layer-docmetadata/src/main/resources/META-INF/user-guide/annotation_create-annotations_docmeta.adoc
```diff
@@ -0,0 +1,6 @@
+= Document metadata
+
+To manage document-level annotations, access the **Document metadata** sidebar on the annotation
```
jcklie: I would like some more documentation before we remove the switch, especially some nice screenshots

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
```diff
@@ -103,10 +104,16 @@ public void renderHead(IHeaderResponse aResponse)
                 return Collections.emptyList();
             }
 
+            JCas jcas = aHandler != null ? aHandler.getEditorCas() : null;
```
jcklie: I have no idea what that is doing.

reckart:
> Does it make sense to have the behaviours there when selecting a layer of type documenttype?

Some probably. Anyway, there is already an issue for this: #1001 - Hide irrelevant layer settings for document-level annotations 

> When I have an integer feature, increment it once and click close, then it is not saved. Is it intended? I need to change another feature or click somewhere unrelated to make it save.

Theoretically, the values are saved on-focus-lost. I guess if you click *Close* the close action takes precedence over the on-focus-lost AJAX action. So it is a bug, but should be tracked in a separate issue. We probably have the same problem in the `AnnotationDetailEditorPanel` when the user clicks on *Clear*. Could you please open a separate issue for this?

> It would be useful to indicate what type a feature is, especially for image URL vs string feature
Image Urls are not clickable and do not show an image

See #993 - Display a preview of images in image feature editor (PR: #997)

> Required is not validated for me

See #1000 - Mark required document-level annotations if required features are not set

For me, these are all valid problems but not blockers for this PR. There are a few additional things that need to be done before the support is really "complete": https://github.com/inception-project/inception/projects/38

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/feature/ConceptFeatureEditor.java
reckart: Adding a JavaDoc comment.

inception-layer-docmetadata/src/main/java/de/tudarmstadt/ukp/inception/ui/core/docanno/sidebar/DocumentMetadataAnnotationSelectionPanel.java
reckart: Introduced them.

PR #1018, opened 2019-03-08
Feature 919/Add button to leave a project

UWinch:
**What's in the PR**
Add a button to the project overview page allowing to leave a project. Ask before actually leaving since re-joining won't be possible without the project manager's involvement.
Leaving the project is not possible for managers/admins to avoid orphaned projects.

**How to test manually**
* In a role that is not the project's manager/admin: Go to projects overview page and press leave Project button
* Should open a window to confirm your choice and then hide this project from you

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.java
```diff
@@ -202,6 +208,51 @@ protected void onConfigure()
...
+                new StringResourceModel("LeaveDialog.text", this));
+        confirmLeaveDialog.setConfirmAction((target) -> {
+            projectService.listProjectPermissionLevel(user, currentProject).stream()
+                    .forEach(projectService::removeProjectPermission);
+            setResponsePage(getPage());
+        });
+        return confirmLeaveDialog;
+    }
+
+    private AjaxLink<Void> createLeaveProjectButton(User user, Project currentProject,
```
Sibgha360: Just a minor thing. In INCepTION, arguments to constructors and methods are prefixed with a, e.g. aVariable. The a stands for argument and is used to differentiate between member variables and arguments. Hence, the member should be called variable.
In Eclipse you can configure this default prefix "a" under Preferences -> Java -> Code Style set the "prefix list" column in the "parameters" row to "a".

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.properties
```diff
@@ -19,3 +19,6 @@ page.help=Help
 page.help.link=doc/user-guide.html#sect_menu
 
 newProject=Create new project ...
+
+LeaveDialog.title=Leave Project
```
Sibgha360: Naming convention is recommended as lowercase in property file. leaveDialog.title in this case is more appreciable.

Sibgha360:
Following scenario is only indirectly related to this pull request. While testing I encountered the following.

I am a project creator user, I create a project and assign it to admin. While assigning, I do not give the admin manager permission (which may or may not make sense but the interface allows me). Then I log in through the admin user, the newly created project shows up with the `Leave Project` button. When I click leave, it does not do anything. In general, it causes confusion with the `Leave Project` button.

UWinch:
> Following scenario is only indirectly related to this pull request. While testing I encountered the following.
> 
> I am a project creator user, I create a project and assign it to admin. While assigning, I do not give the admin manager permission (which may or may not make sense but the interface allows me). Then I log in through the admin user, the newly created project shows up with the `Leave Project` button. When I click leave, it does not do anything. In general, it causes confusion with the `Leave Project` button.

Thanks @Sibgha360 for the review and the finding. I hopefully now fixed it by no longer showing the button to people with no rights for the project.

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.html
```diff
@@ -56,6 +56,10 @@ <h4 class="list-group-item-heading">
             <div class="text-right">
               <i wicket:enclosure="created">Created: <wicket:container wicket:id="created"/></i>
             </div>
+            <div>
+          	  <input wicket:id="leaveProject" type="button" class="btn btn-primary" wicket:message="value:leaveDialog.title"/> 
+        	</div>
+        	<div wicket:id="confirmLeave"></div>
```
reckart: Does the confirmation dialog have to reside in the repeater? Wouldn't it be better to have it once outside of the list?

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/projectlist/ProjectsOverviewPage.java
```diff
@@ -202,6 +208,53 @@ protected void onConfigure()
...
+                    .forEach(projectService::removeProjectPermission);
+            setResponsePage(getPage());
+        });
+        return confirmLeaveDialog;
+    }
+
+    private AjaxLink<Void> createLeaveProjectButton(User aUser, Project aCurrentProject,
+            ConfirmationDialog confirmLeaveDialog)
+    {
+        AjaxLink<Void> leaveProjectLink = new AjaxLink<Void>(MID_LEAVE_PROJECT)
```
reckart: How about something like this:

```
return new LambdaAjaxLink(MID_LEAVE_PROEJCT, 
  _target -> actionConfirmLeavingProject(_target, listitem))
  .add(LambdaBehavior.visibleWhen(() -> your condition);

private void actionConfirmLeavingProject(AjaxRequestTarget aTarget, ListItem<Project> aItem) {
   reconfigure dialog for the given project
   show dialog
}
```

jcklie:
@UWinch @reckart How can we wrap up this PR?

UWinch:
The functionality is implemented as a dropdown now, maybe you can comment on the layout?

Issue #1027, opened 2019-03-10
Remove unused read() methods from KBObject subclasses


reckart:
**Describe the refactoring action**
Remove unused read() methods from KBObject subclasses.

**Expected benefit**
Remove cruft.

PR #1028, opened 2019-03-10
#1027 - Remove unused read() methods from KBObject subclasses

reckart:
**What's in the PR**
- Removing unused code
- Inline a method in RdfUtils which was only used internally in RdfUtils after the code has been removed

**How to test manually**
* There is no need to test anything. If it still compiles, it should be good.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/graph/RdfUtils.java
+++ b/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/graph/RdfUtils.java
@@ -43,10 +43,10 @@ import de.tudarmstadt.ukp.inception.kb.model.KnowledgeBase;
 public class RdfUtils
 {
     public static Optional<Statement> readFirst(RepositoryConnection conn, Resource subj, IRI pred,
-            Value obj, String language, KnowledgeBase aKB)
+            Value obj, KnowledgeBase aKB)
     {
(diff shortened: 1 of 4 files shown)


inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/graph/RdfUtils.java
```diff
@@ -43,10 +43,10 @@
 public class RdfUtils
 {
     public static Optional<Statement> readFirst(RepositoryConnection conn, Resource subj, IRI pred,
-            Value obj, String language, KnowledgeBase aKB)
+            Value obj, KnowledgeBase aKB)
     {
-        try (RepositoryResult<Statement> results =
-            getStatementsSparql(conn, subj, pred, obj, false, language, aKB)) {
+        try (RepositoryResult<Statement> results = getStatementsSparql(conn, subj, pred, obj, false,
```
jcklie: Is the null dangerous here?

reckart: No - `SPARQLQueryStore.languageFilter(String, String)` (where this eventually arrives) allows the language to be null. Anyway, this code is hopefully going away soon in favor for `SPARQLQueryBuilder`.

Issue #906, opened 2019-01-31
SchedulingServiceTest fails sporadically


reckart:
**Describe the bug**
The order in which the tasks are executed by the test does not match the order in which they were enqueued.

**To Reproduce**
Doesn't always happen - no way to consistently reproduce. Has been observed on Windows, OS X and Linux (Jenkins).

**Expected behavior**
Test should execute consistently.

**Screenshots**
```
org.junit.ComparisonFailure: 
[All enqueued tasks should be running] expected:<...er1, project=project[1},
    DummyTask{user=user1, project=project2]},
    DummyTask{use...> but was:<...er1, project=project[2},
    DummyTask{user=user1, project=project1]},
    DummyTask{use...>
	at de.tudarmstadt.ukp.inception.scheduling.SchedulingServiceTest.thatRunningTasksCanBeRetrieved(SchedulingServiceTest.java:83)
```

**Please complete the following information:**
 - Version and build ID: 3672bc04ad23914eef346c6233e5b7fdcbd1a6fd
 - OS: Windows, Linux, OS X

**Additional context**
This [issue](https://bugs.openjdk.java.net/browse/JDK-8132696) report points out that the queue must be fair in order to ensure FIFO behavior.

reckart:
Happened again with 7a02642b6b9996c6a151ea678d1d70bab2b2a3eb built on Windows:

```
ava.lang.AssertionError: 
[All enqueued tasks should be running] 
Expecting:
  <[DummyTask{user=user1, project=project2, trigger=JUnit"}]>
to contain exactly in any order:
  <[DummyTask{user=user1, project=project1, trigger=JUnit"},
    DummyTask{user=user1, project=project2, trigger=JUnit"},
    DummyTask{user=user2, project=project1, trigger=JUnit"}]>
but could not find the following elements:
  <[DummyTask{user=user1, project=project1, trigger=JUnit"},
    DummyTask{user=user2, project=project1, trigger=JUnit"}]>

	at de.tudarmstadt.ukp.inception.scheduling.SchedulingServiceTest.thatRunningTasksCanBeRetrieved(SchedulingServiceTest.java:83)
```

PR #1031, opened 2019-03-12
#906 - SchedulingServiceTest fails sporadicall

jcklie:
**What's in the PR**
- Fix test by changing condition for waiting until test is ready for assertion

**How to test manually**
* Run the unit tests repeatedly
* Make a release and see that it does not crash due to this

**Automatic testing**
* [X] PR includes unit tests

**Documentation**


jcklie:
I checked, in total, there are always 4 Junit threads running so the waiting condition is wrong. I changed it and gave it more time.

reckart:
Looks good. Actually running a release won't be necessary. Let's Jenkins take care of building it (regularly) and watch if it fails again.

Issue #848, opened 2019-01-11
Support Jena/Fuseki FTS mode


reckart:
**Is your feature request related to a problem? Please describe.**
We could improve the interoperability with Jena/Fuseki by supporting their FTS approach: https://jena.apache.org/documentation/query/text-query.html

E.g. ZBW is hosting some Fuseki-based SPARQL endpoints - although I have no idea if they enabled FTS for these: http://zbw.eu/beta/sparql-lab/about/

**Describe the solution you'd like**
Support the Jena FTS as an additional FTS mode beside RDF4J and Virtuoso.

PR #1042, opened 2019-03-14
#848 - Support Jena/Fuseki FTS mode

reckart:
**What's in the PR**
- Added support for the Fuseki FTS implementation
- Updated ZBW SPARQL profiles to use Fuseki FTS
- Added tests running SPARQLQueryBuilder against ZBW endpoints using FTS
- Added class to help building RDF collections compatible with the SparqlBuilder from RDF4J

**How to test manually**
* Create a KB based on one of the ZBW profiles
* Check that FTS is enabled after KB creation
* Try linking entities and/or searching for KB items on the KB page

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


reckart:
I cannot really reproduce your problems. As you can see in the screenshots below, I can see entities and the dropdown for entity linking on the annotation screen works a well:





Irrespective of what's shown in the screenshots at the bottom as commit it, these were taken using the code from c1dc8c67e76ecd24df1e746a4b7bf2bc4e450c04.

jcklie:
Which KB did you use? I used the new ZBW STS, and it looks different from yours.

reckart:
See screenshot: I used the "Integrated Authority File (GND)" KB - there is a default profile for it.

reckart:
I had a look at the description of the [STW Thesaurus](http://zbw.eu/stw/version/latest/relaunch/about.en.html) - it seems to me that the thesaurus does not include any instances. They make a distinction between things that are   `http://zbw.eu/namespaces/zbw-extensions/Thsys` and things that are `http://zbw.eu/namespaces/zbw-extensions/Descriptor`. But since descriptors can have "subtypes", it doesn't make sense for us to treat them as instances. So... looks correct to me that there are no instances in the STW.

The ZBW GND endpoint also uses Fuseki and there we find instances.

Regarding entity linking: make sure that the KB feature's **Allowed values** setting includes concepts. The default seems to be to only include instances which does not work with the STW KB.

jcklie:
Do we document KBs for which we ship profiles?

reckart:
> Do we document KBs for which we ship profiles?

Not so far.

Issue #801, opened 2018-12-23
Handling of ligatures


wpmp:
**Describe the bug**
When PDF editor loads annotations from backend it does not receive annotations that contain ligatures.

**To Reproduce**
1. Create an annotation in brat editor which is a ligature in the PDF file
2. Change to PDF editor
3. The annotation is missing

**Expected behavior**
Annotations that contain ligatures should be correctly loaded in the PDF editor.

PR #1032, opened 2019-03-12
#1021 - PDF editor performance improvements

wpmp:
**What's in the PR**
* add Aho-Corasick algorithm for span annotation alignment
* adapt tests to work with new methods
* PdfExtractFile and DocumentModel now computed only once a PDF file is opened in PDF editor

**How to test manually**
* open PDF document in PDF editor
* check if annotations render correctly
* check if they are selectable
* check if new annotations can be created

**Automatic testing**
* [X] PR includes unit tests (modifies existing tests to use with new methods)

**Documentation**
* [ ] PR updates documentation

PR #1049, opened 2019-03-17
#801 - Handling of ligatures

wpmp:
**What's in the PR**
* handling of ligatures
* bugfix where quotation marks were not escaped when sending annotation to PDFAnno

**How to test manually**
* Open a PDF document that contians ligatures in PDF editor
* Make an annotation including a ligature

**Automatic testing**
* [X] PR includes unit tests

**Documentation**


wpmp:
this PR is based on #1032

reckart:
Is it ready for review?

wpmp:
Basically yes when #1032 is merged :)

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/PdfAnnotationEditor.java
```diff
@@ -338,9 +343,17 @@ public void initialize(AjaxRequestTarget aTarget)
         File pdfFile = documentService.getSourceDocumentFile(getModel().getObject().getDocument());
 
         try {
+            File file = new File(getClass().getClassLoader()
```
reckart: Fetching the file for a resource won't work when the application is packaged as a JAR - it works only when running in the IDE. You need to open the classpath resource as a stream and read it. I'd also suggest to load the substitution table directly from the same location the PdfReader uses such that we don't have to package the resource twice (potentially getting out of sync): `/de/tudarmstadt/ukp/dkpro/core/io/pdf/substitutionTable.xml`

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -42,22 +45,94 @@
     private String stringContent;
 
     /**
-     * Contains position mapping for a character between PDFExtract string
+     * Contains position mapping for characters between PDFExtract string
```
reckart: Is there an entry for *every character* in this map?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -42,22 +45,94 @@
...
      * including and excluding Draw Operations.
      */
     private BidiMap<Integer, Integer> stringPositionMap;
 
+    private Map<String, String> ligatures;
```
reckart: What is the purpose of this map?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -42,22 +45,94 @@
...
+
+    private String ligaturelessContent;
+
+    /**
+     * Contains position mapping for characters between ligaturelessContent and stringContent
+     */
+    private BidiMap<Integer, Integer> ligatureStringPositionMap;
```
reckart: Is there an entry for every character in this map?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -42,22 +45,94 @@
...
+            if (occurrences.containsKey(stringIndex)) {
+                // start of a ligature was found
+                Emit emit = occurrences.get(stringIndex);
+                String replacement = ligatures.get(emit.getKeyword());
+                int keyLen = emit.getKeyword().length();
+                int replacementLen = replacement.length();
+
+                // iterate over chars of ligature string and its replacement, create proper mapping
+                ligatureStringPositionMap.put(ligatureIndex, stringIndex);
+                sb.append(replacement.charAt(0));
```
reckart: Why do you only get the first character here? The substitution map contains some substitutions which go from 3 characters in length to 2 characters

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -42,22 +45,94 @@
...
+        }
+
+        int stringIndex = 0;
+        int ligatureIndex = 0;
+        StringBuilder sb = new StringBuilder();
+        // iterate over the text and create a new string containing replaced ligatures
+        // also create a new map that maps from new ligature string to normal string content
+        while (stringIndex < ligaturelessContent.length()) {
+            char c = ligaturelessContent.charAt(stringIndex);
+            if (occurrences.containsKey(stringIndex)) {
```
reckart: Wouldn't it be more efficient to iterate over the emits?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: Yes, the idea is to have a mapping of indices from the actual pdfextract file content with draw operations to indices from pdfextract file content without draw operations.
Then there is another map that maps indices from sanitized pdfextract file content without draw operations to the original pdfextract file content without draw operations.
This can be computed when a document is opened and while annotation we can just make lookups for the indices.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: It contains the mappings that were extracted from `substitutionTable.xml`

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: Yes, see the answer to line 45/48

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: As mentioned above I build a complete mapping between the original and sanitized text.
So I iterate over each character and the new replacements to get the indices right.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
reckart: I'd suggest using http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/Int2IntMap.html here - it is not bidirectional, but since fastutil doesn't work with Integer objects but with primitives internally, it is probably still more efficient to use two of the fastutil maps than one BidiMap with Integer objects.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: Remaining characters will be added later in the for loop.
I get the first one because I assume that we do have no empty replacements (e.g. "sample" to "") and it helps with indices and if coniditions in the loop.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/PdfAnnotationEditor.java
wpmp:
```
String substitutionTable = "/de/tudarmstadt/ukp/dkpro/core/io/pdf/substitutionTable.xml";
try (InputStream is = getClass().getClassLoader().getResourceAsStream(substitutionTable)) {
```
InputStream is null, how can I get this right?
Or do I need to have to place the file somewhere?
Thought it would be included with the dependency.
I am not very familiar with this stuff.

wpmp: I now used `ResourceUtils.resolveLocation` from dkpro, seems to work

wpmp: @reckart is it ok now?

wpmp:

wpmp:
@reckart ready for another review I think

reckart:

Issue #1051, opened 2019-03-18
Switch completely to Java-based Spring configuration


reckart:
**Describe the refactoring action**
INCEpTION is still using Spring XML files for parts of its configuration, e.g. for the Spring Security setup. This should be migrated to Java-based configuration.

- [x] application-context.xml
- [x] preAuthSecurity-context.xml
- [x] rest-context.xml
- [x] security-context.xml
- [x] static-resources-context.xml

**Expected benefit**
Better maintainability, in particular when upgrading at some point to Spring Boot 2 and Spring Security 5.

PR #1052, opened 2019-03-18
#1051 - Switch completely to Java-based Spring configuration

reckart:
**What's in the PR**
- Replace `security-context.xml` with an approach based on Spring's Java-config capabilities

**How to test manually**
* Log in as a valid user and check if it works
* Try logging in with a bad username / password
* Try accessing the remote API and ensure that it is only accessible by users with `ROLE_REMOTE` (enable `remote-api.enabled=true` in `settings.properties`, access API page via `.../swagger-ui.html`, actually try out invoking an API call - this should bring up a HTTP basic auth dialog in the browser - note that the browser caches the credentials, so try in a private window)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
> After accessing the swagger-ui.html page and invoking an API call, another authentification is required.
Entering the login details of a user without ROLE_REMOTE rights still invokes the API call.

The second login is desired - yes. However, if the user entered there does not have ROLE_REMOTE, then the access API call should be prevented. Thanks for testing this! It definitely needs to be fixed!

reckart:
> After accessing the swagger-ui.html page and invoking an API call, another authentification is required.
Entering the login details of a user without ROLE_REMOTE rights still invokes the API call.

So I tried this:

* I logged in as a user with `ROLE_REMOTE` and accessed the Swagger UI
* There I try out the **list projects** action and entered credentials there for a user who does **not** have `ROLE_REMOTE
* the message I got back form the API was a **403**:



So I cannot reproduce the effect that a user without `ROLE_REMOTE` can access the remote API. 

Tested with: 35af8e07fa182b6913fa759d80b3e3b797af9e8d

Issue #802, opened 2018-12-23
Rendering annotations pagewise


wpmp:
**Is your feature request related to a problem? Please describe.**
Currently, all annotations for a whole PDF document are loaded at once in the PDF editor.
To improve performance PDF editor should when a new page is rendered only load annotations of this page.

**Describe the solution you'd like**
Whenever a page is rendered in PDF editor it should send a request for annotation for the page by sending the pagenumber.

PR #1058, opened 2019-03-18
#905 - Offset for recommendation rendering

wpmp:
**What's in the PR**
* adapt changes from webanno/webanno#1212
* fix bug where span annotation texts contained ` and were not escaped

**How to test manually**
* Upstream changes from webanno/webanno#1212 required
* Open a PDF in PDF editor
* Trigger rendering for recommendations (e.g. by opening sidebar)
* See that the whole document is filled with recommendations

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**
* [ ] PR updates documentation

PR #1068, opened 2019-03-25
#802 - Rendering annotations pagewise

wpmp:
**What's in the PR**
* add pagewise annotation rendering

**How to test manually**
* open a PDF document in PDF editor
* scroll to pages and see rendering

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


wpmp:
PR is based on #1049 
Will also require #1058 for further testing

wpmp:
@reckart somehow I cannot change the labels to set it to `needs review`... any clue why this is the case?

reckart:
@wpmp I don't know. You have write permissions to the repo, so as far as I understand [the GitHub documentation](https://help.github.com/en/articles/repository-permission-levels-for-an-organization), you should be able to change issues including their labels.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
```diff
@@ -105,7 +105,8 @@ public void respond(AjaxRequestTarget aTarget)
             @Override
             protected void respond(AjaxRequestTarget aTarget)
             {
-                aPdfAnnotationEditor.renderPdfAnnoModel(aTarget);
+//                aPdfAnnotationEditor.renderPdfAnnoModel(aTarget);
```
reckart: WUT? Why is the above line commented out and a message logged to system out? If you want to log a warning about an unexpected API call, better log it through the logging system. Or throw an exception... or (since you are in a AJAX call) you could even log a feedback message to the UI.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/Offset.java
```diff
@@ -72,4 +72,9 @@ public String toString() {
             ", end=" + end +
             '}';
     }
+
+    public static Offset empty()
```
reckart: How about using singleton here?

```
static final Offset EMPTY = new Offset(...);

Offset empty() {
  return EMPTY;
}
```

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -199,6 +209,12 @@ private void initializeStringContent(String aPdftxt)
...
 
+            if (page > lastPage) {
+                pageOffsetMap.put(lastPage, new Offset(pageBeginIndex, extractLineIndex - 1));
+                lastPage = page;
+                pageBeginIndex = extractLineIndex;
+            }
+
             // if value of PdfExtractLine is in brackets it is a draw operation and is ignored
             if (!extractLine.getValue().matches("^\\[.*\\]$")
                 && !extractLine.getValue().equals("NO_UNICODE"))
```
reckart: Can you add a comment when `NO_UNICODE` appears?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
```diff
@@ -260,4 +279,19 @@ public Offset getStringIndex(int aExtractIndex)
             return new Offset(sanitizedIndex, sanitizedIndex);
         }
     }
+
+    /**
+     * Get gegin and end offset of a page.
+     * @param page
```
reckart: Either document parameter and return value with a proper description or remove the param/return stanzas.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/render/PdfAnnoRenderer.java
```diff
@@ -187,8 +187,7 @@ private static String getColor(VObject aVObject, ColoringStrategy aColoringStrat
     /**
      * for each span adds context with given windowsize before and after span.
      */
-    private static void addContextToSpans(List<RenderSpan> aSpans, int windowSize,
-                                          String text)
+    private static void addContextToSpans(List<RenderSpan> aSpans, int windowSize, String text)
```
reckart: `aWindowSize` + `aText`.

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
```diff
@@ -108,6 +108,24 @@ window.addEventListener('DOMContentLoaded', async () => {
...
+    "action": "getAnnotations",
+    "page": window.pageNumber
+  }
+  parent.Wicket.Ajax.ajax({
+    "m": "POST",
+    "ep": data,
+    "u": window.apiUrl,
+    "sh": [],
+    "fh": [function () {
+      console.log('Something went wrong on requesting annotations from inception backend.')
```
reckart: Wouldn't it be good to show this to the user, e.g. via a JS popup?

inception-pdf-editor/src/test/java/de/tudarmstadt/ukpinception/pdfeditor/pdfanno/model/PdfAnnoRendererTest.java
```diff
@@ -171,7 +171,7 @@ public void testRender() throws Exception
 
         PdfExtractFile pdfExtractFile = new PdfExtractFile(pdftxt, new HashMap<>());
         PdfAnnoModel annoFile = PdfAnnoRenderer.render(state, vdoc,
-            cas.getDocumentText(), schemaService, pdfExtractFile);
+            cas.getDocumentText(), schemaService, pdfExtractFile, Offset.empty());
```
reckart: If you pass `Offset.empty()` here, will the annoFile contain nothing?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/render/RenderSpan.java
```diff
@@ -44,13 +44,15 @@ public RenderSpan(Offset aOffset)
         end = aOffset.getEnd();
     }
 
-    public RenderSpan(VSpan aVSpan, Span aSpan)
+    public RenderSpan(VSpan aVSpan, Span aSpan, Offset offset)
```
reckart: Why pass an offset with begin/end if you only ever use the begin? I'd suggest to only pass the begin to make the code easier to understand and maintain.

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/PdfAnnoPanel.java
wpmp: I think I need more sleep.. this shouldn't have been commited..
This whole `AbstractDefaultAjaxBehavior` should be removed as all calls regarding annotations are now done over the `apiProvider`.

inception-pdf-editor/src/test/java/de/tudarmstadt/ukpinception/pdfeditor/pdfanno/model/PdfAnnoRendererTest.java
wpmp: No, this offset basically is the page offset.
E.g. if we render page 5 of a document this would contain the character offset to the beginning of the 5th page.
As in the test case we render the whole document we don't need an offset.
I probably should add tests for this with offset other than 0

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
```diff
@@ -189,6 +211,25 @@ async function displayViewer () {
     textLayer.setup(analyzeResult)
     window.annoPage.pdftxt = analyzeResult
 
+// BEGIN INCEpTION EXTENSION - #802 - Rendering annotations pagewise
+    const renderTimeout = 1500
```
wpmp: @reckart this is the timeout that PDFAnno will wait until it renders a new page if the page hasn't changed since last change.
Do you think 1500 ms is a good value?
I don't know if it is too short or too long.

inception-pdf-editor/src/test/java/de/tudarmstadt/ukpinception/pdfeditor/pdfanno/model/PdfAnnoRendererTest.java
reckart: If 0 is a valid offset, then maybe `Offset.empty()` should be `[-1,-1]` ? Or what is the purpose of this empty offset? I assume it is used to indicate a default invalid value.

reckart: Some JavaDoc would be good here ;)

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
reckart: > it renders a new page if the page hasn't changed since last change

I'm not sure I understand this. What does it mean if a page is "new" or has been "changed" and what kind of change is the "last change" - probably a different kind of change than the change to the page...?

inception-pdf-editor/src/test/java/de/tudarmstadt/ukpinception/pdfeditor/pdfanno/model/PdfAnnoRendererTest.java
wpmp: No it is fine as it is, it just indicates we do not have some offset for rendering here.
The offset begin value is added to each span.
If we would have like page 2, the offset would be the index of page 2 begin.
Here it is just at the begin of the document, hence offset 0.
Maybe `empty` is a misleading name.
But as you stated somewhere else in the review we don't need an offset here, just the begin of the offset is required.
I will make changes according to the review now and will add a short comment on this test.

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
wpmp: When you scroll through a PDF document the editor will not render annotations immediately for each page but it will wait for the `renderTimeout` until it renders annotations.
The reason is if you scroll through the document quickly you don't want a render request for each page, but only when you stop at some page and stay on it.
So the question is, how long should you have to wait until editor requests annotations.

reckart: Things appear to be "interactive" if the reaction time is 200ms or less. In some places in our code, we use a delay of 500ms. Using a longer delay may be sensible. By which rationale did you choose 1500ms?

wpmp: I chose it like this because scrolling through a page takes some time, so if it is too short it might trigger rendering too often.
That was my thought.
So when it renders the page will hang and the user can't continue scrolling through the document.
I think this could be annoying.

reckart: Is there a way to render n page ahead and start requesting new ones when the user is e.g. half-way through the already rendered pages?

wpmp: Still it would hang though.
Also the rendering process would take longer I guess.
Currently previous and next page is rendered along to the current page so that you can see annotations at page boundaries.

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
wpmp: With `window.alert()`?
Or do we have Wicket, Webanno or some other stuff for this?

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
wpmp: I think I found a cooler way.
What about if we render only when page changes and no movement in document is detected for 500ms ?
It feels pretty smooth to me and you don't have to scroll too fast through document if you want to get to a certain point without interruption.
I will commit this, then you can test and tell me then whether you like it or not, ok?

inception-pdf-editor/src/main/java/de/tudarmstadt/ukp/inception/pdfeditor/pdfanno/model/PdfExtractFile.java
wpmp: It is in the PdfExtract file.
It is used "when unicode mapping is unavailable".

wpmp:
Ah, changing labels now works again. :)

inception-pdf-editor/src/main/js/pdfanno/src/pdfanno.js
reckart: It sounds very good. Looking forward to testing it.

wpmp:
@reckart seems like the tests broke with webanno update?

wpmp:
I think it is fixed with removal of `state.setFirstVisibleUnit` usage.

reckart:
My guess would be that you should call `state.setPagingStrategy(new SentenceOrientedPagingStrategy())` whenever you create an `AnnotatorStateImpl` to properly fix this.

wpmp:

wpmp:
> My guess would be that you should call `state.setPagingStrategy(new SentenceOrientedPagingStrategy())` whenever you create an `AnnotatorStateImpl` to properly fix this.

Changed it like this.

reckart:
Nice 👍

Issue #1072, opened 2019-03-28
Support Wikidata fulltext search


reckart:
**Is your feature request related to a problem? Please describe.**
We can currently hardly use the official Wikidata SPARQL endpoint because INCEpTION is unable to use full text search on it.

**Describe the solution you'd like**
The Wikidata SPARQL endpoint does offer the ability to perform fulltext searches. We need to figure out how to integrate this:

https://www.mediawiki.org/wiki/Wikidata_Query_Service/User_Manual/MWAPI#Find_all_entities_with_labels_"cheese"_and_get_their_types

**Describe alternatives you've considered**
Wikidata uses blazegraph which also offers integrated fulltext search, but this capability has been disabled on the Wikidata SPARQL endpoint: 
* https://wiki.blazegraph.com/wiki/index.php/FullTextSearch
* https://phabricator.wikimedia.org/T141813

**Additional context**
This is a prerequisite for #784

PR #1078, opened 2019-03-29
#1072 - Support Wikidata fulltext search

reckart:
**What's in the PR**
- Added support for using the Wikidata entity search service
- Updated Wikidata profile to use that service
- Added SPARQLQueryBuilder tests using the Wikidata entity search service
- Moved some of the SPARQLQueryBuilder tests around to group similar tests next to each other
- Turned all restriction statements in limitToInstances into filters. The first one had not been a filter which caused serious performance drops when the query contained a UNION, e.g. when using withLabelMatchingExactlyAnyOf with multiple terms (observed on Wikidata SPARQL endpoint).

**How to test manually**
* Create a new KB using the Wikidata profile (the official Wikidata, not the UKP one!)
* Try performing entity linking
* Try searches on the KB page for KB items

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


reckart:
> 2019-04-01 17:05:25 WARN [SYSTEM] ConceptLinkingServiceImpl - Mention sentence could not be determined. Skipping.

If you can reproduce this, please let me know / open an issue for it.

> Unsupported value type: dateTime

In another PR in progress, this is still logged but without the stack trace.

PR #1026, opened 2019-03-10
#1023 - Query for statements via SPARQLQueryBuilder

reckart:
**What's in the PR**
- Added asStatements() to SPARQLQueryBuilder
- Removed requirement to add a primary condition before asking to retrieve labels or descriptions
- Close query results in SPARQLQueryBuilder.listStatements

**How to test manually**
* Use the KB page, browse some classes and instances, maybe make some modifications to existing statements, delete some, add new ones

**Automatic testing**
* [x] PR includes unit tests

**Documentation**
* [ ] PR updates documentation

**Blocked by**
* [x] #1022 - changes of PR #1025 are included in this PR, so #1025 should be reviewed and merged first

Issue #1023, opened 2019-03-10
Query for statements via SPARQLQueryBuilder


reckart:
**Describe the refactoring action**
Query for statements via SPARQLQueryBuilder to be used when working with statements lists on the KB page.

**Expected benefit**
Better maintainability and performance.

Issue #1020, opened 2019-03-08
Slow displaying of KB items with many IRI references


reckart:
**Describe the bug**
In the KB page, loading an item which has many IRI references is pretty slow. The problem is that for each of these references, we instantiate a IRIValuePresenter which then queries for the label (actually KBObject) of the IRI reference such that we can display a human-readable value instead of just an IRI.

**To Reproduce**
1. Set up DBPedia 
2. Go to the KB page
3. Select some class and them some instance
4. See that loading takes a while

**Expected behavior**
The UI just be more responsive.

One way to mitigate this problem would be to make the statement groups collapsible and to collapse them by default. Then we'd only have to load the presenters when the user expands a group to view it in detail. Groups for "important" properties could be expanded by default (e.g. label/description).

Another option would be to bulk-resolve all IRIs to labels at once, e.g. directly after loading the item statements and then to pass these on to the presenters somehow. The bulk-resolve should happen probably in `StatementsPanel.getStatementGroupBeans()`.

**Screenshots**

This shows that there are many queries:
```
2019-03-08 11:48:17 TRACE [rec] NoReification - listStatements called
2019-03-08 11:48:17 TRACE [rec] NoReification - NoReification.listStatementsForInstance took 33 ms
2019-03-08 11:48:18 TRACE [rec] NoReification - NoReification.listStatementsForInstance took 14 ms
2019-03-08 11:48:18 WARN [rec] NoReification - Properties with blank node values are not supported
2019-03-08 11:48:22 DEBUG [rec] SPARQLQueryBuilder - [58cf7a4a] Query returned a result in 19ms
2019-03-08 11:48:22 DEBUG [rec] SPARQLQueryBuilder - [6882c8ea] Query returned a result in 13ms
2019-03-08 11:48:22 DEBUG [rec] SPARQLQueryBuilder - [7f29eb75] Query returned a result in 178ms
2019-03-08 11:48:22 DEBUG [rec] SPARQLQueryBuilder - [279e1abd] Query returned a result in 40ms
2019-03-08 11:48:23 DEBUG [rec] SPARQLQueryBuilder - [4f2dd332] Query returned a result in 32ms
2019-03-08 11:48:23 DEBUG [rec] SPARQLQueryBuilder - [64ffe92b] Query returned a result in 35ms
2019-03-08 11:48:23 DEBUG [rec] SPARQLQueryBuilder - [3dda90f6] Query returned a result in 35ms
[...]

PR #1092, opened 2019-04-01
Bugfix/140 no statements with reification enabled

reckart:
**What's in the PR**
- Rewrote code which fetches the statements for the Wikidata reification strategy
- Statements can now be retrieved and are shown in the UI
- ... however, the retrieval strategy is now no longer aligned with the saving of statements, causing several unit tests to fail
- Added a new WikiDataReificationTest class
- Also: the KB page fires **REALLY MANY** queries in reified more (even more than in non-reified mode) which causes the official Wikidata SPARQL endpoint to trigger its rate limit. This will probably be addressed by #1020 but the present scope of #1020 may have to be extended to cover more cases causing slowness (or more issues need to be opened)

**How to test manually**
* Configure a reified Wikidata KB
* Try viewing some instance which has only a few statements

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

**Blocked by**
* [ ] #1023 - changes of PR #1026 are included in this PR, so #1026 should be reviewed and merged first


reckart:
Ok, I see what you mean. The Wikidata SPARQL endpoint serves reified and non-reified data:

- when reification is **enabled**, we **only** see the reified data - this excludes for example the descriptions and the labels because these are only available as non-reified data in Wikidata
- when reification is **disabled**, we see the non-reified data as well as cryptic statements which are actually reified data (which we see as proper statements when reification is enabled)

This is indeed a problem, but I think it should be treated as a separate issue.

PR #1118, opened 2019-04-12
# 542 recommender layer anchor mismatch

UWinch:
**What's in the PR**
The anchoring mode (e.g. lock-to-token) can be changed by the user even after a layer has been created. Recommenders usually do not work for all modes - only for certain ones. This means that the layer configuration can become inconsistent with the configured recommenders.

- When a recommender job is run (train/evaluate/predict) the scheduler must check if the recommender still accepts the layer configuration - if not, display an error to the user and skip the recommender
- On the recommender tab in the project settings warnings should be shown if recommenders
are inconsistent with the configuration of the layer they are defined on

**How to test manually**
* Create a layer with a given anchoring mode
* Create a recommender that supports this mode
* Change the anchoring mode of the layer to something the recommender does not support
* Go to the project settings Recommender tab, an error message should be shown
* Go to the annotation page and open the recommender sidebar, an error message should be shown

**Automatic testing**
* [x] PR includes unit tests

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
@@ -51,12 +51,11 @@ import org.apache.wicket.markup.html.panel.Panel;
 import org.apache.wicket.model.CompoundPropertyModel;
 import org.apache.wicket.model.IModel;
 import org.apache.wicket.model.PropertyModel;
+import org.apache.wicket.request.cycle.RequestCycle;
 import org.apache.wicket.spring.injection.annot.SpringBean;
 import org.apache.wicket.validation.IValidatable;
 import org.apache.wicket.validation.IValidator;
 import org.apache.wicket.validation.ValidationError;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import de.tudarmstadt.ukp.clarin.webanno.api.AnnotationSchemaService;
 import de.tudarmstadt.ukp.clarin.webanno.api.WebAnnoConst;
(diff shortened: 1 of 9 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
```diff
@@ -352,16 +352,19 @@ private void autoUpdateName(AjaxRequestTarget aTarget, TextField<String> aField,
...
-        // check if recommender and layer still match
-        RecommendationEngineFactory factory = recommenderRegistry
-                .getFactory(aToolModel.getObject().getKey());
-        if (!factory.accepts(recommender.getLayer(), recommender.getFeature())) {
-            error(String.format("Recommender %s configured with invalid layer or feature.",
-                    recommender.getName()));
-            Optional<AjaxRequestTarget> target = RequestCycle.get().find(AjaxRequestTarget.class);
-            if (target.isPresent()) {
-                target.get().addChildren(getPage(), IFeedback.class);
+        if (recommenderModel.getObject() != null && aToolModel.getObject() != null) {
```
reckart: @jcklie would say: check if one of the objects is null and then return early.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
```diff
@@ -347,12 +347,15 @@ private void autoUpdateName(AjaxRequestTarget aTarget, TextField<String> aField,
...
     /**
-     * Check if the selected recommender still accepts the configured layer and feature.
-     * If not show an error message.
+     * Check if the selected recommender still accepts the configured layer and feature. If not show
+     * an error message.
      */
     private void checkRecommenderLayerMatch(IModel<Pair<String, String>> aToolModel)
     {
-        if (recommenderModel.getObject() != null && aToolModel.getObject() != null) {
+        if (recommenderModel.getObject() == null || aToolModel.getObject() == null) {
```
reckart: Since you return in the "if" block, the "else" block is not required. Cf. http://blog.timoxley.com/post/47041269194/avoid-else-return-early

UWinch: Thx

Issue #1089, opened 2019-03-31
PDF Editor Viewport undefined


wpmp:
**Describe the bug**
Sometimes when opening a PDF document in the PDF editor it states `TypeError: Cannot read property 'viewport' of undefined`.
However when clicking on `close` and triggering rendering of annotations the editor just works fine.
I don't know yet how to reproduce with a 100% chance and whats the exact problem.

**To Reproduce**
1. Open a PDF in PDF editor
2. Open another PDF in PDF editor
3. Repeat till error occurs
4. Changing to Brat editor and back to PDF editor also sometimes triggers this error

**Expected behavior**
This error should not occur.

**Screenshots**
![2019-04-01-003718_482x174_scrot](https://user-images.githubusercontent.com/5390724/55296200-6d395b80-5416-11e9-80dd-8031849618d3.png)

```
[Log] Trace
	Console Evaluation (Console Evaluation 1:3)
	evaluateWithScopeExtension
	_evaluateOn
	_evaluateAndWrap
	mergeRects (pdfanno.page.bundle.js:1:71145)
	newInstance (pdfanno.core.bundle.js:1:211601)
	(anonymous function) (pdfanno.core.bundle.js:1:233674)
	forEach
	(anonymous function) (pdfanno.core.bundle.js:1:233578)
	forEach
	importAnnotations041 (pdfanno.core.bundle.js:1:233533)
	(anonymous function) (pdfanno.core.bundle.js:1:232828)
	forEach
	(anonymous function) (pdfanno.core.bundle.js:1:232556)
	initializePromise
	Promise
	importAnnotations (pdfanno.core.bundle.js:1:232451)
	importAnnotation (pdfanno.page.bundle.js:80:20669)
	importAnnotation
	(anonymous function) (Anonymous Script 1 (line 1186:18318))
	Eval Code (Anonymous Script 1 (line 1186:18395))
	eval
	(anonymous function) (wicket-ajax-jquery.js:1008)
	run (wicket-ajax-jquery.js:148)
	proxy (jquery-2.2.4.js:497)
	processNext (wicket-ajax-jquery.js:163)
	start (wicket-ajax-jquery.js:175)
	complete (wicket-ajax-jquery.js:738)
	fire (jquery-2.2.4.js:3187)
	fireWith (jquery-2.2.4.js:3317)
	done (jquery-2.2.4.js:8772)
	(anonymous function) (jquery-2.2.4.js:9123)
```

**Please complete the following information:**
[...]

PR #1122, opened 2019-04-16
#1089 - PDF Editor Viewport undefined

reckart:
**What's in the PR**
- Avoid accessing the first rendered page if it has not been rendered yet

**How to test manually**
* Basically open and PDF - for me the error described in the issue happened *always* when opening a PDF document after coming from the dashboard page

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-pdf-editor/src/main/js/pdfanno/src/page/textLayer.js
+++ b/inception-pdf-editor/src/main/js/pdfanno/src/page/textLayer.js
@@ -163,6 +163,19 @@ function withinMargin (x, base, margin) {
   return (base - margin) <= x && x <= (base + margin)
 }
 
+/**
+ * Returns the scaling factor of the PDF. If the PDF has not been drawn yet, a factor of 1 is
+ * assumed.
+ */
 function scale () {
-  return window.PDFView.pdfViewer.getPageView(0).viewport.scale
+// BEGIN INCEpTION EXTENSION - #1089 - PDF Editor Viewport undefined
+/*
+  return window.PDFView.pdfViewer.getPageView(0).viewport.scale;
+*/
+  if (window.PDFView.pdfViewer.getPageView(0) === undefined) {
+    return 1;
+  } else {
+    return window.PDFView.pdfViewer.getPageView(0).viewport.scale;
+  }
+// END INCEpTION EXTENSION - #1089 - PDF Editor Viewport undefined
 }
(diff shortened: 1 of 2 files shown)


reckart:
I found it pretty easy to debug. I set the JS debugger in the browser to stop on any exception and then waited until this particular exception turned up. Then I inspected the stack trace in the JS debugger.

Are you sure you rebuilt the JS code and cleared your browser cache before testing?

wpmp:
Yes, I do so too, however there are many usages of `viewport` and as the error gives no clue where it comes from there are many possibilities I think.

I am, I can see the additions you made in the JavaScript files.
I made a clean install and force reloaded in the browser.
For me it never happens immediately as it did for you, it always occurs after opening some PDFs in the PDF editor.

wpmp:
Oh wait, I didn't debug like you did.. Nevermind, I will try like this.

reckart:
> Yes, I do so too, however there are many usages of viewport and as the error gives no clue where it comes from there are many possibilities I think.

That is why I set a breakpoint on exceptions. It makes the debugger stop whenever an exception is thrown and in this way I can locate the code from which the exception originates.

wpmp:
Yes, sorry, was too quick on this.
For me it happens in `pdfanno/src/core/src/render/appendChild.js` on line 29.
Will you have a look at it?
Otherwise I maybe can further look into this tomorrow.

reckart:
From a quick look I would say the method should be made to return if `window.PDFView.pdfViewer.getPageView(0) === undefined`. But if I cannot reproduce it, it is difficult to fix it.

reckart:
An alternative to changing all the code which fetches the viewport would probably be to defer the loading of the annotations until a viewport is actually available, e.g. by listening to the "pagerendered" event...

wpmp:
> From a quick look I would say the method should be made to return if `window.PDFView.pdfViewer.getPageView(0) === undefined`. But if I cannot reproduce it, it is difficult to fix it.

Well, that's true :)
I'll try to work on this tomorrow.

> An alternative to changing all the code which fetches the viewport would probably be to defer the loading of the annotations until a viewport is actually available, e.g. by listening to the "pagerendered" event...

This is probably the best approach.
Do we still keep your commit then if this works?

reckart:
I'd wait on deciding that until an alternative approach has been found to be working :)

wpmp:
@reckart this commit seems to do the trick for me.

Issue #867, opened 2019-01-14
Group search results by feature value


reckart:
**Is your feature request related to a problem? Please describe.**
When refactoring annotations (e.g. changing many named entities with the label `LOCATION` to the label `CITY`) it would be beneficial to be able to group search results by feature values.

Consider the user running a search like `<NamedEntity/>`. The results are now grouped by documents and it is not possible to see what each entity is annotated as. 

**Describe the solution you'd like**
An option to group search results by a feature. E.g. considering the example above, it would be nice to see the entity matches grouped by "LOCATION", "PERSON", etc. In this way, the annotator could more conveniently find the relevant annotations to refactor. 

- [x] Add choice to group by document / feature value
- [x] Offer user to select one (or maybe more) features
- [x] For feature grouping, the values of the selected features are then displayed as the group headers in the search results

**Describe alternatives you've considered**
None so far.

**Additional context**
Probably best to do this one before moving on to implement #705.

mariaBio:
All four options are very useful. 
The 1st one gives all the flexibility to the user. 
The 2nd option is handy when user believes that only a restricted number of feature values have been used during the annotation, or user is interested in a specific feature subset. 
The 3rd helps to ease and speed up the revision process.
The 4th one compensates to some extent for absence of the 3rd feature, if only the 1st option is offered.

MdeBoer95:
@reckart I have the following problem for layers where stacking is enabled:
![grafik](https://user-images.githubusercontent.com/37872452/58748729-0f100f80-847d-11e9-8076-560403de0e61.png)
In this example I have two stacked Chunk annotations at token 55. One of them has a feature value Tag = "myTag" (I guess that's elements 10 and 11) and the other one has no feature value (element 12 then). Is there any way to find out whether token 11 belongs to 10 or 12 other than looking at the order in the list ?

reckart:
I don't see any way how we can tell to which "Chunk" the "Chunk.Tag" belongs by looking at the index data. There are a few fields in the `MtasToken` which I believe we do not use yet, e.g. the "real offset" fields - but I am not sure if we can simply abuse them to store the UIMA Feature Structure addresses in there....

reckart:
What are the bracketed `[-------.------]` thingies in your screenshot? Maybe these are fields we could abuse?

MdeBoer95:
This one is the "real offset". The empty ```[----]``` fields at the end are ```parentId``` and ```tokenPayload```.

reckart:
[...]

PR #1143, opened 2019-04-30
#867 - Group search results by feature value

MdeBoer95:
**What's in the PR**
* Give the user the option to group the search results by feature value

**How to test manually**
* Go to the annotation page
* Search for something in the search-sidebar and group by feature value
* See if grouping works properly 

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


MdeBoer95:
@reckart You can have a look at this approach now. I think it makes sense to include the feature value for each search result at a lower level to make the access easier in SearchAnnotationSidebar.java . Probably adding it at [MtasDocumentIndex.java](inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java))

Issue #808, opened 2018-12-30
Project settings panel for simulated evaluations of recommenders


Sibgha360:
**Is your feature request related to a problem? Please describe.**
We want to evaluate the performance of recommenders without having to annotate the document manually on the website.

**Describe the solution you'd like**
* Add a panel in the recommender settings.
* In the panel, give the user option to upload a pre-annotated document. 
* With the help of the annotations of the uploaded document, plot the learning curve diagram for the selected recommender.

PR #1158, opened 2019-05-08
#808 Project Settings Panel for Simulated Evaluation

Sibgha360:
- Bug fix

**What's in the PR**
* Javascript bug fix
* Removed `tag.put("my:canvas.chartid", getMarkupId());`


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/chart/ChartPanel.java
```diff
@@ -97,7 +95,7 @@ public void renderHead(IHeaderResponse aResponse)
                 "   $.ajax({",
                 "       url:'" + chartAjaxBejavior.getCallbackUrl().toString() + "',",
                 "       type:'post',",
-                "       cache:!1,",
+                "       cache:1,",
```
UWinch: According to the documentation http://api.jquery.com/jquery.ajax/ , we don't need the `cache` parameter in POST requests. For other requests, in my opinion, it would have made sense to set cache to false because the chart data changes between requests.

Issue #1162, opened 2019-05-10
Update checkstyle maven plugin


reckart:
**Describe the refactoring action**
Update checkstyle maven plugin.

**Expected benefit**
Use current version of the plugin.

PR #1163, opened 2019-05-10
#1162 - Update checkstyle maven plugin

reckart:
**What's in the PR**
- Update plugin
- Update checkstyle version
- Add workaround for Eclipse checkstyle integration quirk

**How to test manually**
* Import project into an Eclipse which is prepped with the Checkstyle Eclipse integration plugins and see that the settings from Maven are properly recognized
* Run a build and see that checkstyle doesn't crash

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@UWinch I basically use the same config in INCEpTION now as are used in DKPro Core and WebAnno - don't see any reason really why it shouldn't work.

Issue #310, opened 2018-06-20
Always show description of linked concept in popover 


reckart:
When a concept is proposed by the NamedEntityLinker recommender, the description of the concept is shown in the mouse popover of the suggestion. However, once the suggestion has been accepted, the mouse popover no longer shows the concept description. This is a problem because then it is no longer clear if the mention has been linked to the correct concept.

When there a concept feature is rendered, its description and identifier must be added into the mouse popover.

Also when regular recommender (e.g. String matching recommender) proposes a value for a concept feature, the description and identifier of the proposed entity must be shown in the popover. Just having the label is not enough to tell if the suggestion is correct or not.

rcffc:
@Rentier @reckart  I once asked whether we could add a _description_ feature that would be saved in the CAS, but I think one of you declined . Unfortunately, I cannot find the issue anymore. Do you remember what you suggested instead?

reckart:
If I remember correctly, I suggested that the description would be obtained from the KB during the rendering phase.

PR #1166, opened 2019-05-10
#310 - Always show description of linked concept in popover

reckart:
**What's in the PR**
- Add entity label and description as a lazy detail to the popover
- Transfer language and description to handle when creating KBHandle from KBObject

**How to test manually**
* Link an annotation to an entity via a concept feature (preferably one which has a description)
* Move mouse over annotation to trigger popover
* See label and description being shown in the popover

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I think I can fix this quickly and add it to this PR.

reckart:
Fixed it, but I have to wait for the WebAnno master build to go through because I also had to make minor changes there. I'll push the changes here already, but the first build will fail.

UWinch:
Tested again, the description is now also shown for recommendations. However, the word "description" is used two times in the suggestion-popover, once for the comment-text and once for the id. I wouldn't mind this, depends on how you feel about it.
popover_recommendation

Issue #1187, opened 2019-05-19
Upgrade dependencies (0.9.1)


reckart:
**Describe the refactoring action**

- [x] Spring 5.1.5.RELEASE -> 5.1.7.RELEASE
- [x] Sprint Boot 1.5.19.RELEASE -> 1.5.21.RELEASE
- [x] Wicket JQuery UI 8.1.0 -> 8.3.0
- [x] RDF4J 2.5.0-inception-1 -> 2.5.1-inception-1

**Expected benefit**
Latest bugfixes and security updates.

PR #1193, opened 2019-05-20
#1187 - Upgrade dependencies

reckart:
**What's in the PR**
- Upgrade RDF4J to be more robust against DBPedia issues

**How to test manually**
* Try opening "Barack Obama" in DBPedia via the knowledge base page

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
When I search for Obama and click on the first entry, then nothing happens; I do not get the right side to show the instance.

reckart:
So for me it works, but it seems that DBPedia may be a bit slow/flakey today. First time I tried it I got a connection timeout, when I tried again, it worked ok.

Issue #1132, opened 2019-04-25
Unify training data selection for evaluation


UWinch:
**Is your feature request related to a problem? Please describe.**
During evaluation, different recommenders select different amounts of training data given the same data and the same DataSplitter. To make recommender evaluation more transparent and comparable, all recommenders should use the same amount of training data during evaluation when the same DataSplitter is used.

**Describe the solution you'd like**
Some recommenders only consider data from sentences with at least one annotation, others use all data. It makes sense to use the former strategy for all recommenders.

reckart:
If you consider a sentence-based recommender (i.e. one that performs sentence classification), I guess it would be reasonable to train on non-annotated sentences, no?

reckart:
For POS recommenders, I think it may be desirable be a bit more restrictive on which sentences to train to avoid having a large `GAP` class. E.g. one could consider to train only on sentences in which at least 50% of the tokens have been annotated.

UWinch:
These are good points. So, we should keep the different data selection overall. However, we might consider changing the number of training instances displayed in the **EvaluationSimulation**, because like you, users might be surprised by seeing different amounts of data when the same data is used for evaluation. I see two alternatives:

- keep it how it is, document it in the user-guide
- show the amount of data before filtering out un-annotated data

reckart:
Maybe we can display the amount of unannotated/filtered out data in the chart?

PR #1199, opened 2019-05-23
#1132 Unify training data selection for evaluation

UWinch:
During evaluation, different recommenders select different amounts of training data given the same data and the same DataSplitter. To make recommender evaluation more comparable, the used training data is now given as ratio compared to all training data of the selected recommender for the evaluation simulation.

**What's in the PR**
* Implemented ratio of used training data as field in the evaluation result.
* Also added a message field to the result object to communicate possible errors to the user via the learning curve chart.

**How to test manually**
* Configure a project with a recommender and annotated data.
* Start an evaluation simulation on the recommender tab in the project settings.
* Check that the learning curve on the annotation page is still properly shown.

**Unrelated Changes**
* Update to admin guide regarding the learning.curve flag.


reckart:
> Set the learning curve to be enabled via the project.settings

I believe the learning curve is already "always enabled" and cannot even be turned off anymore.

UWinch:
You are right, the flag for the learning curve is gone. I'll use this issue to update the admin docu on this.

inception-imls-dl4j/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/dl4j/pos/DL4JSequenceRecommender.java
```diff
@@ -493,12 +493,17 @@ public EvaluationResult evaluate(List<CAS> aCas, DataSplitter aDataSplitter)
 
         int testSetSize = testSet.size();
         int trainingSetSize = trainingSet.size();
+        double trainRatio = (double) trainingSetSize / (double) (data.size() - testSetSize);
         
         if (trainingSetSize < 2 || testSetSize < 2) {
-            log.info("Not enough data to evaluate, skipping!");
+            String info = String.format(
+                    "Not enough training data: training set [%s] items, test set [%s] of total [%s].",
+                    trainingSetSize, testSetSize, data.size());
```
jcklie: You can directly log that with log.info and "{}" as place holders, no need for String.format.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/evaluation/EvaluationResult.java
```diff
@@ -36,7 +36,13 @@
 {
     private final int trainingSetSize;
     private final int testSetSize;
+    
+    /**
+     * Rate of this training data compared to all training data
+     */
+    private final double trainingDataRatio;
     private boolean skippedEvaluation;
+    private String errorMsg;
```
jcklie: I would like that String to be Optional

reckart: I think we should avoid `Optional` fields. If it makes sense to make the getter optional, then return `Optional.ofNullable(...)`.

jcklie:
When I use the Example Recommender Project, then I do not see a graph in the coordinate system and the x-axis is from -1 to 1.

UWinch:
Thx, I will do some debugging.

Issue #1203, opened 2019-05-24
Improve HTML annotation editor


reckart:
**Is your feature request related to a problem? Please describe.**
The HTML-based annotation editor is currently hardly more than a skeleton and needs to be improved.

**Describe the solution you'd like**
- [x] Display all annotations, not just the first few
- [x] Actually allow annotations to be added / edited / removed

PR #1204, opened 2019-05-24
#1203 - Improve HTML annotation editor

reckart:
**What's in the PR**
- Display all annotations in the CAS
- Customized AnnotatorJS (can be built from source)
- Support for color
- Support for HtmlDocumentReader from DKPro Core
- Support for selecting, creating and deleting annotations in the HTML editor component

**How to test manually**
* Switch to the HTML editor mode and try out stuff

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
First quick test:
I cannot select previously made annotations.
I cannot see the label, I need to hover
Colors change of existing annotations when I annotate a new layer
Hovering over an annotation without label says "No comment" which is a bit unuseful
Annotating a token on two layers only has color of one, I need to hover to see that I have two annotations

reckart:
AnnotatorJS is unable to render multiple annotations at the same location. The only way to access them is via the popup menu. That is a limitation of AnnotatorJS's rendering code. I do not think there is an easy fix for this.

Of the points you mentioned, only the color-change when adding annotations on a new layer sound like a bug. For the rest, we could only discuss how much it impacts the user and whether we can mitigate the problem somehow with reasonable effort.

PR #1221, opened 2019-05-29
Fix random search for java api

qpwo:
Related:
* inception-project/inception/pull/1181
* isi-vista/curated-training-annotator/issues/16
* inception-project/inception/issues/987

**What's in the PR**
* Fixed bug in `ElasticSearchProvider.java` where the term match score overshadows the random noise for randomly ordering search results

**How to test manually**
* In previous version with random order enabled, if you search a term, then more relevant results still show first. Now, results are really random

**Automatic testing**
* [ ] PR includes unit tests -- nope

**Documentation**
* [ ] PR updates documentation -- nope

**Remaining Problems**
* **Seeding is not working.** In one session, if you search the same term twice, then you will get results in a different order.


reckart:
@rgabbard Looks good to me. Shall we merge it?

Issue #1215, opened 2019-05-28
Write recommender evaluation scores to event log


UWinch:
**Is your feature request related to a problem? Please describe.**
After having now implemented more than one score for the recommender I want to add them to the `Details` field of the `LoggedEvent`. This will allow us to retrieve and show different metrics in the learning curve on the annotation page (see https://github.com/inception-project/inception/issues/1103). 

**Describe the solution you'd like**
Store the `EvaluationResult` object from the evaluation that is done in the `SelectionTask` in the `RecommenderEvaluationResultEvent` and then put the respective scores in the details json in the `RecommenderEvaluationResultEventAdapter`. 

**Describe alternatives you've considered**
To not compute all scores, one would have to either store the confusion-matrix (too big) or inform the selection task which metric to compute which I think is not feasible.

Issue #1103, opened 2019-04-04
Metric Selection for Recommender Evaluation


Sibgha360:
**Is your feature request related to a problem? Please describe.**
Right now, the user cannot select which metric they want to visualize on the learning curve graph in order to evaluate the recommenders' performance.

- [x] Learning Curve on recommendation Tab

- [x] Evaluation Simulation on evaluation page

**Describe the solution you'd like**
Add a drop down beside the chart where user has an option to change the metric; accuracy, recall, precision or F1.

PR #1225, opened 2019-05-31
#1215 Write recommender evaluation scores to event log

UWinch:
After having now implemented more than one score for the recommender I want to add them to the Details field of the LoggedEvent. This will allow us to retrieve and show different metrics in the learning curve on the annotation page (see #1103).

**What's in the PR**
Store the EvaluationResult object from the evaluation that is done in the SelectionTask in the RecommenderEvaluationResultEvent and then put the respective scores in the details json in the RecommenderEvaluationResultEventAdapter.

**How to test manually**
* Add recommender to your project
* Annotate and check that learning curve is still drawn accurately

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/log/RecommenderEvaluationResultEventAdapter.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/log/RecommenderEvaluationResultEventAdapter.java
@@ -67,13 +67,18 @@ public class RecommenderEvaluationResultEventAdapter
             Details details = new Details();
 
             details.recommenderId = aEvent.getRecommender().getId();
+            
             EvaluationResult result = aEvent.getResult();
             details.accuracy = result.computeAccuracyScore();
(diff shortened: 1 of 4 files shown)


inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/log/RecommenderEvaluationResultEventAdapter.java
```diff
@@ -66,7 +67,11 @@ public String getDetails(RecommenderEvaluationResultEvent aEvent)
             Details details = new Details();
 
             details.recommenderId = aEvent.getRecommender().getId();
-            details.score = aEvent.getScore();
+            EvaluationResult result = aEvent.getResult();
```
reckart: how about including the training and test set sizes as well?

UWinch: good idea, i will do it

Issue #1240, opened 2019-06-07
Allow adding external JARs to the standalone application


reckart:
**Is your feature request related to a problem? Please describe.**
Currently, one has to recompile and repackage the whole fat jar of INCEpTION in order to add functionality. But it would be nice if one could download "plugin" JARs and have the standalone application load them if available.

**Describe the solution you'd like**
When starting the standalone application, it should look for JAR files in `<application-home>/plugins` and add them to the classpath if any are found.

**Describe alternatives you've considered**
There is PF4J and its Spring Boot integration, but so far we didn't really get it to work, so let's try something simpler.

PR #1241, opened 2019-06-07
#1240 - Allow adding external JARs to the standalone application

reckart:
**What's in the PR**
- Implemented a custom launcher for the standalone JAR which checks for a "plugins" folder under the INCEpTION home directory and adds any JARs found there to the classpath.

**How to test manually**
* Build the standalone JAR
* Build the JAR for the `inception-example-imls-data-majority` module
* Copy the `inception-example-imls-data-majority` JAR to `.inception/plugins` (or adjust if you use a different home folder for INCEpTION)
* Start up with `-Dloader.debug`
* Check if the `inception-example-imls-data-majority` JAR is detected during startup as a plugin JAR, that the recommender it provides is available and that it works

NOTE: if a plugin requires any JARs which are not yet part of INCEpTION, it won't work. The launcher does not yet support "fat plugin JARs", i.e. plugins with nested JARs.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-app-webapp/pom.xml
+++ b/inception-app-webapp/pom.xml
@@ -827,17 +827,24 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-maven-plugin</artifactId>
             <version>${spring.boot.version}</version>
-            <configuration>
-              <executable>true</executable>
-              <layout>ZIP</layout> <!-- To use the PropertiesLauncher -->
-            </configuration>
             <executions>
               <execution>
                 <goals>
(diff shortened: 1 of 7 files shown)


reckart:
#1243 (PF4J-based plugins) probably supersede this one - we might still care to keep it though, e.g. in order to be able to drop in database drivers without having to package them up as plugins first.

reckart:
@reckart why is the build failing?!

reckart:

PR #1245, opened 2019-06-12
#628 Improve performance of loading results from external search

### LINKED ISSUE #628: Improve performance of loading results from external search
It takes ages to load results and the data that is returned from the endpoint seems to be huge.

- [ ] Change the query such that only relevant information is returned for the query results table: document id, document title (in our current case the URI), score, highlights
- [ ] Remove `text` field from `ExternalSearchResult`
- [ ] Introduce some class to represent a full external document, e.g. `ExternalDocument`
- [ ] for importing and displaying in the detail page, use `ExternalSearchProvider.getDocumentById()` which should return the `ExternalDocument` that contains a field with the full text

**Describe alternatives you've considered**
Load less query results (currently 1000 documents are loaded).

az79nefy:
**What's in the PR**
* query does not return source document information (e.g. text)
* therefore ExternalSearchResult does not store text anymore
* ExternalSearchProvider.getDocumentText() is used to get the text through a separate request
* resolves #628 

**How to test manually**
* search for documents in an external search repository (should be faster hopefully)

**Automatic testing**
* [X] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
+++ b/inception-external-search-elastic/src/main/java/de/tudarmstadt/ukp/inception/externalsearch/elastic/ElasticSearchProvider.java
@@ -24,7 +24,6 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.http.HttpHost;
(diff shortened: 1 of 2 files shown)


az79nefy:
In this PR I am not setting any keyword offsets anymore. 
Are the offsets in `ExternalSearchHighlight.java` still used?
The keywords in the search highlights are correctly marked without offsets (in the search page and sidebar).
It seems like the offsets are only requested in the `highlightKeywords` method of the `ExternalSearchAnnotationSidebar` when `onRenderAnnotations` is executed, and that method is unused.

Issue #1239, opened 2019-06-07
Support recommendation of simple relations


jcklie:
**Is your feature request related to a problem? Please describe.**
Recommender right now cannot predict simple relations

**Describe the solution you'd like**
- [x] Add a simple "String matching " recommender for relations (see #406)
- [x] Add a renderer for this type of recommendations
- [ ] Group suggestions for same source and target
- [ ] Add traits to relation recommender for feature of source and target layer that is used for recommendation
- [ ] Render confidence, hint, explanation, ...
- [ ] Add accepting and rejecting
  - [X] Accepting
  - [ ] Rejecting: rejecting arcs does not trigger `DoActionResponse`
- [ ] Add hiding of recommendations

jcklie:
The suggestions can now be received in the `RecommendationRelationRenderer`. Right now, I cannot get back the source and target fs back from the relationposition, the low level CAS says they are not in there. The addresses in the `aCas` and the position look different when I manually select them.

reckart:
The IDs are re-generated when the user opens a document. While the user is editing the document, the IDs should be stable. Mind to fetch the IDs from the `original CAS`, not from the `prediction CAS`.

jcklie:
@reckart  Rejecting: rejecting arcs does not trigger DoActionResponse

jcklie:
@reckart  Is there a reason why we both have an abstract class `AnnotationSuggestion_ImplBase` and an interface `AnnotationSuggestion`? Cant we put all in the abstract class?

reckart:
So we don't have to use the longish `AnnotationSuggestion_ImplBase` everywhere (`_ImplBase being a marker for abstract classes`)...

reckart:
Btw.: https://github.com/webanno/webanno/issues/1579

jcklie:
We could just call the abstract class `AnnotationSuggestion`

reckart:
If you have a strong feeling about it, rename it.

jcklie:
I just dont understand why we have an interface AND an abstract class instead of just an abstract class. I do not care about the name, just about having 2 bases which could be one.

reckart:
Well, as I said: just the habit of appending `_ImplBase` as a marker to abstract classes and that `AnnotationSuggestion_ImplBase ` is quite an unwieldy name. There is no technical reason. Feel free to refactor if you find it annoying.

jcklie:
I started implementing rejections, but now I have issues with getting the layerId from the VID.

PR #1246, opened 2019-06-13
#1239 - Support recommendation of simple relations

jcklie:
# What's in the PR
- Rename String Recommender package
- Start simple projecting relation recommender
- Renamed `AnnotationSuggestion` to `SpanSuggestion`
- Introduced `AnnotationSuggestion` interface on top of `AnnotationSuggestion_ImplBase`

# How to test manually  

_**This PR removes columns from the database, switching to another branch after running this might be difficult before it is merged!**_

* Create a new project and upload a document
* Define a new span layer with a String feature 
* Add a relation layer ontop of that previously created layer and add a relation layer
* Add a string relation recommender on that relation layer
* Mark a span "a" and a span "b" in the same sentence
* Draw a relation between them and give it a label
* In another sentence, also mark a span "a" and a span "b" in the same sentence
* It should recommend the label you defined for the first sentence

* Check that active learning and recommender still work
* Check that existing string recommender in other projects still work

**Automatic testing**
* [x] PR includes unit tests

**Documentation**

**Things missing:**

- [x]  Check active learning
- [x] Write unit tests
- [x] Sorting by confidence
- [x] Max suggestions
- [x] Rejection hiding and learning record
- [x] Lazy details (confidence, label)
- [x] Several recommender grouping


jcklie:
@reckart This is the relation recommendation PR.

reckart:
Lots of conflicts...

jcklie:
I can have a quick look today why that is, I assume it is just because of renaming stuff.

reckart:
It would be useful to have some checklist in the PR telling what is done and what still needs to be done.

jcklie:
This PR implements a  simple relation recommender. It does not contain code for creating `AnnotationSuggestion` pendants for relations and rendering yet. I thought about it and am not sure how to best refactor `AnnotationSuggestion` to also handle relations, coreference and the like. We can have an abstract `AnnotationSuggestion` class and then a `SpanSuggestion` and `CoreferenceSuggestion` and `RelationSuggestion`. We would e.g. move `begin`, `end`, `coveredText` from `AnnotationSuggestion` to `SpanSuggestion`. We have to think how active learning should work with relations. Later, we have to think about how we can build relation recommendations based on other span recommendations.

jcklie:
I wanted to call the base AnnotationSuggestion and the other ones like SpanSuggestion or so.

reckart:
Refactoring is in progress. Be sure there will be more renaming :)

reckart:
If we rename too much now, we'll have issues merging in changes from master.

jcklie:
I have trust in your judgement.

reckart:
@jcklie I made the changes we discussed. I'll approve the changes so you can have another look and eventually merge this as the first PR towards relation recommendation.

UWinch:

jcklie:
Currently I do not know how to add relation suggestions to the learning log, as that wants begin/end offsets. I could save source and target XMI ids like I do in other places, but these are not persistent and therefore we cannot save them in the DB.

reckart:
These were the two things I noted when going through the changes:

* upsert relation should create new relation if stacking is allowed
* make "adjuct feature" in relation traits configurable - or maybe use all the label features

@jcklie mentioned that the relation recommender is probably not enabled for stacking layers anyway and for the other thing, we have an issue now as far as I saw.

jcklie:
The adjunct feature is configurable now and I added documetnation

UWinch (commented):
Does not work for me yet, I'm not getting any recommendations. I see the following error in the logs, maybe this has something to do with it:

```

2021-03-29 13:16:34 ERROR [SYSTEM] SelectionTask - [admin][[NERel@relation] String Matcher for relations]: Failed
java.lang.NullPointerException: null
	at de.tudarmstadt.ukp.inception.recommendation.tasks.SelectionTask.run(SelectionTask.java:200) [classes/:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628) [?:?]
	at java.lang.Thread.run(Thread.java:834) [?:?]

```

I also get a `SQLSyntaxErrorException: Unknown column 'offsetCharacterBegin' in 'where clause'` when I accept recommendations of normal string matchers.

jcklie:
@UWinch Thank you for the feedback,  incorporated it. I cannot reproduce the NPE, can you check in your database whether the db changes were applied correctly?

jcklie:
@UWinch I think I found the error, I will fix it asap.

jcklie:
@UWinch I fixed it, you need to annotate enough if you decide to not skip evaluation.

UWinch:
I'm now getting a `Error: SQLIntegrityConstraintViolationException: Column 'suggestionType' cannot be null` when accepting recommendations from normal Stringmatchers.

The relation recommenders work well for me, the robot is the cutest :robot: However, suggestions seem to be shown in double
rel-rec

Issue #1085, opened 2019-03-30
Creating Stacking Spans in PDF editor


wpmp:
**Describe the bug**
It is not possible to create a span annotation within the text of another span annotation.
Text selection does not work on the text of a span annotation.
However it works to make a new span if it starts before an existing span but covers it completely.
Partial covering of an existing annotation however does not work.

**To Reproduce**
1. Open a PDF document in the PDF editor
2. Create a span annotation
3. Try to create another span annotation on the same text or on text within the created span annotation

**Expected behavior**
It should be possible to create multiple span annotations on the same text if the layer allows stacking spans

**Please complete the following information:**
 - Version and build ID: 0.9.0-SNAPSHOT (2019-03-30 13:13:11, build 27391004409d33da4da95540dbe7524fbb01c108)
 - OS: Linux
 - Browser: Google Chrome

**Additional context**
Rendering stacking spans, which were created in the brat editor or produced by the recommenders, works fine.

PR #1230, opened 2019-05-31
#1226 - Clean up PDFAnno code

wpmp:
**What's in the PR**
- remove change comments

**How to test manually**
* Start INCEpTION
* Play around with PDF editor

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**
* [ ] PR updates documentation

PR #1254, opened 2019-06-20
#1085 - Creating Stacking Spans in PDF editor

wpmp:
**What's in the PR**
* Enables the creation of stacking spans in the pdf editor

**How to test manually**
1. Open a document in the PDF editor
2. Create a span
3. Create another span that shadows the previously created span

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


wpmp:
This PR is based on #1230
It is ready for a review after #1230 was merged

UWinch:
This does not work for me. I tried with custom layers:
* span, token-level, any overlap, crossing sentences
* span, char-level, only stacking
* span, single-token, stacking only.

Maybe I'm doing something wrong? I'm clicking and dragging to create the first annotation and then try to click into the annotation again but cannot select anything inside it. However, when I annotate stacked annotations in the brat-view, these stacked annotations do show up in the pdf-view.

wpmp:
@UWinch which browser do you use?

reckart:
@UWinch if you are using Eclipse, make sure you follow these steps to ensure you have the latest JS code for the PDF module:

* switch to the PR branch
* select all inception modules
* right click -> Maven Update (necessary to trigger a rebuild of the JS code in the pdf module)
* in the browser on the annotation page with the PDF editor activated, clear cache and force-reload to ensure that the browser uses the latest JS code

wpmp:
Nice :)

Issue #1256, opened 2019-06-21
Integrate curation into annotation page.


UWinch:
**Is your feature request related to a problem? Please describe.**
Integrate curation into annotation page as a new Tab instead of the Curation Page on the Dashboard to benefit from already implemented features on the annotation page.

**Describe the solution you'd like**

- [x] The user should be able to open the curation tab and select the users whose annotations she wants to see. The annotations are shown in a differing color additionally to the users' annotations. 

- [x] The user can accept other users' annotations into her own annotation document or the curation document. 

- [x] An automatic merge of consistent annotations analogous to the existing implementation in the curation can be triggered by button click.
- [x] Aggregate other users' annotations as suggestions (e.g. 3 people annotated this span).
- [x] Hide other users' annotations if the same annotation was already merged.
- [x] Integrate some help on the feature for better usability.
- [x] The message `You need users with finished documents to start curating` is confusing on an already closed document, remove or instead display `This document has already been finished`. 

Implementation is done using an AnnotationEditorExtension to save the accepted annotations and trigger the render of the ones from other users.

reckart:
We have done all of this now for a while - although the feature is still marked as experimental.

PR #1257, opened 2019-06-21
#1256 Integrate curation into annotation page

UWinch:
**What's in the PR**
The user should be able to open the curation tab and select the users whose annotations she wants to see. The annotations are shown in a differing color additionally to the users' annotations.
The user can accept other users' annotations into her own annotation document or the curation document.
An automatic merge of consistent annotations analogous to the existing implementation in the curation can be triggered by button click.

Implementation is done using an AnnotationEditorExtension to save the accepted annotations and trigger the render of the ones from other users.

**How to test manually**
* Check out following webanno branch https://github.com/webanno/webanno/pull/1618
* Set up a project and annotate some documents with different users
* Go to the annotation page and open the curation tab.
* Select users whose annotations you want to see
* Accept some of the shown annotations into your curated document

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


UWinch:
@reckart @jcklie can be reviewed. I would like some feedback on the automatic merge option. It works analoguesly to the curation page one. This means the curation CAS changes its annotations to the annotations that all (other) users agree on. This can result in deleting annotations that the curator made. An option would be to only add the agreed upon annotations to the already existing one i.e. without deletions. What do you think?

reckart:
> An option would be to only add the agreed upon annotations to the already existing one i.e. without deletions. What do you think?

I think that would work - in particular since the merging code nowadays should be able to deal properly with issues such as not allowing that stacking agreed-on annotation to be merged on top of an annotation that the curator might already have made.

reckart:
@UWinch looks like this one depends on https://github.com/webanno/webanno/pull/1396 but that one is still marked as WIP?

UWinch:
> @UWinch looks like this one depends on [webanno/webanno#1396](https://github.com/webanno/webanno/pull/1396) but that one is still marked as WIP?

Sorry, forgot to remove the WIP label from that.

reckart:
I finally had the time to look at this PR in action and not just in code :)

I would suggest to remove the dropdown choice which allows to select between manual and automatic merging. If a user clicks on something, it it always manual merging. Instead, a button for "re-merge" should be added but outside of the "cogwheel" menu, i.e. directly in the sidebar. I would also rename the "merge" button in the sidebar to "apply" meaning that the selected users' annotations would be shown or hidden. Alternatively, it could be removed entirely and whenever the user (de)activates a checkbox, an AJAX update could be triggered and the display could be re-rendered immediately. But the button is probably the better idea since doing the AJAX thing would take much more time when (de)activating multiple users at once.

UWinch:
Made the last changes that Richard suggested, can be reviewed again. My idea with the dropdown was that we might want additional ways to merge e.g. majority based in the future, but it makes sense to keep the UI simpler for now.

reckart:
On the curation page, there is a popup when pressing the "remerge" button where merge options (or rather one option) can be configured. Do we have this popup here as well?

UWinch:
> On the curation page, there is a popup when pressing the "remerge" button where merge options (or rather one option) can be configured. Do we have this popup here as well?

Not the same pop-up, just one with a message warning the user that annotations might be lost when merging. Do we need the `merge incomplete` option here? You can just select one or less users if you want to merge only their (complete) annotations.

reckart:
The "merge incomplete" merges annotations that not all users made but on which the users that made them did not disagree. So lets say you have three words:

* user A annotated word 1 as X
* users A and B annotated word 2 as Y
* user B annotated word 3 as Z

Merging with incomplete annotations would result in X, Y and Z being merged.

If you only select user A or user B, then either X or Z would not be merged, right?

reckart:
Btw. the initial use-case for "merge incomplete" was the case that user A annotated only POS and user B annotated only Lemma and the curator wanted to merge the POS and Lemma annotations into a single document.

UWinch:
> Btw. the initial use-case for "merge incomplete" was the case that user A annotated only POS and user B annotated only Lemma and the curator wanted to merge the POS and Lemma annotations into a single annotation document.

I remember the use case now. I'll add the option.

reckart:
@jcklie you can probably test this using the *[German example project (POS + dependencies)](https://webanno.github.io/webanno/downloads/)* from WebAnno.

reckart:
Sounds like the file `marker-wicket-module` is missing from the module. This needs to be present if a module contains Wicket components so the HTML/properties files are picked up from `src/main/java`.

jcklie:
@UWinch I merged master, build is failing now.

UWinch:
@jcklie should work now

lfoppiano:
Hi, I've just updated to the latest version but I cannot find this feature. 

Could you share some more information about it or a screenshot? I can't find the curation tab in the annotation page, I'm not sure, though.

UWinch:
Hi @lfoppiano , this is still an experimental feature, that's why you need to enable it by adding `curation.sidebar.enabled=true` to your `settings.properties` file (needs to be created in your inception home folder which is by default `.inception`). Some more information can be found here: https://inception-project.github.io//releases/0.15.0/docs/user-guide.html#_curation.

We would be happy for your feedback on this feature if you decide to try it.

lfoppiano:
@UWinch I gave it a try. The feature is great, but I'm not sure what is the use case it was built for. 
Ideally, the nice thing about having annotation and curation separated is to define two separate steps in the process. 

In our case, we have 5 annotators, and 3 of them are also cross-checking documents through the annotation page. In this case for them is better to have the process of curation as a separate process of annotation. 

One nice feature would be to implement the same UI (staking up multiple layers of annotations on the same text) in the curation page, for example when only annotation from just one annotators have been completed.

UWinch:
Thanks for the feedback. We plan to eventually replace the separate curation page with the curation mode on the annotation page to be able to use some of the nice features that are already implemented for the annotation page like the stacking of annotations that you mentioned.

We'll keep your feedback in mind, that curation and annotation should be logically separate steps even if they are integrated in the UI in this way.

lfoppiano:
> Thanks for the feedback. We plan to eventually replace the separate curation page with the curation mode on the annotation page to be able to use some of the nice features that are already implemented for the annotation page like the stacking of annotations that you mentioned.
> We'll keep your feedback in mind, that curation and annotation should be logically separate steps even if they are integrated in the UI in this way.

Thank you. 

In general my only concern on mixing annotation and curation is that users might mistakenly save the resulting document as their own. 
If I may suggest, I would add a configuration flag that disables the checkbox `my document` and create a new curated document as output: 

image

reckart:
My thought is that we should move away from a singular curation document and instead have a more user-centric approach. That means the default target document should be the user's own document. It kind of assumes that a person who is an annotator *and* a curator might care to work with two user accounts, one with the role annotator and one with the curator role. 

The current curation concept on the curation page only supports a process with a single curation step. E.g. it is not possible to do a tiered curation. By tiered curation, I mean a process where you have annotators a1-a4 and curators c1-c3 where c1 and c2 each curate a pair of annotators and c3 curates the pre-curated data from c1 and c2. 

Another type of use-cases to be addressed by the curation option on the annotation page are those where annotators work more closely together and regularly want to see each others data.

lfoppiano:
@reckart I see your point. Indeed this way makes it more flexible to different workflows.

PR #1259, opened 2019-06-24
#1274 - Score explanation

munterkalmsteiner:
**What's in the PR**
Allows recommenders to provide an explanation (string) for the achieved score. Motivation: if the score is a combination of different predictors, it would be helpful for the annotator to see the contribution components of each predictor to make a decision. 

**How to test manually**
Currently, no official inception recommender uses this feature, hence it cannot be tested manually.

**Automatic testing**
The following existing unit tests were extended:
- StringMatchingRecommenderTest
- AnnotationSuggestionTest
- PredictionGroupTest
- RecommendationServiceImplIntegrationTest
- VisibilityCalculationTests

**Documentation**
No documentation updates


jcklie:
It looks ok, but I would like to see an example for that to showcase this feature, e.g. for the data majority recommender which is only an example.

jcklie:
Thank you for this PR, it looks nice! One last thing that would be good to have is to add a guide on how to set this explanation in the development documentation `inception-example-imls-data-majority/src/main/resources/META-INF/asciidoc/developer-guide/implementing_recommender_tutorial.adoc` . There we already state on how to create a recommender.

jcklie:

munterkalmsteiner:
I've fixed the formatting and updated the documentation. 
Note that the current [developer documentation](https://zoidberg.ukp.informatik.tu-darmstadt.de/jenkins/job/INCEpTION%20(GitHub)%20(master)/de.tudarmstadt.ukp.inception.app$inception-app-webapp/doclinks/3/#sect_recommendation) is broken:
image

I don't see where the problem is, though...

reckart:
I have updated the assembly configuration for the documentation to also fetch documentation from the inception-example-imls-data-majority - that fixes the missing include. I pushed it into this PR.

jcklie:
I already fixed that in #1315

reckart:
@jcklie if we move the documentation out of the example, we cannot use code includes in the asciiidoc documentation, right?

jcklie:
I thought we can, as we use absolute paths.

reckart:
We could use relative paths which point out of the usual documentation folder and into the general project structure. Adjusting the assembly though has the benefit that we can use the same inclusions style everywhere and don't have to resort to longish references in some documentation modules.

reckart:
@jcklie anything missing from this PR before it can be approved?

jcklie:
I want to merge my documentation first, then I will merge this afterwards.

jcklie:

reckart:

jcklie:
@munterkalmsteiner Thank you for this PR! This should land in INCEpTION 0.11.0 soon.

PR #1261, opened 2019-06-25
Sorted list was not used, hence recommendations appeared in random order

munterkalmsteiner:
**What's in the PR**
Fix an oversight that led to sorted recommendations being ignored, rendering them in random order.

**How to test manually**
Use a recommender with a score feature that produces more than 1 suggested annotation. 
Expected: first suggestion (bottom of list, closest to the text) is the one with the highest score
Actual: the suggestions are not sorted by score  

**Automatic testing**
No GUI testing adopted in project, hence no tests provided

**Documentation**
N/A

### DIFF
--- a/inception-active-learning/pom.xml
+++ b/inception-active-learning/pom.xml
@@ -20,7 +20,7 @@
   <parent>
     <groupId>de.tudarmstadt.ukp.inception.app</groupId>
     <artifactId>inception-app</artifactId>
-    <version>0.10.1-SNAPSHOT</version>
+    <version>0.11.0-SNAPSHOT</version>
     <relativePath>..</relativePath>
   </parent>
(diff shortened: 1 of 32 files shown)


reckart:
@munterkalmsteiner thanks for the PR (and the other one!) - A quick comment on this one: the order in which annotations are rendered is only indicative to the frontend (e.g. brat) which may choose to re-order them (e.g. for layout reasons).

reckart:

reckart:
I think it would be a good idea to re-base this to the `0.10.x` branch so it can be included in a bugfix release.

munterkalmsteiner:
@reckart Hm, was not able to rebase. Will create new PR based on 0.10.x

PR #1265, opened 2019-06-25
#1273 - Make prediction possible even if there was no training

munterkalmsteiner:
**What's in the PR**
A recommender may rely on a static model AND training. Therefore, we must mark
as ready for prediction, even if no training happened.


reckart:
Right now, every recommender internally has to call `context.markAsReadyForPrediction()` when it is done. I wonder if we should move this out of the recommenders and into `TrainingTask.run()`. But there are cases where trainable recommenders to not mark the context a ready at the end of training, e.g. the `OpenNlpPosRecommender` doesn't mark the context as ready if it was unable to extract and training data points from the CASes. This causes the entire recommender to be skipped during prediction instead of trying to call it with an untrained model. Thus, if we pull it out to `TrainingTask.run()`, there would be no way anymore to entirely skip the untrained recommenders - but I guess the performance impact would be minimal / hardly noticeable. 

We could also/additionally replace the boolean `recommendationEngine.requiresTraining()` by a method which could return `TRAINING_REQUIRED`, `TRAINING_SUPPORTED`, and `TRAINING_NOT_SUPPORTED`.

Any thoughts?

munterkalmsteiner:
Yeah, I think this makes sense, but I think a recommender needs to communicate to <code>TrainingTask.run()</code> whether training has succeeded or not in case 3.
Case 1 (NOT_SUPPORTED) -> mark as ready to predict
Case 2 (SUPPORTED) -> train if new annotations are there, then mark as ready to predict, independently of whether training has succeeded or not
Case 3 (REQUIRED) -> train if new annotations are there, then mark as ready to predict if training succeeded.

I thought of adding a boolean return value to <code>RecommendationEngine.train()</code>, but I guess you want to use <code>RecommenderContext</code> for communicating state in order to implement parallel execution somewhen in the future. 

I can work on this and submit an updated PR, just not in the coming weeks (some teaching then vacaction in July are coming up).

munterkalmsteiner:
@jcklie I've implemented the changes as suggested by @reckart. Not sure though if this is in line with your long-term design vision.

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/recommender/RecommendationEngine.java
```diff
@@ -31,6 +31,12 @@
 import de.tudarmstadt.ukp.inception.recommendation.api.model.Recommender;
 
 public abstract class RecommendationEngine {
+	
+	public enum RecommendationEngineCapability {
```
jcklie: I would move this to a seperate class, as we use this field outside of `RecommendationEngine`.

jcklie:
@munterkalmsteiner  Thank you for the PR. Before I test this: 
1) This PR wants to merge into 0.10, which is a bug fix only branch. I think that that is not good, as this is a feature. @reckart This should be in master, should it not?
2) Please adhere to the [DKPro Pull request Style ](https://dkpro.github.io/contributing/) next time, I will squash it at the end and fix it this time.

jcklie:
Its not your fault, its ours that we do not present the process more clearly. For that I created #1271 . We are very happy that you are contributing.  You also need to add a license header to the newly created class I think, as the build will fail otherwise.

munterkalmsteiner:
done ;)

reckart:
I had suggested to treat #1262 as a bug and target it to 0.10.x. I agree this PR would better go to master.

inception-example-imls-data-majority/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/datamajority/DataMajorityNerRecommender.java
```diff
@@ -65,10 +66,15 @@ public void train(RecommenderContext aContext, List<CAS> aCasses)
 
         DataMajorityModel model = trainModel(annotations);
         aContext.put(KEY_MODEL, model);
-        aContext.markAsReadyForPrediction();
     }
 // end::train[]
 // tag::extractAnnotations[]
+    
+    @Override
+    public RecommendationEngineCapability getTrainingCapability() {
```
jcklie: The brace should be in the next line

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderFactory.java
```diff
@@ -51,7 +52,7 @@ public String getId()
     }
 
     @Override
-    public RecommendationEngine build(Recommender aRecommender) {
+    public RecommendationEngine build(Recommender aRecommender, RecommenderContext aContext) {
```
jcklie: The brace should be in the next line

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/ner/OpenNlpNerRecommenderFactory.java
```diff
@@ -51,7 +52,7 @@ public String getId()
     }
 
     @Override
-    public RecommendationEngine build(Recommender aRecommender) {
+    public RecommendationEngine build(Recommender aRecommender, RecommenderContext aContext) {
```
jcklie: The brace should be in the next line

inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/StringMatchingRecommender.java
```diff
@@ -74,13 +72,18 @@ public StringMatchingRecommender(Recommender aRecommender,
...
         if (aData != null) {
-            pretrainData = aData;
+            for (GazeteerEntry entry : aData) {
+                learn(dict, entry.text, entry.label);
+            }
         }
-        else {
-            pretrainData = emptyList();
+        
+        if (aContext != null) {
```
jcklie: Why can the context be null?

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/recommender/RecommendationEngine.java
```diff
@@ -85,14 +84,15 @@ public abstract EvaluationResult evaluate(List<CAS> aCasses, DataSplitter aDataS
 // end::methodDefinition[]
 
     /**
-     * Whether or not this engine supports training. If training is not supported, the call to
-     * {@link #train} should be skipped and {@link #predict} should be called immediately. Note
-     * that the engine cannot expect a model to be present in the {@link RecommenderContext} if
-     * training is skipped - this is meant only for engines that use pre-trained models.
+     * Which training capabilities this engine has.
```
jcklie: `Returns which training capabilities this engine has. `

inception-recommendation-api/src/main/java/de/tudarmstadt/ukp/inception/recommendation/api/recommender/RecommendationEngineFactory.java
```diff
@@ -46,7 +46,11 @@ default boolean isMultipleRecommendationProvider()
         return true;
     }
 
-    RecommendationEngine build(Recommender aRecommender);
+    default RecommendationEngine build(Recommender aRecommender) {
+    	return build(aRecommender, null);
```
jcklie: Can we use an empty context instead null here?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/tasks/TrainingTask.java
```diff
@@ -151,19 +153,33 @@ public void run()
...
 
+                    // If no data for training is available, but the engine requires training, 
+                    // do not mark as ready 
+                    if (cassesForTraining.isEmpty() && 
+                            capability == RecommendationEngineCapability.TRAINING_REQUIRED) {
+                    	log.info("[{}][{}]: There are no annotations available to train on",
+                                user.getUsername(), recommender.getName());
+                    	continue;
+                    }
+
```
jcklie: I would rewrite the if else a bit:

```
if (cassesForTraining.isEmpty()) {
// the two capability checks
} else {
  // mark as ready
}

```

inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/StringMatchingRecommender.java
munterkalmsteiner: Right, should not. It's not tested anywhere else, so removed it here too.

jcklie:
This looks ok now for me. @reckart WDYT?

reckart:

jcklie:

reckart:
@jcklie looks like we need to fix the test in master before we can merge this one...

inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -273,8 +272,12 @@ public EvaluationResult evaluate(List<CAS> aCasses, DataSplitter aDataSplitter)
     }
 
     @Override
-    public boolean requiresTraining()
+    public RecommendationEngineCapability getTrainingCapability() 
     {
-        return traits.isTrainable();
+        if (traits.isTrainable()) {
+            return RecommendationEngineCapability.TRAINING_REQUIRED;
```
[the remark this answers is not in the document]
munterkalmsteiner: Will this recommender be able to predict if it is not trained?

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommender.java
```diff
@@ -89,10 +90,18 @@ public void train(RecommenderContext aContext, List<CAS> aCasses)
         params.put(BeamSearch.BEAM_SIZE_PARAMETER, Integer.toString(beamSize));
         
         DoccatModel model = train(nameSamples, params);
-        if (model != null) {
-            aContext.put(KEY_MODEL, model);
-            aContext.markAsReadyForPrediction();
+        
+        if (model == null) {
```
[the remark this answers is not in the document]
munterkalmsteiner: Also here, will this recommender be able to predict if it is not trained?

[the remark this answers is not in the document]
reckart: > If a recommender needs to be trained (TRAINING_REQUIRED), but no model is available, I think throwing an exception makes sense. If a recommender works also without training (TRAINING_SUPPORTED), then failure to train a model with new data does not lead to an exception.

At prediction time yes - throwing an exception would make sense. But at training time, if there is not sufficient data to train a model, then I think throwing an exception overshoots a bit.

I have prepared an update for this PR with the following changes:

* RecommenderContext is now an AutoClosable and will *always* be closed by TrainingTask now
* the decision whether a context is ready for training is now made by the RecommendationEngine at prediction time in accordance with the training capabilities it declares
* no exceptions thrown if a model could not be trained unless there was an actual error during training
* isTrainable means TRAINING_SUPPORTED not TRAINING_REQUIRED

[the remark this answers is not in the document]
reckart: Also, my update locks down the RecommenderContext once it is closed so that now data can be added anymore. This presently causes exceptions to be thrown because of #1276.

reckart:

reckart:
> @jcklie looks like we need to fix the test in master before we can merge this one...

The failure in the majority recommender unit test was due to a network timeout - probably a transient one - not because of the outdated hashes in DKPro Core 1.10.0.

inception-imls-opennlp/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommender.java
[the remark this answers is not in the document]
munterkalmsteiner: Wow, ok. This looks good (and I learned something new)! This feature exploded a bit from my first commit...

reckart:

reckart:
When I implemented the hard closing of a RecommenderContext after training, I hadn't thought about the fact that we re-use the same context all the time. To make this cleaner, I have now changed the code such that every training creates a new RecommenderContext which is closed and published after the training is complete. To still provide support for incremental training, the new context is initialized with the data from the previous context. This in principle allows recommenders access to the model from the last training iteration. However, just updating the model object might not be a good idea since that may interfere with a concurrently running prediction task. Since we do not have any incrementally trained models atm, I'm inclined to leave it at this for now and consider introducing a different strategy when we actually add incrementally trained models and if we do consider it necessary at that point. After all, the recommender might just be smart e.g. by internally copying the model and thus leaving the previous (potentially concurrently used) model in tact. An alternative would be to move the `RecommenderContext.copy(...)` method into the RecommenderEngine API such that every recommender can decide if/which data from the previous context is copied over to the new one... or we could simply make two contexts available to the `train()` method, the old and the new one... chaining contexts would not be a good idea because that would be a memory leak ;)

reckart:

reckart:
@munterkalmsteiner @jcklie so I'm good now, but I made some more changes so you might care to comment before we finally merge thisl

reckart:
Considering the extent this PR has meanwhile taken, I think we should change the target branch to master...

reckart:

reckart:
Ok, so my testing looks good now and since there was no further feedback from your sides, I'm going to merge this now.

Issue #790, opened 2018-12-19
DB accesses during in-project search slow down process


reckart:
**Describe the refactoring action**
`MtasDocumentIndex.doQuery(IndexReader, SearchQueryRequest, String, MtasSpanQuery, List<String>)` makes DB queries for every matching document to check whether to use matches from the source document or from the annotation document.
This may heavily slow down search in larger projects. We have TWO DB accesses here, one for the source document and one checking whether there is an annotation document. This is quite bad. However, if we do not do this here, then the user will not get any results  from documents which the user has not opened yet or alternatively duplicate results. 

Better find a solution which handles this during indexing time - or possibly do a batch query for all document states before iterating over the query matches.

**Expected benefit**
Faster in-project query performance.

PR #1287, opened 2019-07-07
#790 - DB accesses during in-project search slow down process

MdeBoer95:
**What's in the PR**
* bugfix for #790 
* make a DB query for all SourceDocuments that have an AnnotationDocuments once before iteration over all the documents of a span
* change SearchQueryRequest to save the User instead of just the Username

**How to test manually**
* make queries in the search sidebar, see if it still works and if possible test if it is faster then before for large projects 

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Tested:
1) rebuild index and then search on an existing document in a project - found results
2) import new document into project, open the previously existing document (i.e. not the new document) and searching for a simple string - found results in both documents

Issue #1200, opened 2019-05-24
Selected Document Annotation not visible in details


wpmp:
**Is your feature request related to a problem? Please describe.**
When selecting a document annotation in the left sidebar while annotating a document, it is not clear to which document annotation the Details belong.

![doc_annotations](https://user-images.githubusercontent.com/5390724/58327943-2f165200-7e31-11e9-84ac-8be28dcbb707.png)

**Describe the solution you'd like**
The currently selected document annotation could be highlighted in some way.
For example highlight the name of the document annotation.
It would also be possible to show the document annotation name besides the `Details` text in the details box.

reckart:
Another alternative could be to display the document-level annotations in such a way that the "expand" when the user selects them.

PR #1288, opened 2019-07-08
#1200 - Selected Document Annotation not visible in details

wpmp:
**What's in the PR**
- document annotation side panel now works as an accordion instead of two separate panels
- the accordion only shows details for the currently selected annotation

**How to test manually**
* open a document in the brat or pdf editor
* create document annotations
* see if everything still works (create, select, delete, update values)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


wpmp:
> What is the purpose of the "selection cache"?

When filling a document annotation with data it seems the panel is reloaded, forgetting which annotation was selected and opened.
So when you fill out a value the annotation would close.

reckart:
Before this change, the details and the list were completely separate: one list, on detail panel. So when changes were made to the details, it was necessary/straightforward to simple reload the entire list in order to update the annotation label.

Now, every item in the list has its own private details, so this could probably be solved in a more elegant way. 

However, I would suggest we merge it as is after you added some comments to the places where the selection cache is being used explaining why it is there and what it does.

wpmp:
I think I found a better way than using this selectedCache variable.  
In the `detailPanels` list the existing detail panels are saved.  
Those are now reused, if present, when the rerendering is triggered.  
Otherwise a new detail panel is created.  
This is also necessary because previously with my last solution the list kept growing over time, never being cleared or cleaned.  
How do you like it now?  
Only thing I still don't like is passing the selection panel to the detail panels, but currently I don't see any way around it.

wpmp:
Oh and I noticed that now when creating a new annotation it does not open.
You have to open it manually.
Is this behaviour ok or should it be like before and be opened after it is created?

reckart:
I think it would be convenient for it to be opened when it is created.

wpmp:
Done

wpmp:
@reckart there seems to be an issue with the concept linking package:

```
[WARNING] Used undeclared dependencies found:
[WARNING]    org.dkpro.core:dkpro-core-api-ner-asl:jar:1.11.0:compile
[WARNING]    org.dkpro.core:dkpro-core-api-segmentation-asl:jar:1.11.0:compile
[WARNING]    de.tudarmstadt.ukp.dkpro.core:de.tudarmstadt.ukp.dkpro.core.api.segmentation-asl:jar:1.10.0:compile
[WARNING]    de.tudarmstadt.ukp.dkpro.core:de.tudarmstadt.ukp.dkpro.core.api.ner-asl:jar:1.10.0:test
```

At least I don't see any other reason for this build to fail

reckart:
I'm looking into the dependency issues...

Issue #1312, opened 2019-07-17
Embed Docker Compose script into documentation


reckart:
**Describe the refactoring action**
Embed the Docker Compose script into the Admin Guide proper.

**Expected benefit**
This allows use to auto-generate the proper version of the image into the script instead of having to use `latest` which apparently is not recommended.

Cf. https://github.com/webanno/webanno/pull/1362

PR #1313, opened 2019-07-17
#1312 - Embed Docker Compose script into documentation

reckart:
**What's in the PR**
- Embed script into admin guide and replace `latest` in script with the proper version

**How to test manually**
* Build and check that the Docker Compose script properly appears in the documentation

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-doc/pom.xml
+++ b/inception-doc/pom.xml
@@ -62,4 +62,5 @@
         </plugin>
       </plugins>
     </pluginManagement>
-  </build></project>
+  </build>
+</project>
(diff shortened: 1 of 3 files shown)


jcklie:
It looks ok and builds locally, though I would consider making the user and password also environment variables, as there are many places where you need to changes these.

reckart:
I'm not familiar with Docker Compose. Do you want to adapt the PR accordingly?

jcklie:
Done and started locally. I can log in to the db from the command line with the users once I set the credentials via export.

reckart:
@jcklie do you really think this is convenient? Is it a common practice? I'm slightly afraid that people might forget the credentials and then ask us how to recover...

jcklie:
You still have the option of replacing it in the compose file. I am used to do pass credentials in the environment for cloud deployments, like Travis or Heroku. You can also use a file you do not check in which sets these and source it.

jcklie:
See e.g https://docs.docker.com/compose/environment-variables/#the-env-file

reckart:
Roger. Thanks for the initiative and approval :)

Issue #918, opened 2019-02-04
Refactor dev documentation


jcklie:
**Describe the refactoring action**
A clear and concise description of what the action is.

- [ ] Introduction
    - Core technology (Java, Wicket, UIMA, DKPro, …)
- [x] Setting up the development environment
    - Source code management
        - Git
        - Github (Pull requests, branch naming, issues, projects)
        - Brunching (feature/bug, branching workflow, …)
    - IDE
        - Intellij/Eclipse import
        - Checkstyle
        - Maven config
    - Describe Project import (Maven) for both
- [x] Code style
- [ ] Architecture
    - Core services
        - Annotation
        - User
    - Events
    - Websockets
- [ ] Code constructs
    - Spring
    - Registries
- [x] Releasing
- [x] Remove database schema
- [x] Reduce external recommender toc length
- [x] Fancy javascript to autofold chapters
- [ ] Documentation on how to add extensions
    - [x] Recommender
    - [x] External recommenders
    - [ ] Menu items
    - [ ] Pages
    - [x] Layer types
    - [ ] Project initializers
    - [ ] Custom editors

reckart:
Currently no further refactorings planned.

PR #1315, opened 2019-07-18
#918 - Refactor dev documentation

jcklie:
**What's in the PR**
- Add intellij and git to setup documentation

**How to test manually**
* Build the documentation and read it

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-doc/src/main/resources/META-INF/asciidoc/docinfo.html
+++ b/inception-doc/src/main/resources/META-INF/asciidoc/docinfo.html
@@ -1,0 +1,75 @@
+<!-- Generate a nice TOC -->
+<script src="https://code.jquery.com/jquery-1.11.3.min.js"></script>
+<script src="https://code.jquery.com/ui/1.11.4/jquery-ui.min.js"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.tocify/1.9.0/javascripts/jquery.tocify.min.js"></script>
+<!-- We do not need the tocify CSS because the asciidoc CSS already provides most of what we neeed -->
+
+<style>
+  .tocify-header {
+    font-style: italic;
+  }
+
+  .tocify-subheader {
+    font-style: normal;
+    font-size: 90%;
+  }
+
+  .tocify ul {
+    margin: 0;
+  }
+
+  .tocify-focus {
+    color: #7a2518;
+    background-color: rgba(0, 0, 0, 0.1);
+  }
+
+  .tocify-focus > a {
+    color: #7a2518;
+  }
+</style>
+
+<script type="text/javascript">
+    $(function () {
+        // Add a new container for the tocify toc into the existing toc so we can re-use its
+        // styling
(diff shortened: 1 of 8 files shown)


inception-doc/src/main/resources/META-INF/asciidoc/docinfo.html
```diff
@@ -0,0 +1,75 @@
+<!-- Generate a nice TOC -->
+<script src="https://code.jquery.com/jquery-1.11.3.min.js"></script>
```
reckart: I'm all for a collapsing toc, but this should be done without having to resort to CDN hosts. The necessary scripts must be served by INCEpTION so it works even without internet connection.

jcklie: That is your code from stackoverflow.

reckart: I know :) And the reason I never integrated it into WebAnno or INCEpTION is that so far I was  too lazy to look into packaging the JS along the application. Wicket provides JQuery but I believe it generates a dynamic URL for it so we can't just hardcode that URL into the docs.

jcklie: We can use the webjar which we have as a dependency anyways and extract the js by maven, e.g. via [this magic](https://stff.net/unpack-static-resources-from-webjar-with-maven-for-js-testing/).

reckart: Webjars is a good idea. I adjusted build such that we use JQuery and tocify from the Webjar dependencies - but using Wicket magic instead of Maven magic :)

jcklie: If we copy the doc target folder to our website, do the paths work out then?

reckart: I don't understand the question. My last commit already takes care of loading the JS files from our webjar dependencies. That covers the issue, doesn't it?

jcklie: The jquery paths are related to wicket. We normally upload the docs to inceptio.github.io . I do not understand how that works, as there is no wicket there, only jekyll.

reckart: Got it. I forgot the website and considered only the case for the integrated help function in the application. I guess on the website, we could use the CDN (although we still might care to copy the JS to our website repo). We could add an ant task in inception-app-webapp/pom.xml line 602ff to replace the website-based URLs with Wicket-based URLs as the documentation is copied to its spot for being packaged as online help.

jcklie:
This can be merged once the build succeeds.

reckart:
ok, PR is merged, but I guess #918 is not resolved yet. So we move it off to 0.12.0?

Issue #1216, opened 2019-05-28
Linking span to document annotations


wpmp:
**Is your feature request related to a problem? Please describe.**
For peer review functionality review form elements will be represented by document annotations.
Hence a possibility is required to link a span annotation to a document annotation to represent the span belonging to a review form element.

**Describe the solution you'd like**
A solution would be to add a feature to the span annotation layer which can link to document annotations, similar to the option to link to span annotations.

wpmp:
Requires webanno/webanno#1378

PR #1329, opened 2019-07-26
#1216 - Linking span to document annotations

wpmp:
**What's in the PR**
- Adjust event inheritance for DocumentMetadata annotation events to match those of the other layer types
- Removed code which tried to channel feature states through the annotator state and instead rely on the feature state being passed along as part of arming a slot
- add listening for RenderSlotsEvent for QualifierFeatureEditor and SubjectObjectFeatureEditor

**How to test manually**
1. Create a new document annotation layer with a link feature to a span annotation layer.
2. Open a document
3. Create a document annotation with the type of your new layer
4. Select the new slot.
5. Annotate something in the text.
6. Link should be created

Also try out having a span layer with a link to another span.
Then have a document annotation and a span annotation both with links open and play around to see if the slots are armed/disarmed correctly.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


wpmp:
@reckart I managed to solve occurring exceptions when trying to create a link between a document and a span annotation, however it seems that the span and the link are not created and I don't see an exception.
Do you have any idea what the problem could be?

wpmp:
It seems that the selection in the annotatestate is not set.

reckart:
Should I have a look at this? How would I best investigate?

wpmp:
That would be nice.
But I could also show it to you on Tuesday if this is more convenient.
If you want to have a look earlier:
1. Create a new document annotation layer with a link feature to a span annotation layer.
2. Open a document
3. Create a document annotation with the type of your new layer
4. Select the new slot.
5. Annotate something in the text, however a span won't be created.
In `BratAnnotationEditor.java` in the `requestHandler` (beginning line 139) if you debug here and look into the `AnnotatorState` you will see that the annotation is not set (`getModelObject().getSelection().getAnnotation().isSet()`) which however is required for `AnnotationDetailEditorPanel.java` in `setSlot` line 1173 to create the new annotation.
Maybe the reason for this is, that once you select the slot not everything is correctly handled because it does not happen in the right side bar but in the left side bar?

reckart:

wpmp:
This won't compile without a new webanno release, or am I mistaken?
This requires webanno/webanno#1404

reckart:
The upstream issue/PR is already resolved/merged.

wpmp:
Yes but it is merged in master (3.6.0-SNAPSHOT) and INCEpTION curretnly depends on 3.6.0-beta-12

reckart:

Issue #1054, opened 2019-03-18
Dealing with many results in search sidebar 


reckart:
**Is your feature request related to a problem? Please describe.**
For queries with a large number of results, it takes very long to popular the sidebar and if becomes tedious to scroll in it.

**Describe the solution you'd like**
Here are some ideas:

- a hard limit of results to be shown in the sidebar (e.g. 1000 or 10000 -- maybe configurable by server admin)
- paging, e.g. showing 100 (proper paging, i.e. by re-issuing query when switching pages...)
- use a tree or somehow otherwise collapse the results such that first only documents are shown and the hits in side the documents appear when expanding a particular document

**Describe alternatives you've considered**
I put all alternatives that came to mind above.

MdeBoer95:
@reckart for the DataProvider that feeds the paging components in the UI I need the total number of results. I would do this by querying with mtas and subsequently iterate over the results just counting how many results we have and caching this value for future calls of the DataProvider's ```size()``` method. So basically an (almost) empty loop over the mtas results to find out the size. An alternative where we don't need the total number of results would be a scrollbar that loads new items when the user scrolls to the bottom. But I find the 'classic' paging i.e. something like ```< << 1 2 3 ... >> >``` is a much more convenient way to navigate. So I would sacrifice performance for a more convenient UI here. Let me know if you agree or disagree with my approach.

Regarding the annotation of search results:
Should we limit the selection of search results to annotate to the currently displayed page or should we keep track of all selections of all pages that the user has loaded before as well?

reckart:
I think having the empty loop which counts should be sufficient for the moment. So I agree with your opinion.

I'd tend towards not limiting to the current page. GitHub does e.g. limit selections in the issues list to the current page and I tend to find this annoying.

PR #1331, opened 2019-07-31
#1054 - Dealing with many results in search sidebar

MdeBoer95:
**What's in the PR**
- [x] integrate paging directly in the query
- [x] implement wicket Dataprovider and Dataview for search results
- [x] adjust SearchService and Sidebar to work with the paging
- [x] Hide page navigator if there are no search results yet 

When no grouping is activated (i.e. default grouping by document) and "low level paging" Checkbox is checked, paging will be applied at "query-level" (MtasDocumentIndex). 

Otherwise, all results are fetched at once and paging is applied for the rendering. 

This is done because if we apply paging at query level and grouping together, members of a
group might me scattered over multiple pages. (Grouping by document is an exception because the query iterates over documents. So results from the same document appear in a sequence)

**How to test manually**
* Make queries and annotations in the search sidebar, see if it still works. Selections of search results should persists when switching between different pages.
* Test queries with a large number of results, see if it is reasonably fast.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


MdeBoer95:
@reckart I wrapped this issue up. Can be reviewed.

reckart:
@MdeBoer95 Cool! I'm very curious :) Could you briefly summarizes how you implemented the feature?

MdeBoer95:
__Basically 3 Steps:__
1. Extend doQuery function in MtasDoumentIndex to take arguments _first_ and _count_. So that the function will only actually build the SearchResult Objects for all results in the given range (first, first + count).

2. Implement a Dataprovider``SearchResultsProvider`` whose ``iterator(long first, long count)`` - method calls doQuery via the SearchService interface. This data provider could be used with a Dataview in the UI already but there is the problem that grouping and paging together can have the effect that members of the same result-group appear on different pages (e.g. first 5 members on page 1 and next 5 members on page 3).

3. So as we discussed I implemented a DataProviderWrapper ``SearchResultsProviderWrapper``  that, per default, fetches all results at once and applies the paging only for rendering, not directly within the query. This way we can keep all members of the group together. 
If the option _low level paging_ (Checkbox in the UI)  is activated AND we are grouping by document the DataProviderWrapper just calls the DataProvider, so that paging will be applied within the query. This option is meant to be activated if the number of results for a query is very large. The drawback is, that we don't know in advance how many members a result group has in total. So the groupsize that is shown in the UI in the header of each group is only calculated for the current page(e.g. (7) ). Otherwise it shows something like (7/54).
_low level paging_ can only be activated if we are grouping by document because the doQuery method that produces the search results iterates over documents. So in this case, all results of a group (i.e. a document) will appear one after another without results of other groups (i.e. other documents) inbetween.

There is a cache for the pages inside the DataProvider. It might make more sense to maintain this cache in the DataProviderWrapper instead, but funcionallitywise it doesn't make much of a difference.
Right now a wicket model of the cache object is initialized in the SearchAnnotationSidebar and then passed to the DataProvider(Wrapper) which then fills the cache object. I was just thinking that the code maybe can be improved by utilizing a LoadableDetachableModel instead. 

__Tests__:
I tested this PR with some documents that contained the same word many times and then searched for this word. Without any paging (current Master) it loads basically forever. With paging only at "rendering-level" it loads the results in around 2 seconds. With low-level paging activted, results are loaded immediately. So from my tests it seems to have the desired effect.

Issue #1335, opened 2019-08-08
Recommendationservice performance


munterkalmsteiner:
**Describe the bug**
The performance of the recommendation service decreases considerably with a larger corpus. The methods `cloneCAS` and `monkeyPatchTypeSystem` seem to be the culprits (see screenshot).

**To Reproduce**
1. Import a relatively large dataset (mine had 1572 lines and 11350 words)
2. Create a recommender (e.g. POS with data majority)
3. Create an annotation.
4. Go to the next document
5. Prediction takes ~60 seconds

**Screenshots**
Shows runtime sampled with VisualVM, when switching to the next document in inception, i.e. `computePredictions` is invoked.
![image](https://user-images.githubusercontent.com/720692/62711448-5d033f80-b9f9-11e9-8f05-f9fa94dd9922.png)

**Please complete the following information:**
 - Version and build ID: 0.11.0-SNAPSHOT
 - OS: Linux
 - Browser: Firefox
 - CPU: Intel(R) Core(TM) i7-5600U CPU @ 2.60GHz

reckart:
The cloning and monkey-patching can most likely be done in a single step. That should save some time. Also, the `aTargetCas.getJCas()` in `cloneCas` can be dropped since we mostly use the CAS API these days - that will also save a lot of time. Although, the time saving will be removed if a recommender should try to use the JCas API later on. I am also not sure how the performance will be when we switch to UIMAv3.

munterkalmsteiner:
The performance issue is likely related to inefficiencies underlying in UIMA. When using the same amount of data, but in a single document (the above figures stem from ~400 documents), cloning and patching takes less than a second.

reckart:
The creation and initialization of a JCas container takes a lot of time - with UIMAv2 at least, using a simple CAS is much faster. The overhead of the creation of a (J)Cas object currently is paid twice, once during cloning and once during patching. So there is a lot of potential for performance improvement just by optimizing the way the UIMA framework is used in INCEpTION code.

reckart:
I have tried optimizing this locally and will create a PR later - would be great if you could have a look at it then.

munterkalmsteiner:
Sure, I'll benchmark it against the dataset I have.

reckart:
Here you go: https://github.com/inception-project/inception/pull/1337

munterkalmsteiner:
Prediction time has been improved:
[...]

PR #1346, opened 2019-08-18
#1335 - Recommendationservice performance

reckart:
**What's in the PR**
- Remove call to upgrade "original" CAS
- Changed TrainingTask "containsTargetAnnotation" to be able to deal with the case where the target type and/or feature are not defined in a CAS yet
- Changed RecommendationServiceImpl "calculateVisibility" to be able to deal with the case where the target type and/or feature are not defined in a CAS yet
- Added button to the recommender sidebar to explicitly trigger a retraining

**How to test manually**
* Assume a project with multiple documents
* Create a new span layer with a new string feature
* Define some recommender on that string feature (best a StringMatchingRecommender)
* Open the first document in the annotation page (meaning the first document is upgraded to the new type system, but the other documents are still on the old type system)
* Make some annotation using the span layer with the recommender
* See if predictions come up

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@munterkalmsteiner this should speed up the recommender still a bit more. Could you profile it and it would also be great if you could test it just in case I broke something.

munterkalmsteiner:
This is an improved over the previous patch tested [here](https://github.com/inception-project/inception/issues/1335#issuecomment-521707865):
image

I still get predictions.

Issue #1377, opened 2019-09-10
Do not show recommendations during curation


UWinch:
**Describe the bug**
If you annotate on the annotation page, get recommendations there and then switch to the curation page to curate, annotations will still be shown.

**To Reproduce**
1. Configure a project with a recommender.
2. Annotate until you get recommendations.
3. Switch to curation page.
4. See recommendations.

**Expected behavior**
Recommendations should not be shown on the curation page.

**Please complete the following information:**
 - Version and build ID: 0.12.0-SNAPSHOT
 - OS: ubuntu
 - Browser: chrome

PR #1390, opened 2019-09-17
#1388 Should not show recommendations for others

### LINKED ISSUE #1388: Should not show recommendations for others
**Describe the bug**
As an admin, it is possible to view other users' documents when opening documents on the annotation page. This should be read-only. However, when recommenders are enabled, it is possible to see and accept recommendations in others' documents thereby changing these documents.

**To Reproduce**
1. Configure recommenders for your project
2. Get some suggestions for a user
3. Log in as an admin user
4. Open a document of the other user
5. See and accept recommendations even though you should not be allowed to.

**Please complete the following information:**
 - Version and build ID: 0.11.2
 - Browser: Chrome

UWinch:
**What's in the PR**
If a user is viewing others' documents (as admin or for curation):
- disable recommendation sidebar forms
- do not show predictions
- do not trigger training

Closes #1388, #1377

**How to test manually**
For #1388 
* Configure recommenders for your project
* Get some suggestions for a user
* Log in as an admin user
* Open a document of the other user
* Make sure that no recommendations are shown and the recommendation sidebar is disabled

For #1377 
* Switch to curation page after annotating with recommendations
* Make sure that no recommendations are shown on the curation page

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
@@ -251,7 +251,7 @@ public class RecommendationEditorExtension
                        int aWindowBeginOffset, int aWindowEndOffset)
     {
         // do not show predictions during curation or when viewing others' work
-        if (aState.getMode().equals(Mode.CURATION) || 
+        if (!aState.getMode().equals(Mode.ANNOTATION) || 
                 !aState.getUser().equals(userRegistry.getCurrentUser())) {
             return;
         }
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
@@ -53,7 +53,9 @@ import de.tudarmstadt.ukp.clarin.webanno.brat.message.DoActionResponse;
 import de.tudarmstadt.ukp.clarin.webanno.brat.message.SpanAnnotationResponse;
 import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationFeature;
 import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationLayer;
+import de.tudarmstadt.ukp.clarin.webanno.model.Mode;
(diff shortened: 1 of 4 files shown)


reckart:
@jcklie could you have a look at this PR please?

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/RecommendationEditorExtension.java
```diff
@@ -245,6 +250,12 @@ private void actionRejectRecommendation(AnnotationActionHandler aActionHandler,
     public void render(CAS aCas, AnnotatorState aState, VDocument aVDoc,
                        int aWindowBeginOffset, int aWindowEndOffset)
     {
+        // do not show predictions during curation or when viewing others' work
+        if (aState.getMode().equals(Mode.CURATION) ||
```
jcklie: Rather than checking for not curation (blacklisting), I would check here whether it is annotation (whitelisting), so that the error does not appear in the next mode we add

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/sidebar/RecommendationSidebar.java
```diff
@@ -102,14 +107,29 @@ public RecommendationSidebar(String aId, IModel<AnnotatorState> aModel,
...
     }
 
     @Override
     protected void onConfigure()
     {
         // using onConfigure as last state in lifecycle to configure visibility
         super.onConfigure();
+        configureMismatched();
+        AnnotatorState state = getModelObject();
+        if (state.getMode().equals(Mode.CURATION)
```
jcklie: You could do 
`boolean enabled = state.getMode().equals(Mode.CURATION) || !state.getUser().equals(userRepository.getCurrentUser()))`
and then do

	     form.setEnabled(enabled);
        recommenderInfos.setEnabled(enabled);

Then it is not so convoluted to me. It would also be nice to have a comment here why we need that.

reckart: Since the sidebar is only shown on the annotation page and not on the curation page, do we need to check the Mode at all?

UWinch: thx, good ideas, done

Issue #1294, opened 2019-07-10
Add more feature config switches


reckart:
**Is your feature request related to a problem? Please describe.**
There are some features which an admin might care to disable to save on resources and others which are not quite up to standards yet, so more feature switches should be introduced.

**Describe the solution you'd like**
Add feature switches for:

- [x] recommender subsystem (default: on)
- [x] recommender evaluation functionality (default: on)
- [x] knowledge base subsystem (default: on)
- [x] external search subsystem (default: on) - alternatively disable subsystem if no search providers are enabled
- [x] ElasticSearch provider for external search
- [x] PubAnnotation provider for external search
- [x] internal search subsystem (default: on)

reckart:
Simply adding feature switches to disable things such as the PropertyFeatureSupport causes problems to existing projects which already have a Fact layer. There are two options

* change the code that generates the type system to skip unknown features instead of failing - this will drop all data for these features from the CASes while the features still show up in the project configuration
* make it such that e.g. disabling fact linking suppresses initializing new projects with the Fact layer and disables the ability to create new Property or SubjectObject features but still keep them working for existing projects

PR #1401, opened 2019-09-24
#1294 add more feature config switches

reckart:
**What's in the PR**
- Added switch "recommender.enabled"
- Added switch "recommender.evaluation-page.enabled"
- Added switch "external-search.enabled"
- Added switch "external-search.elastic-search.enabled"
- Added switch "external-search.pub-annotation.enabled"
- Added switch "knowledge-base.enabled"
- Added switch "knowledge-base.fact-linking.enabled"
- Added switch "knowledge-base.entity-linking.enabled"
- Added switch "search.enabled"
- Allow reification only on wikidata-type KBs (remote ones)

**How to test manually**
* Enable/disable the switches in different combinations and see if the application still works as expected

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


reckart:
@jcklie this is ready for review.

reckart:
@jcklie ping?

jcklie:
When I create an elastic repository and then set

```
external-search.pub-annotation.enabled=false
external-search.elastic-search.enabled=false
```

then it still shows up in the search list.

jcklie:
When I create a KB and then disable it and then try to annotate NER, then I get
`Error: Unsupported feature: [identifier]`

jcklie:
`knowledge-base.entity-linking.enabled` also enables the KB page in the dashboard, is this intended?

reckart:
> When I create an elastic repository and then set...

I think when disabling a feature, then creating new stuff of that feature should no longer work - but I didn't want to cripple existing projects.

reckart:
> `knowledge-base.entity-linking.enabled` also enables the KB page in the dashboard, is this intended?

Not sure what you mean. If I use these settings, then the KB page is not there:

```
knowledge-base.enabled=false
knowledge-base.entity-linking.enabled=true
```

If I use these settings, the KB page is there:

```
knowledge-base.enabled=true
knowledge-base.entity-linking.enabled=false
```

Seems ok to me. Anything unexpected for you? @jcklie

jcklie:
> > `knowledge-base.entity-linking.enabled` also enables the KB page in the dashboard, is this intended?
> 
> Not sure what you mean. If I use these settings, then the KB page is not there:
> 
> ```
> knowledge-base.enabled=false
> knowledge-base.entity-linking.enabled=true
> ```
> 
> If I use these settings, the KB page is there:
> 
> ```
> knowledge-base.enabled=true
> knowledge-base.entity-linking.enabled=false
> ```
> 
> Seems ok to me. Anything unexpected for you? @jcklie

My bad, it works as expected.

reckart:
> When I create a KB and then disable it and then try to annotate NER, then I get `Error: Unsupported feature: [identifier]`

This one seems not to be a small fix... making the application resilient against ripping out a `LayerSupport` or `FeatureSupport` from underneath a project which contains layers/features using these supports seems to be quite a bit of effort...

jcklie:
Is it possible to feature gate internal search? Is it planned?

reckart:
Theoretically yes - but there is no issue/roadmap for it. Btw. I'm looking into the indexing process right now (#1533) - so please don't start refactoring the indexing code.

jcklie:
I thought the issue for feature gating stuff is this issue/PR or I do not understand why we did not add it to the todo.

reckart:
This issue is about enabling/disabling features globally - not on a per-project basis.

reckart:
Sorry, I think I mis-understood you. There is a feature switch for internal search in the PR.

jcklie:
It was not on the to-test list, so I asked.

reckart:
The switch is called `search.enabled` - I added it above.

reckart:
> When I create a KB and then disable it and then try to annotate NER, then I get Error: Unsupported feature: [identifier]

Not sure what to do about this one. The problem here is that if KB is enabled and a project is created, then the KB feature `identifier` is created (if a project is created when KB is not active, this does not happen). 

If we disabled KB afterwards, then the `FeatureSupport` for the identifier features is no longer there. This has a number of effects including:

* type system information for KB features can no longer be generated - this causes KB features to be purged from annotation documents when their type system is upgraded
* it also causes the system to no longer be able to recognize the KB feature definitions and the system cannot provide feature editors for them

 I added a note to the documentation that disabling the KB should be done before creating any projects using KB features, but I don't have a good idea at the moment of how to fix this in any better way.

So the question is: merge?

jcklie:
I tested the `search.enabled` and it works . From my side, you can merge it after fixing build errors

reckart:
The reason that our builds fail atm is this: https://github.com/dkpro/dkpro-core/issues/1469

I'm trying to add a workaround in INCEpTION so we do not have to wait for a new DKPro Core release (or for CC to fix their website - I have pinged them about it as well...).

PR #1425, opened 2019-10-02
Bugfix/#1422 unused local variables and functions should be removed

JaviValera:
I have deleted some unnecessary variables and functions in JavaScript, almost everything in the PDF module.

### DIFF
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/chart/resources/chart.js
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/chart/resources/chart.js
@@ -25,7 +25,6 @@
 
 function updateLearningCurveDiagram(arrayOfLearningCurves) {
   var xAxixType = 'indexed';
-  var plotType = 'step';
   
   var xTick = {
     format : function(a) {


JaviValera:
@munterkalmsteiner

munterkalmsteiner:
@JaviValera thanks for the PR! There are a couple of issues, though:

1. most of the refactorings are on code that is either generated (e.g. annotator-full.js) or is third party code that was re-used as is in INCEpTION (the pdf annotation code). For the generated code, it would be better to solve the root cause of the issue, but that is also a 3rd party package.
2. There are a couple of commits in the PR that are irrelevant for the issue itself (reverts, deletion of sonar configuration). This makes it difficult to assess what happened in this patch.

I think the refactoring in chart.js is ok. As git reset won't work on an already publicly shared branch, I'd suggest you to close this PR and create a new one that does only contain the intended refactorings.

JaviValera:
@munterkalmsteiner Updated, deleted all the useless commits

munterkalmsteiner:
@JaviValera what about my first comment regarding generated or 3rd party code? I'd rather not touch code that is not under our control.

JaviValera:
I'm not sure which parts are the generated or 3rd party code, should I only let the change in the chart.js?

reckart:
Every module that has third-party code should include a NOTICE.txt file which explains which parts are third-party code.

reckart:
Also, the license headers in the files should indicate which files are part of INCEpTION proper any which come from other sources.

reckart:
Looks ok, just the commit message doesn't conform to our conventions. I'll squash-merge it.

reckart:

PR #1428, opened 2019-10-04
Bugfix/1427 string literals should not be duplicated

JaviValera:
Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences. On the other hand, constants can be referenced from many places, but only need to be updated in a single place.


inception-imls-opennlp/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/opennlp/doccat/OpenNlpDoccatRecommenderTest.java
```diff
@@ -82,8 +83,7 @@ public void setUp()
     }
 
     @Test
-    public void thatTrainingWorks() throws Exception
-    {
+    public void thatTrainingWorks() throws IOException, UIMAException, RecommendationException {
```
reckart: Please avoid changes not relevant to the issue. For new code or modified code, please use the DKPro Core code style. An Eclipse formatter configuration file is attached here: https://dkpro.github.io/contributing/

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -153,16 +154,16 @@ private void createProject(Project aProject) throws Exception
...
+    private final void uploadDocument(Pair<SourceDocument, String>... aDocuments) throws IOException {
         Project project = null;
         for (Pair<SourceDocument, String> doc : aDocuments) {
             project = doc.getLeft().getProject();
             
             try (InputStream fileStream = new ByteArrayInputStream(
                     doc.getRight().getBytes(UTF_8))) {
                 documentService.uploadSourceDocument(fileStream, doc.getLeft());
+            } catch (UIMAException e) {
+                e.printStackTrace();
```
reckart: Printing a stack trace is not a good way to handle an exception. Rather throw it.

PR #1429, opened 2019-10-04
Bugfix/1427 string literals should not be duplicated

JaviValera:
Duplicated string literals make the process of refactoring error-prone, since you must be sure to update all occurrences. On the other hand, constants can be referenced from many places, but only need to be updated in a single place.

### DIFF
--- a/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java
+++ b/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java
@@ -249,7 +249,7 @@ public class KnowledgeBaseServiceImplWikiDataIntegrationTest  {
         PROFILES = KnowledgeBaseProfile.readKnowledgeBaseProfiles();
         KnowledgeBase kb_wikidata_direct = new KnowledgeBase();
         kb_wikidata_direct.setProject(project);
-        kb_wikidata_direct.setName("Wikidata (official/direct mapping)");
+        kb_wikidata_direct.setName(KB_NAME);
         kb_wikidata_direct.setType(PROFILES.get("wikidata").getType());
         kb_wikidata_direct.applyMapping(PROFILES.get("wikidata").getMapping());
         kb_wikidata_direct.applyRootConcepts(PROFILES.get("wikidata"));
--- a/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
+++ b/inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
@@ -138,7 +138,7 @@ public class RecommenderEditorPanel
         form.add(nameField);
         
         autoGenerateNameCheckBox = new CheckBox(MID_AUTO_GENERATED_NAME,
-                PropertyModel.of(this, "autoGenerateName"));
+                PropertyModel.of(this, MID_AUTO_GENERATED_NAME));
         autoGenerateNameCheckBox.add(new LambdaAjaxFormComponentUpdatingBehavior("change", t -> {
             autoUpdateName(t, nameField, recommenderModel.getObject());
             t.add(autoGenerateNameCheckBox);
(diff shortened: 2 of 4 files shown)


JaviValera:
@munterkalmsteiner

JaviValera:
updated deleting the useless commits, thanks for the tutorial @munterkalmsteiner

munterkalmsteiner:
@JaviValera looks good! Thanks!

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/AbstractInfoPanel.java
```diff
@@ -171,7 +171,7 @@ public ViewMode(String id, CompoundPropertyModel<? extends KBObject> compoundMod
             Label identifier = new Label("idtext"); 
             TooltipBehavior tip = new TooltipBehavior();
             tip.setOption("autoHide", false);
-            tip.setOption("content",
+            tip.setOption(CONTENT_MARKUP_ID,
```
reckart: The `CONTENT_MARKUP_ID` is used to identify a Wicket component in the HTML code. However, the "content" literal replaced here does refer to the option `content` that is set on the tooltip - the two have nothing to do with each other. The constant `CONTENT_MARKUP_ID` cannot be used here.

inception-recommendation/src/test/java/de/tudarmstadt/ukp/inception/recommendation/service/VisibilityCalculationTests.java
```diff
@@ -88,7 +88,7 @@ public void setUp() throws Exception
         project.setMode(WebAnnoConst.PROJECT_TYPE_ANNOTATION);
 
         List<AnnotationFeature> featureList = new ArrayList<AnnotationFeature>();
-        featureList.add(new AnnotationFeature("value", "uima.cas.String"));
+        featureList.add(new AnnotationFeature(FEATURE, "uima.cas.String"));
```
reckart: Looks ok.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/project/RecommenderEditorPanel.java
```diff
@@ -138,7 +138,7 @@ public RecommenderEditorPanel(String aId, IModel<Project> aProject,
         form.add(nameField);
         
         autoGenerateNameCheckBox = new CheckBox(MID_AUTO_GENERATED_NAME,
-                PropertyModel.of(this, "autoGenerateName"));
+                PropertyModel.of(this, MID_AUTO_GENERATED_NAME));
```
reckart: The MID is the identifier that binds the Wicket CheckBox to its HTML element defined in the respective `.html` file. However, the`autoGenerateName` literal used in the `PropertyModel` refers to the name of a field defined in the RecommenderEditorPanel (`private boolean autoGenerateName;`). The HTML element ID/markup ID  be changed independently from the field name. Java 8 introduced method references, but unfortunately (afaik), there are no field references (yet). So the PropertyModel works using reflection. I think using the MID here in place of the field name is not sensible.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java
```diff
@@ -249,7 +249,7 @@ private KnowledgeBase buildKnowledgeBase(Project project, String name) throws IO
         PROFILES = KnowledgeBaseProfile.readKnowledgeBaseProfiles();
         KnowledgeBase kb_wikidata_direct = new KnowledgeBase();
         kb_wikidata_direct.setProject(project);
-        kb_wikidata_direct.setName("Wikidata (official/direct mapping)");
+        kb_wikidata_direct.setName(KB_NAME);
```
reckart: Looks reasonable.

inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/AbstractInfoPanel.java
munterkalmsteiner: @JaviValera if you address reckart's comment above, the other two changes can be merged.

PR #1440, opened 2019-10-06
Refactor/issue 1433 rename class and filed

Taaaaaserface:
**What's in the PR**
* Refactor/issue 1433 rename class and filed

**How to test manually**
* Rename the classes and fileds, no need to test

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


inception-app-webapp/src/main/java/de/tudarmstadt/ukp/inception/INCEpTION.java
```diff
@@ -72,138 +72,120 @@
...
-                CurationPageMenuItem.class,
-                MonitoringPageMenuItem.class,
-                AgreementPageMenuItem.class,
-
-                // INCEpTION uses its recommenders, not the WebAnno automation code
-                AutomationService.class, 
-                AutomationMiraTemplateExporter.class,
-                AutomationTrainingDocumentExporter.class
-        })})
+@ComponentScan(basePackages = { "de.tudarmstadt.ukp.inception",
```
munterkalmsteiner: Why do you make these code formatting changes?

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/ranking/BaselineRankingStrategy.java
```diff
@@ -34,7 +34,7 @@
 
 public class BaselineRankingStrategy
 {
-    private static Comparator<CandidateEntity> INSTANCE = (e1, e2) -> new CompareToBuilder()
+    private static Comparator<CandidateEntity> instance = (e1, e2) -> new CompareToBuilder()
```
munterkalmsteiner: Why is lower case better than upper case?

inception-imls-stringmatch/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/stringmatch/trie/Trie.java
```diff
@@ -36,7 +36,7 @@
 public class Trie<V>
 // implements Map<CharSequence, V>
 {
-    private int _size = 0;
+    private int s_size = 0;
```
munterkalmsteiner: Why is the new name better than the old?

reckart: FYI: I wrote this code a long time ago in the context of a different project. At that time, I used the convention that fields start with an underscore which would make it easy to distinguish fields from parameters and variables without having to prefix them with `this.`. We do not use this convention in INCEpTION though, but I didn't bother to update the file. According to the conventions we use these days, the underscore would simply be removed.

reckart:
> I think INSTANCE here is a  local variables rather than constant names(maybe not).

INSTANCE is a statically initialized singleton. Adding `final` to it and making it a proper constant seems to work and would in my opinion be the proper change in this case.

Taaaaaserface:
OKAY!

Richard Eckart de Castilho <notifications@github.com> 于2019年10月10日周四
上午11:58写道：

munterkalmsteiner:
@Taaaaaserface I would suggest you to proceed as follows:

1. Delete the two commits (revert is not enough as it pollutes the commit history). See [here](https://stackoverflow.com/questions/1338728/delete-commits-from-a-branch-in-git) how to do it. Read also carefully the consequences of pushing modified history when someone else (in your team for example) has pulled your branch.
2. Make sure you create one commit for each logically related fix. E.g. if you remove leading underscores in variable names, do that in one commit, for all files, and nothing else besides that change. That helps us to review and also allows you to fix a pull request in case we would accept a pull request, except a particular commit. Then it is easy to delete that commit and get the pull request accepted.
3. Don't refactor uppercase into lowercase in variables. Fix however what @reckart mentioned [here](https://github.com/inception-project/inception/pull/1440#issuecomment-540495428).
4. Refactor the variable names with leading underscore by removing the underscore
5. Make sure not to include any unwanted formatting changes.

Thanks!

Taaaaaserface:
Failed to delete commit, will creat a new pull request.

PR #1441, opened 2019-10-06
Bugfix/#1426 generic exceptions should never be thrown

SergioFQ:
Changed the generic exceptions for specific ones to show the true exceptions


SergioFQ:
@munterkalmsteiner

inception-example-imls-data-majority/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/datamajority/DataMajorityNerRecommenderTest.java
```diff
@@ -204,11 +210,16 @@ public void thatIncrementalNerEvaluationWorks() throws Exception
...
+           
+            try {
+            	 assertThat(score).isStrictlyBetween(0.0, 1.0);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+            }
             i++;
         }
-    }
```
munterkalmsteiner: This `}` should stay, the one on line 219 is too much.

inception-imls-dl4j/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/dl4j/pos/DL4JSequenceRecommenderTest.java
```diff
@@ -347,7 +348,13 @@ public void thatNerPredictionWorks() throws Exception
         addScoreFeature(cas, NamedEntity.class, "value");
         sut.predict(context, cas);
 
-        List<NamedEntity> predictions = getPredictions(cas, NamedEntity.class);
+        List<NamedEntity> predictions = null;
+		try {
+			predictions = getPredictions(cas, NamedEntity.class);
+		} catch (Exception e) {
+			// TODO Auto-generated catch block
```
munterkalmsteiner: Remove generated comment.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -153,8 +148,8 @@ private void createProject(Project aProject) throws Exception
     }
 
     @SafeVarargs
-    private final void uploadDocument(Pair<SourceDocument, String>... aDocuments)
-        throws Exception
+    private final void uploadDocument(Pair<SourceDocument, String>... aDocuments) throws IOException, UIMAException
+    //throws Exception
```
munterkalmsteiner: remove comment

reckart:
> Overall, looks good, thanks! As almost all changes affect test code, I'm not sure if throwing generic exceptions was a design choice or not (@reckart maybe can shed some light on this).

@munterkalmsteiner in test code, I usually just throw exception (in particular in the root test methods) since test code is usually not expected to handle exceptions anyway. If there is an exception, then it should cause the test to fail. I feel it is just cumbersome in test code to have a (potentially huge) list of exceptions on test methods.

inception-example-imls-data-majority/src/test/java/de/tudarmstadt/ukp/inception/recommendation/imls/datamajority/DataMajorityNerRecommenderTest.java
```diff
@@ -98,7 +99,12 @@ public void thatPredictionWorks() throws Exception
 
         sut.predict(context, cas);
 
-        Collection<NamedEntity> predictions = getPredictions(cas, NamedEntity.class);
+        Collection<NamedEntity> predictions = null;
+        try {
+            predictions = getPredictions(cas, NamedEntity.class);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
reckart: Catching exception and wrapping it in a runtime exception doesn't really contribute anything here. It just means that Exception doesn't appear on the `throws` statement. It's not a proper handling. I prefer just having `throws Exception` and that's it.

reckart: This applies to other changes in this PR likewise.

inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -479,7 +482,7 @@ public TestContext()
                 FileUtils.deleteDirectory(new File(temporaryFolderPath));
             }
             catch (IOException e) {
-                e.printStackTrace();
+                throw new RuntimeException(e);
```
reckart: Either a failed deletion should be ignored and logged (as before) or it would be sufficient to just let it thrown the IOException. Throwing the exception is probably the better choice.

reckart:
@munterkalmsteiner in particular if "managing" the sizes of the `throws` list results in catching exceptions and wrapping them as unchecked exceptions (e.g. `RuntimeException`), I surely prefer just having `throws Exception`. Keeps the code short and wrapping checked exceptions in unchecked exceptions IMHO isn't a good way of exception handling anyway - at least in the general case.

munterkalmsteiner:
> @munterkalmsteiner in test code, I usually just throw exception (in particular in the root test methods) since test code is usually not expected to handle exceptions anyway. If there is an exception, then it should cause the test to fail. I feel it is just cumbersome in test code to have a (potentially huge) list of exceptions on test methods.

@SergioFQ, this is what I suspected... if there any instances of this improvement **not** affecting test code, you are welcome to update this pull request (getting rid of the older commits to not pollute the commit history) or submit a new pull request if cleaning up this one is too much of a hassle.

reckart:
In a production environment using Sonarcube, I would try to set up a rule that excludes test code from the "throws Exception" check.

reckart:
@munterkalmsteiner Is there a way to drop some config file into the repo that Sonarcube would pick up? Maybe an outcome of some PRs could be to set up such a file instead of addressing the suggestions from Sonarcube.

munterkalmsteiner:
> @munterkalmsteiner Is there a way to drop some config file into the repo that Sonarcube would pick up? Maybe an outcome of some PRs could be to set up such a file instead of addressing the suggestions from Sonarcube.

@reckart good idea! @SergioFQ could you look into that? [Tip](https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/).
It would be great if you could setup INCEpTION with sonarcloud (@reckart, did I understand you correctly?) and try to reduce the number of false positives by excluding the exception check in test code.

reckart:
I haven't used Sonarcloud yet, but sounds like a useful idea :)

PR #1449, opened 2019-10-06
Refactor/#1446-Make the enclosing method static or remove this set.

jianglinyao1996:
**What's in the PR**
* Make the enclosing method static or remove this set.

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java
+++ b/inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java
@@ -84,7 +84,7 @@ public class KnowledgeBaseServiceImplWikiDataIntegrationTest  {
     private KnowledgeBaseServiceImpl sut;
     private Project project;
     private KnowledgeBase kb;
-    private Reification reification;
+    private static Reification reification;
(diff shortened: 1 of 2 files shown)


jianglinyao1996:
I think it need to make the following method static, so that across all class instances there is only one copy of that method making updates to the static (i.e. shared across all class instances) field.
File:(1)inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseSubPropertyLabelTest.java Line:196
(2)inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseSubPropertyLabelTest.java Line:212
(3)D:inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/KnowledgeBaseServiceImplWikiDataIntegrationTest.java Line:249

reckart:
Static fields should normally not be mutable.

jianglinyao1996:
So these enclosing methods should not make static.

munterkalmsteiner:
@jianglinyao1996 I still do not understand the theoretical benefit of this refactoring. In practice, there are arguments against static mutable fields. See also the discussion [here](http://code.joshpollak.com/2011/06/03/java-static-variables-are-wrong-almost-always/). Could you provide some arguments for your proposed change that show the benefits of your suggestion?

PR #1450, opened 2019-10-06
Refactor/issue 1447 Remove the unused function in JavaScript

Taaaaaserface:
**What's in the PR**
Remove the unused function in JavaScript* ...

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


munterkalmsteiner:
This change affects only 3rd party code (see license text in the header of the file). I think it is not a good idea to modify code that we don't control. A better approach would be to create a patch upstream.

PR #1458, opened 2019-10-08
We made a lot of modifications to variables or member methods in diff…

Jackwulujie:
…erent classes

**What's in the PR**
We made a lot of modifications to variables or member methods in different classes.

**How to test manually**
We test by running the application.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I am sorry, but I don't understand why these variables were renamed. Could you explain?

Jackwulujie:
When we used sonarqube to test the Inception application,
there would be a lot of duplicated blocks of code, which included these variables and member methods. 
These renamed variables themselves do not affect the use of the application,
we tried to remove some of them. But most of them are extended from parent classes 
or implemented from interfaces or  be called in there classes, so we couldn't just remove them which would
affect the use of the applicaiton. Therefore, in order to enable developers to better distinguish between these 
variables or methods with the same names from different classes, we renamed them. The code smell we need to 
solve is to remove duplicated blocks of code. We can't remove them all, so some parts of them were renamed by us.





At 2019-10-10 02:24:42, "Richard Eckart de Castilho" <notifications@github.com> wrote:


I am sorry, but I don't understand why these variables were renamed. Could you explain?

reckart:
Are you saying that you renamed the variables to prevent sonarcube from detecting code as being repeated?

For example, you renamed several `actionXXX` methods to `actionXXX1`. These methods are local to the respective classes. What kind of risk of developers being unable to distinguish them is there?

Jackwulujie:
We are responsible for solving this problem detected by sonarqube:
Duplicated blocks of code must be removed.
We have discussed with our professor Deepika about this issue.
Our group members all thought it couldn't be an issue at all, 
because we thought the codes didn't have problem(there is no need
to modify the code, the current codes or structures are good), we didn't know 
how to modify them to satisfy the requirement. We have tried many methods, 
like removing repeated codes, creating new parent classes or interfaces, 
but these ways could only make the codes of application more complex. 
So we consulted our professor Deepika for advice, she told us that we 
could just renamed the names of the variables or member methods in classes. 
Then we made these modifications. The develops can easily find the modules
that they are responsible for according to the naming method. And if there is something 
wrong, maintenance personnels can find faulty components faster.

reckart:
@munterkalmsteiner Why do you think?

reckart:
To me it looks like Sonarcube is a bit overzealous here and the renaming of the variables and fields feels to me like a hack which causes sonarcube to no longer detect the (maybe intentional/unavoidable) code duplication. Maybe it is a matter of settings some duplication detection thresholds in Sonarcube? Or am I and @Jackwulujie really missing to see the code smell here since - if I understand @Jackwulujie right - we both think the code is good?

munterkalmsteiner:
@Jackwulujie I agree with you and @reckart that the reported issues are false positives. I wonder however what sonarqube actually detects here. Can you post an example of the exact finding instance and description from sonarqube?

In any case, the automatically changed method names (adding a numeric suffix) does not improve the readability of the code at all... when there is a method with a number n in the name, I always wonder where is n-1 or n+1. Really distracting...

Jackwulujie:
> @Jackwulujie I agree with you and @reckart that the reported issues are false positives. I wonder however what sonarqube actually detects here. Can you post an example of the exact finding instance and description from sonarqube?
> 
> In any case, the automatically changed method names (adding a numeric suffix) does not improve the readability of the code at all... when there is a method with a number n in the name, I always wonder where is n-1 or n+1. Really distracting...


image
image
image

For example, here are three classes that implement the same interface named "FormatSupport". In fact, sonarqube detected that far more than these three classes inherited the interface. These three classes have the same memeber methods and variables, like getID(), ID, getName(),NAME. But since these member methods and variables implement are all implemented from the same interface, they don't need to be modified or removed(these methods are called in other classes). At the same time, we tried to modify the interface to solve this problem, but found that the interface could not be modified at all, and we thought that doing these modifications would only make the application code more complex and more difficult to understand.

munterkalmsteiner:
> For example, here are three classes that implement the same interface named "FormatSupport". In fact, sonarqube detected that far more than these three classes inherited the interface. These three classes have the same memeber methods and variables, like getID(), ID, getName(),NAME. But since these member methods and variables implement are all implemented from the same interface, they don't need to be modified or removed(these methods are called in other classes). At the same time, we tried to modify the interface to solve this problem, but found that the interface could not be modified at all, and we thought that doing these modifications would only make the application code more complex and more difficult to understand.

Ok, this particular smell makes actually sense to fix:

1. Create an abstract base class `FormatSupportBase` that implements the interface `FormatSupport`. The abstract class should implement the common code.
2. Then, let `LifFormatSupport` etc. extend the base class. That should remove some of the duplicated blocks (not the ones for isReadable/isWritable).

What I find strange is that sonarqube detects this as a smell at all. The [default settings](https://docs.sonarqube.org/latest/user-guide/metric-definitions/#header-3) require 10 consecutive lines for a code clone.

For all the other refactorings, they need to be reasoned on a case by case basis. It does not make always sense. You should review the changes suggest by sonarqube and group them logically into separate commits. Do not blindly autorefactor the code, but think what makes sense. 

Delete the two commits (revert is not enough as it pollutes the commit history). See [here](https://stackoverflow.com/questions/1338728/delete-commits-from-a-branch-in-git) how to do it. Read also carefully the consequences of pushing modified history when someone else (in your team for example) has pulled your branch.

reckart:
I don't think the abstract base class will help much. The defaults are already present as default methods in the FormatSupport interface.

Jackwulujie:
The problem we are responsible for is removing duplicate blocks of code.

Indeed, sonarqube helps us detect many duplicate code blocks in this application, but after our detailed discussion, analysis and actual modification, we don't think that these code must be modified. The detected code block duplication problem can be roughly divided into two types: one is the code duplication caused by the same name of methods or variables implemented in the same interface in many classes; the other is the duplication caused by multiple calls to the same variable or member method in the same class. We have tried many methods for this, such as deleting duplicate code, creating new parent classes or interfaces, but these methods will only make the application code more complex and difficult to understand. After that, we renamed these repeated variables, but we still think that the irregular naming standard (adding prefix or suffix to variables and methods) will not improve the readability and comprehensibility of the code itself, and it will not get the support and recognition of the developers, just for the sake of not being detected by sonar, so we and the developers finally recognized it. These changes are meaningless, do not make the code itself better understood, and do not help developers distinguish variables or member methods with the same name from different classes. If the developer or maintainer wants to identify which class the variable or method comes from, just hold down Ctrl and click the left mouse button to jump to the corresponding class, which is much easier than simply renaming or adding an abstract base class. Methods or variables in a class implemented from the same interface do not need to be modified, and their names should be the same as those in the interface. Multiple calls to a variable or method in the same class will not affect the normal use of the method or variable with the same name in another class, because they are all local and can only be called by the object entity of the class, so we think their repetition is inevitable and does not need to be changed. The two changes we submitted before will not greatly help the application code itself (developer also thinks so), and will not greatly improve the readability and understandability of the code. Therefore, we decided not to make changes to these repeated code blocks and close the pull request.

PR #1470, opened 2019-10-18
Refactor/1434 remove useless braces

Taaaaaserface:
**What's in the PR**
* I remove some useless braces around the statement.


Taaaaaserface:
Remain other changes and commits

PR #1472, opened 2019-10-18
Refactor/#1434 Remove the useless curly braces

Taaaaaserface:
I remove the useless curly braces around the statement to make the code more readability and easy to read for developers


Taaaaaserface:
There are some other changes except removing useless curly braces
in:
inception-ui-kb/src/main/java/de/tudarmstadt/ukp/inception/ui/kb/project/AccessSpecificSettingsPanel.java
from line 185:

IModel<String> adapter = new LambdaModelAdapter<String>(() -> {
            return model.getObject() != null ? model.getObject().stringValue() : null;
        }, str ->

I should remove both useless curly braces and delete the semicolon and return like this

IModel<String> adapter = new LambdaModelAdapter<String>(() -> 
             model.getObject() != null ? model.getObject().stringValue() : null
        , str ->

Can the **return** be removed here?

reckart:
The curlies might be a "hack" to let the Eclipse formatter format the code in a way compatible with the checkstyle rules - or to make the code "more readable" my making more explicit that there ware actually two embedded code blocks in the LamdaModelAdapter constructor ... but general, I don't see why they would be necessary.

munterkalmsteiner:
> Can the **return** be removed here?
@Taaaaaserface, yes (see [Section 8.4](https://www.baeldung.com/java-8-lambda-expressions-tips)). Do you get compilation errors otherwise?

Taaaaaserface:
> > Can the **return** be removed here?
> > @Taaaaaserface, yes (see [Section 8.4](https://www.baeldung.com/java-8-lambda-expressions-tips)). Do you get compilation errors otherwise?

There is no compilation errors after modifing.

reckart:

reckart:
Looks good - let's see what Jenkins thinks.

reckart:

reckart:

reckart:
Jenkins complains about more checkstyle issues - now unused imports in the "Task" class. 

Please do run `mvn clean install` on the command line - it should generate the same checkstyle issues that Jenkins is reporting.

munterkalmsteiner:
@Taaaaaserface please check reckart's [comment](https://github.com/inception-project/inception/pull/1472#issuecomment-548271839).

Taaaaaserface:
> Jenkins complains about more checkstyle issues - now unused imports in the "Task" class.
> 
> Please do run `mvn clean install` on the command line - it should generate the same checkstyle issues that Jenkins is reporting.

Is this means I should delete all the imports as following:
import static org.apache.commons.lang3.Validate.notNull;
import org.apache.uima.cas.CAS;
import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationDocument;
import de.tudarmstadt.ukp.clarin.webanno.model.Project;
import de.tudarmstadt.ukp.clarin.webanno.model.SourceDocument;
in Task.java class and run `mvn clean install` on the command line?
*By the way, can I ask where is the commend line in eclipse?Is that means in git bash?

reckart:
Sounds like you might be on Windows. In that case, best look how to install Maven on Windows so that you can use it from the command line. [This article](https://www.mkyong.com/maven/how-to-install-maven-in-windows/) might help.

reckart:
Removing unused imports in Eclipse is usually a matter of using the ["Organize imports"](https://www.mkyong.com/intellij/eclipse-ctrl-shift-o-in-intellij-idea/) functionality of Eclipse. Do not apply this to all files - only to files which you actually modified.

Taaaaaserface:
> Removing unused imports in Eclipse is usually a matter of using the ["Organize imports"](https://www.mkyong.com/intellij/eclipse-ctrl-shift-o-in-intellij-idea/) functionality of Eclipse. Do not apply this to all files - only to files which you actually modified.

I ran **mvn clean install**  on the command line and it reported error like this
image
then I turn to Task.java and  used the "Organize imports" functionality of Eclipse, but seems nothing happend. what should I do?
Should I do like this:
1. delete all imports in Task.java and use "Organize imports" functionality (ctrl+shift+o)
2. Then choose type to import which are 
image
3. After importing either two type the mvn clean install comment is still wrong.

I wanna ask where my order is wrong and what should be the correct order.

reckart:
Looks like your system cannot execute the tests - maybe because it doesn't have enough memory. Try `mvn -DskipTests clean install`.

Taaaaaserface:
> Looks like your system cannot execute the tests - maybe because it doesn't have enough memory. Try `mvn -DskipTests clean install`.

Is this the same question?
image

reckart:
Whoohoo!!! 2684 checkstyle violations! :) I wonder what's happening there...

If you scroll up in the terminal window, you should see messages telling you what checkstyle thinks is wrong. 

Do you see any local modifications in your git client?

reckart:

Issue #1117, opened 2019-04-11
Ability to export recommender model


reckart:
**Is your feature request related to a problem? Please describe.**
Let's say I annotate a text with the help of the string matching recommender. The recommender learns all the possible aliases for my concepts. It would be nice if there was a way I could export this dictionary. And while at it, it would be interesting in general to export the recommender models.

**Describe the solution you'd like**
Offer an option in the recommender sidebar where the user can select a recommender and press a button to download the current recommender model.

**Describe alternatives you've considered**
Downloading recommender models could also be done in the project settings. But considering that models are currently per-user, allowing the download from the annotation page might make more sense.

munterkalmsteiner:
@reckart I assume that the export should provide some sort of human readable (text) representation of the model? The purpose is to inspect the model, but there is currently no need to import the model back?

reckart:
There are probably many motivating scenarios for exporting the model. Here are three

* using the model with outside INCEpTION, e.g. with OpenNLP on the command line
* re-importing the model as a baseline into another project (re-importing doesn't yet have an issue I believe)
* inspecting the model (e.g. for dictionaries learned by the string matcher, but certainly not for OpenNLP models)

So: the representation would depend on what classifier/recommender is used

* for the StringMatchingRecommender, it would be good to export in the same format that is used to supply gazeteers to the recommender.
* for the OpenNLP-based recommenders, it would be good to use the binary format(s) which OpenNLP works with

reckart:
Recommenders store they models in-memory within a `RecommenderContext` instance.

I guess a good approach to address this issue would be roughly:

* obtain all annotation CASes from all users (or from a specific user) in the current project (cf. `de.tudarmstadt.ukp.inception.recommendation.tasks.TrainingTask.readCasses(Project, User)`)
* manually construct a `RecommenderContext`
* instantiate the recommender factory via `de.tudarmstadt.ukp.inception.recommendation.api.RecommendationService.getRecommenderFactory(Recommender)`
* instantiate the recommender via the factory
* pass the context and the CAS to the `train` method so the recommender trains the model and stores it into the context
* add a new method to the `RecommendationEngine` maybe with a signature like `void exportModel(RecommenderContext aContext, OutputStream aTarget)` which would take the model and serialize it to the target stream - and implement that method in e.g. the StringMatchingRecommender.

reckart:
[...]

PR #1473, opened 2019-10-18
feature/#1117 export recommender model 

clarkhywh:
**What's in the PR**
Offer an option in the recommender sidebar where the user can select a recommender and press a button to download the current recommender model.

Now we only change the UI that add the button of exporting the recommender model. But we have 3 question.
1. what the recommender model format should I export
2. Now I only can get the recommender which the user has, but we don't know how to find the data of recommender. Is it in the model object or recommender context?


clarkhywh:
@reckart could you expand a little bit on the problem?

munterkalmsteiner:
@clarkhywh did you see the explanations in #1117?

munterkalmsteiner:
Closing this one as I assume #1534 is the more recent PR.

PR #1474, opened 2019-10-18
Implementation/1353 Jump to slot filler

Jackwulujie:
We group 4 are responsible for implementing the function of jumping to slot filler.
Honestly, we don't really understand this issue, especially what slot represents in this application.
First, according to the instructions in the user guide, we have enabled the document metadata function, as follows 
image
Then because we don't know what slot represents here, we have no idea how to use the things I created, like test, Jack, yes. 
image
Maybe we need to follow this step, but I failed to fill the slot by double-clicking on a span annotation.
Actually, we don't know what will happen when this function is implemented. 
Our guess is，we don't know if it is correct.，there can be several annotations on a word, but one of them can be linked to other annotations, which is a slot filler, but we can't see it directly. So in order to distinguish this annotation (slot filler) with link function from other annotations, we should add a visible icon on it.
Do I understand this correctly or can you help me explain it in more detail?


Jackwulujie:
Sorry, I used the wrong branch.

PR #1475, opened 2019-10-18
feature/1436 rename variable

clarkhywh:
This is my second time to create the pull request. Last  PR is closed by me because some problem. 
What should I do Is to clear code smell which is ""X" hides or potentially hides a variable declared in an outer scope at line Y.". After discussing with Michael I found the files I should improve are all automatically generated files or files of Mozillam, so they should not be changed manually.


inception-html-editor/src/main/java/de/tudarmstadt/ukp/inception/htmleditor/annotatorjs/resources/annotator-full.js
```diff
@@ -13,13 +12,14 @@
 
 //
 
+
 // Generated by CoffeeScript 1.6.3
```
munterkalmsteiner: This file is, as the comment suggests, also generated. Since this is the only file in the pull request and no other instances of this smell appear in the code, I suggest to close this pull request.

PR #1478, opened 2019-10-21
feature/1437 extract nest method

DEMONSTAT:
This is my second time to create the pull request. Last PR is closed by me because some problem.

**Describe the refactoring action**
An example of the issue：
In the part of the main package I modified, I found that the content of the has method is the same, so I tried to extract this part and then package it, so that the packaged code part becomes less difficult to understand.

**Expected benefit**
Long code segments are not conducive to understanding and modification. For these large pieces of code, we can refactor them to improve readability and modifiability. I don't think it's very difficult to make specific changes, but I spent a lot of time understanding the code. After my modifications, the code became more orderly. Good design is half the battle. Stopping to improve the design through refactoring may slow down the current speed, but the latecomer advantage it brings cannot be underestimated.

### DIFF
--- a/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/querybuilder/SPARQLQueryBuilder.java
+++ b/inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/querybuilder/SPARQLQueryBuilder.java
@@ -266,22 +266,14 @@ public class SPARQLQueryBuilder
                 classPatterns.add(VAR_SUBJECT.has(Path.of(oneOrMore(subClassProperty)), aContext));
                 classPatterns.add(VAR_SUBJECT
                         .has(Path.of(typeOfProperty, zeroOrMore(subClassProperty)), aContext));
-                if (OWL.CLASS.equals(aKB.getClassIri())) {
-                    classPatterns.add(VAR_SUBJECT.has(
-                            Path.of(OWL_INTERSECTIONOF, zeroOrMore(RDF_REST), RDF_FIRST),
-                            aContext));
-                }
-                
+                SelectValue(aKB, aContext, classPatterns);
+
                 return GraphPatterns.union(classPatterns.stream().toArray(GraphPattern[]::new));
             }
             case CLASS: {
                 List<GraphPattern> classPatterns = new ArrayList<>();
                 classPatterns.add(VAR_SUBJECT.has(Path.of(oneOrMore(subClassProperty)), aContext));
-                if (OWL.CLASS.equals(aKB.getClassIri())) {
-                    classPatterns.add(VAR_SUBJECT.has(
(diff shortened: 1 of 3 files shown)


inception-kb/src/main/java/de/tudarmstadt/ukp/inception/kb/querybuilder/SPARQLQueryBuilder.java
```diff
@@ -450,7 +426,23 @@ protected GraphPattern rootsPattern(KnowledgeBase aKb)
             }            
         }
     }
-    
+
+    private static void selectValue(KnowledgeBase aKB, Iri aContext, List<GraphPattern> classPatterns) {
```
munterkalmsteiner: I think you should refactor these two methods into one. Maybe you could have a method like `private void createPattern(KnowledgeBase aKB, RdfSubject subject, RdfObject object, List<GraphPattern> patterns)`. When called, subject and object can be defined as needed.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/querybuilder/SPARQLQueryBuilderGenericTest.java
```diff
@@ -127,29 +127,37 @@ public void setup() throws Exception
 
         switch (kb.getType()) {
         case LOCAL: {
-            LuceneSail lucenesail = new LuceneSail();
-            lucenesail.setParameter(LuceneSail.LUCENE_RAMDIR_KEY, "true");
-            lucenesail.setBaseSail(new MemoryStore());
-            repo = new SailRepository(lucenesail);
-            repo.init();
-            importData(repo, profile.getAccess().getAccessUrl());
+            localRepo();
```
munterkalmsteiner: Maybe `createLocalRepo` is better as it indicates the action.

inception-kb/src/test/java/de/tudarmstadt/ukp/inception/kb/querybuilder/SPARQLQueryBuilderGenericTest.java
```diff
@@ -127,29 +127,37 @@ public void setup() throws Exception
...
         }
         case REMOTE: {
-            Assume.assumeTrue(
-                    "Remote repository at [" + profile.getAccess().getAccessUrl()
-                            + "] is not reachable",
-                    isReachable(profile.getAccess().getAccessUrl()));
-            
-            repo = new SPARQLRepository(profile.getAccess().getAccessUrl());
-            repo.init();
+            RemoteRepo();
```
munterkalmsteiner: Same here. `createRemoteRepo()`

munterkalmsteiner:
@DEMONSTAT do you intend to work on my suggested changes?

reckart:
Assuming this is no longer going to be worked on.

Issue #994, opened 2019-02-28
Restrict URLs for image features


reckart:
**Is your feature request related to a problem? Please describe.**
Any URL can currently be entered for an image feature. It would be good to limit the URL. Note that this is not really critical because the image is loaded browser-side - meaning the user cannot access anything beyond what is accessible by the browser anyway. In particular, it is not possible to access any resources on the INCEpTION server.

**Describe the solution you'd like**
As traits for the image feature, it would be good to be able to
* limit the possible URL scheme (http, https, ftp, etc.)
* required that the URLs point to a particular server
* possibly define a regex used for URL validation

**Additional context**

henriktao:
Hi @reckart, how is the status of this issue? Is it resolved yet? 

I have looked into the inception-image module and found the use of org.apache.wicket.validation.validator.UrlValidator for URL validation inside the class ImageFeatureEditor.java. The validator is currently only accepting HTTP and HTTPS as URL schemes. Is this a solution to the issue or am I mistaken?

reckart:
Nobody has worked on the issue yet. The initial description of the issue lists several possible improvements - an URL validator that only accepts http/https only addresses the first point.

reckart:
The issue is not how the regex would look or which URLs to accept/deny. The issue is that the feature configuration UI offers an option for the project manager to enter a regex that defines which URLs are acceptable. E.g. if the project manager knows that all images should come from `https://safe.images.com/` data includes an image URL to `http://here.is.virus/image.png`, then the manager could set up a regex that matches only the former and not the latter. In that way the manager could establish a protection from obtaining images from untrusted sources.

PR #1482, opened 2019-10-23
feature/994-Restrict-URLs-for-image-features

JaviValera:
**What's in the PR**
*  To ensure the security and prevent future errores is necessary to control the input URL from external images given by the user. Therefore we are going to check the protocol of the image URL , if this URL has the correct format and to verify that the server connected to is a controlled one. For this purpose we are going to make use of regular expressions to validate the URL inputs.

**How to test manually**
* Settings -> new Layer -> new Fact -> Feature -> URL image. After we redirect to any project and click into annotation there we create a new annotation and add a layer fact, in the we insert the Url and it can be debugged.

**Automatic testing**
* Not applies

**Documentation**
* Not applies

### DIFF
--- a/inception-image/src/main/java/de/tudarmstadt/ukp/inception/image/feature/ImageFeatureTraitsEditor.properties
+++ b/inception-image/src/main/java/de/tudarmstadt/ukp/inception/image/feature/ImageFeatureTraitsEditor.properties
@@ -13,3 +13,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+regex="Matching URLs against a suitable regex"


reckart:
Thanks for the PR :) Adding FTP to the list of valid protocols doesn't really address the issue though. The idea of having a regular expression to check the URL is not only to ensure it uses a particular protocol, but also to e.g. ensure that it is scoped to a particular (set of) domain(s), that it doesn't contain URL arguments, that it has a maximum length, etc etc. I don't know if you are still working on the PR. Btw. GitHub has a concept of [PR drafts](https://github.blog/2019-02-14-introducing-draft-pull-requests/) which allow to open a PR and indicate that one is still working on it. Might be something to try out (I haven't worked with PR drafts yet either). @munterkalmsteiner

munterkalmsteiner:
@JaviValera, I agree with @reckart. I have communicated to a student (Jesus) which I think is also in your group the following:
Isn't the idea of this requirement to be able to add constraints to urls in image features? See the screenshot where I created a custom layer and a new image url feature. The user should now be able to define here the three constraints on the URL.

- limit the possible URL scheme (http, https, ftp, etc.)
- required that the URLs point to a particular server
- possibly define a regex used for URL validation 

That is, you don't hardcode anything, you need to provide the functionality in the configuration of a feature that the user can add any of the three constraints. Then, you need to check whether the constraints are fulfilled when a user creates such an annotation.
image

@reckart, is my inpretation of #994 correct?

reckart:
Right :) But I wouldn't add three settings (URL scheme, server, regex) but only regex because by matching URLs against a suitable regex, you can already constrain the scheme / server. And if you have multiple settings, it would bring up the question how they interact with each other, in particular regex with the other two.

JesusDAR:
We are now discussing this issue, from what we understand we have to add a new checkbox to validate the img url and protocol through a regex that is already implemented. Are we right? @munterkalmsteiner

JesusDAR:
We have seen that the modules we need to modify are from webanno, more specifically the checkboxes. The file we think we need to modify are read-only because they belong to webanno. We ask guidance about this issue. @munterkalmsteiner  @reckart

reckart:
INCEpTION re-uses parts of WebAnno and our team develops in both repositories simultaneously. It happens regularly that issues are opened in the INCEpTION repo, but the some (or all) of the changes need to be made in WebAnno. In such cases, we open a second issue in WebAnno and prepare a PR there. We also usually "link" the two issues by putting e.g. the URL of the INCEpTION issue into the description of the related WebAnno issue.

reckart:
For what it's worth: as part of our work on INCEpTION, we also did contribute improvements to DKPro Core, UIMA, uimaFIT, RDF4J and possibly other projects - that's open source community development :)

reckart:
Anyway: the image feature is not part of WebAnno. It is part of INCEpTION here:

https://github.com/inception-project/inception/tree/master/inception-image

You can add settings specific to the image feature to `ImageFeatureTraits` and  `ImageFeatureTraitsEditor`.

munterkalmsteiner:
@JaviValera please notify me if I should review the code again. Otherwise, please close the pull request.

PR #1488, opened 2019-10-24
Feature/#1268Regex-based validation string field

jianglinyao1996:
**What's in the PR**
implement a new feature that allow user can optionally use regular expression to validate string field.
Please do not merge this changes. I just changed something that I can create pull request. I need a clearer explanation before I submit a formal code change.

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-support/src/main/java/de/tudarmstadt/ukp/inception/kb/io/FileUploadDownloadHelper.java
+++ b/inception-support/src/main/java/de/tudarmstadt/ukp/inception/kb/io/FileUploadDownloadHelper.java
@@ -41,7 +41,7 @@ public class FileUploadDownloadHelper
 
     private final IFileCleaner fileTracker;
 
-    private final String INCEPTION_TMP_FILE_PREFIX = "inception_file";
+    private static final String INCEPTION_TMP_FILE_PREFIX = "inception_file";
 
     public FileUploadDownloadHelper(Application application)
     {


jianglinyao1996:
As we discussed in issues. You said"The UimaStringTraits and UimaStringTraitsEditor would be extended so the project admin can optionally enter a regular expression on the layers configuration page for String features to validate values entered by the annotators on the annotation page.
The validation should probably happen in PrimitiveUimaFeatureSupport.setFeatureValue(...) where we also check if a tag is valid/can be added (in case a tagset is configured).readUimaStringTraits(...) can be used  in PrimitiveUimaFeatureSupport.setFeatureValue(...) to extract the traits object from the AnnotationFeature object."
 These classes are in the package of webanno. Inception used remote link to visit these classes which means I can not push the changes here. Should I fork it and create a new pull request in webanno project?(https://github.com/webanno/webanno)
If I have changes both on Inception-project and webanno, should I push it separately?

jianglinyao1996:
For further discussion. Your need to describe more about this issues. what would happen if the values from annotation page not match the regular expression which entered by project admin? The  project push an error message to user or something else?

reckart:
> Your need to describe more about this issues. what would happen if the values from annotation page not match the regular expression which entered by project admin?

A String feature can optionally be associated with a TagSet. If the `PrimitiveUimaFeatureSupport.setFeatureValue(...)` determine that a value is invalid because it is not in the tagset, an `IllegalArgumentException` is thrown. The same could be done if a regex is set and the value does not match the regex.

https://github.com/webanno/webanno/blob/8db27bc420e302352675718fef68cbb3042ab98f/webanno-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/feature/PrimitiveUimaFeatureSupport.java#L143-L165

> If I have changes both on Inception-project and webanno, should I push it separately?

INCEpTION re-uses parts of WebAnno and our team develops in both repositories simultaneously. It happens regularly that issues are opened in the INCEpTION repo, but the some (or all) of the changes need to be made in WebAnno. In such cases, we open a second issue in WebAnno and prepare a PR there. We also usually "link" the two issues by putting e.g. the URL of the INCEpTION issue into the description of the related WebAnno issue.

Issue #1492, opened 2019-10-26
New style for document metadata sidebar


wpmp:
**Describe the refactoring action**
Currently document annotations in the document metadata list can only be opened by clicking on the UI label of its features. Also nothing indicates that the elements of the list can be opened.
In this refactoring the style of the list will be changed to be similar to an accordion.
Also if no UI label of the features is available nothing will be shown instead of the name of the document annotation.

**Expected benefit**
Better UX for document annotations

PR #1493, opened 2019-10-26
#1492 - New style for document metadata sidebar

wpmp:
**What's in the PR**
- apply accordion style to the document annotation list
- if no feature preview is available show nothing instead of the document annotation label

**How to test manually**
* Play around with the document metadata sidebar

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I'd suggest to move the arrow to the left side. 

When I open the sidebar, currently all arrows point down. When I click on an arrow, the entry opens. When I click on it again, it closes and the arrow points up. I guess it should also point up at the star then.

Actually, instead of having the arrow pointing up, I think it should be pointing to the right - or alternatively consider using a "+" and "-" instead of the arrows.

wpmp:
Done.

It should have shown the angle down when closed and up when opened, I must have broken it at some point and didn't notice it then.
But this should work correctly now.

wpmp:
Need to check something before the next review

wpmp:
Ok, now it's ready.

Issue #1041, opened 2019-03-14
add possibility of defining keyboard shortcuts


mromanello:
**Is your feature request related to a problem? Please describe.**

In the context of the [impresso project](http://impresso-project.ch) we are conducting a pretty large annotation campaign (on historical newspaper materials) entirely based on INCEpTION as an annotation environment. 

We realised that having to select each annotation type with the mouse (for a larger number of annotations) takes up a considerable amount of time. 

**Describe the solution you'd like**

Having the possibility of defining keyboard shortcuts (at a user- or project-level) for certain types of annotations (i.e. named entities in our case) would be extremely helpful. E.g. a shortcut would map the letter `p` to the NE type `Person`.

**Describe alternatives you've considered**

Just accept that it takes longer :)

reckart:
Could you describe the actions the user should take in more detail? It seems to me that having hot-keys to select certain layers doesn't solve your problem because you still need to mark the text with the mouse.

reckart:
Normally, I would expect that you have a single named entity layer and that you specify the type of named entity using a feature. So my expected workflow would be:

1. user selects layer **Named entity**
2. user marks mention of entity in the text
3. cursor automatically jumps to the feature editor in the right sidebar where the type of entity can be specified
4. user starts typing (e.g. `P`) and the auto-complete selects the first available tag with starting with *P*, e.g. *Person*
5. user presses enter
6. go back to step 2.

Here, there is no need to switch between layers.

mromanello:
Yes, I had in mind the same workflow with only one difference:

4. user hit the keystroke ctrl+P (*) which automatically selects the Person entity (pressing enter is not necessary).

(*) this shortcut has been defined in the settings, where each Value in an annotation layer may get assigned a different shortcut; such shortcuts can be defined e.g. for the most frequent entities. Defined shortcuts should probably be displayed in the entity dropdown alongside the value.

mromanello:
@reckart just wondering if you had a chance to look at this? If you need more details on the scenario we had in mind just let me know.

reckart:
To be honest, at the moment, I do not see a benefit of pressing `CTRL-P` over `P + ENTER`.

mromanello:
Well, the main benefit is having only to press a keystroke combination without having to select the text field, start typing and then select from the dropdown.

reckart:
[...]

PR #1515, opened 2019-11-17
#1041 - add possibility of defining keyboard shortcuts

reckart:
**What's in the PR**
- First throw at key bindings for concept features (WIP)

**How to test manually**
* Open the `identifier` feature of the `Named Entity` layer in the project settings
* On the bottom of the feature editor, there are two new fields followed by a plus icon.
* Into the first field, enter a space-separated key combo following the names defined by [`wicket.contrib.input.events.key.KeyType`](https://github.com/wicketstuff/core/blob/master/input-events-parent/input-events/src/main/java/wicket/contrib/input/events/key/KeyType.java#L26-L28). The second field is actually an auto-completing concept feature editor which allows to select some item from the KB (according to the other traits settings).
* Define one or two single key shortcuts
* Go to the annotation page
* Create a new `Named Entity` annotation
* Note that the focus does **not** go into the `identifier` input field! If the focus is on a text field, the shortcuts are not considered
* Press the key for the shortcut you defined
* See that the value is being set

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@mromanello this is in an early prototype stage. But if you care to give it a try, please let me know what you find/think.

Issue #1518, opened 2019-11-18
Document where the settings.properties file lives


UWinch:
**Is your feature request related to a problem? Please describe.**
There is some description in the admin guide-> application home folder. However, it should be made clearer that the settings.properties file is not present automatically but needs to be created.

shohrehhd:
Hello,
I am following the guidelines on "Storing data on the host" to set up inception using docker. I need to change the CSRF setting on the setting.properties, however I cannot find the setting file as said in: srv/inception

> Thus, if you follow the instructions above, the settings file would go to /srv/inception/settings.properties on the host system.

Do I need to create it myself and if so what settings should I add aside from the ones I need to set for CSRF?

Thanks in advance,

UWinch:
Hi @shohrehhd , yes, you will need to create the settings.properties file yourself. If you only want to change this one specific property, you don't need to add anything else to the file.

jcklie:
Reopen as there were some unclear points

PR #1521, opened 2019-11-22
#1518 Document where the settings.properties file lives

UWinch:
**What's in the PR**
Added some clarification that the file needs to be created and linked to the explanation on application home folder

**How to test manually**
* Read the documentation

### DIFF
--- a/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
+++ b/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
@@ -60,7 +60,7 @@ include::{include-dir}remote-api-webhooks.adoc[leveloffset=+1]
 
 
 
-include::{include-dir}settings.adoc[leveloffset=+1]
+include::{include-dir}settings.adoc[]
 
 include::{include-dir}settings_general.adoc[leveloffset=+1]
 
--- a/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
+++ b/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
@@ -60,7 +60,7 @@ include::{include-dir}remote-api-webhooks.adoc[leveloffset=+1]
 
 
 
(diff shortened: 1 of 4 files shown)


inception-doc/src/main/resources/META-INF/asciidoc/admin-guide.adoc
```diff
@@ -60,7 +60,7 @@ include::{include-dir}remote-api-webhooks.adoc[leveloffset=+1]
 
 
 
-include::{include-dir}settings.adoc[]
+include::{include-dir}settings.adoc[leveloffset=+1]
```
reckart: Why did you increase the indentation level?

inception-layer-docmetadata/src/main/resources/META-INF/asciidoc/user-guide/annotation_create-annotations_docmeta.adoc
```diff
@@ -18,8 +18,8 @@ On the annotation page, you can now:
 
 image::metadata-sidebar.png[align="center"]
 
-NOTE: This is an experimental feature not recommended for serious use. However, you can try it out
-      and enable it by adding `documentmetadata.enabled=true` to the `settings.properties` file. In 
+NOTE: This is an experimental feature not recommended for serious use. In 
       particular, curation of document metadata annotations is not possible and they are not supported
-      by the TSV format. They can be im-/exported via XMI or Binary CAS though.
+      by the TSV format. They can be im-/exported via XMI or Binary CAS though. However, you can try it out
+      and enable it by adding `documentmetadata.enabled=true` to the `settings.properties` file. This file is optional and might need to be created first in the application home folder <<admin-guide.adoc#sect_home_folder>>.
```
reckart: I think the link for the home folder documentation could be in `settings.adoc` and be removed here. And instead of always repeating that the settings file is optional, a link to `settings.adoc` could be added - hoping that users who want to know more about the file would follow the link and then read there that the file is optional and needs to be created manually if necessary.

Issue #1531, opened 2019-11-26
PDF recommendation color is really difficult to read


jcklie:
**Is your feature request related to a problem? Please describe.**
PDF recommendation color is really difficult to read

**Describe the solution you'd like**
Change the color to something more readable, maybe move the recommendation behind the text and make it solid instead of transparent.

reckart:
New PDF editor (https://github.com/inception-project/inception/issues/3096) has an improved coloring style.

PR #1536, opened 2019-11-27
#1531 - PDF recommendation color is really difficult to read

reckart:
**What's in the PR**
- Upgrade to pdf.js 1.10.100 - the version previously used (1.4.something) didn't support setting the background color of the canvas.
- Configure pdf.js to render the canvas background transparently. This isn't enough because surrounding HTML elements have a white background and also the z-indices of pdf.js and pdfanno parts need to be adjusted in order to move the annotations behind the text. There could also be a problem with PDFs that explicitly declare a background color.
- ... WIP

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Need to re-do from scratch since there were many changes to the PDF code meanwhile - also it might not work at all because of the PDFs painting the canvas background instead of leaving it transparent.

Issue #1549, opened 2019-12-03
Filter entity candidates by description


reckart:
**Is your feature request related to a problem? Please describe.**
Searching for an entity matches against the label of the entity in the knowledge base. The description is not considered. Also, searching in the description of all entities would not be terribly helpful because it would likely generate way too many results.

**Describe the solution you'd like**
I would like to enter something like `county :: queen` into the KB feature editor. The expected behavior is that the usual match against the KB label is done using the part before the `::` - so we retrieve any items matching `county` in the label. Afterwards, the term after the `::` is used to match in the description. So here, any matching KB items with the term `queen` in their description should be matched. The usual Levenshtein-distance match should be applied to the description match and the items should be ranked according to the distance.

**Describe alternatives you've considered**
Searching in the description of all entities would not be terribly helpful because it would likely generate way too many results.

PR #1554, opened 2019-12-05
#1549 - Filter entity candidates by description

reckart:
**What's in the PR**
- Allow filtering candidates by description using `::` separator
- Allow filtering candidates by label using `"..."` quoting
- Added documentation for the new features

**How to test manually**
* Configure Wikidata as the KB
* Annotate a document on the Named Entity layer
* See documentation how to use the quote and :: operators
* Try using them to filter candidates return by Wikidata

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
That is because Wikidata doesn't seem to return results for `Barack Hussein Obama`.

And the other one was a bug :) Thanks for testing!

Issue #1559, opened 2019-12-11
Problems if multiple entities have the same label


reckart:
**Describe the bug**
The KB autocomplete field fails to send the right selection to the backend if multiple items in the dropdown list have the same label. This is because the Kendo auto-complete JS code assumes that every item has a different label and otherwise does not properly trigger change events amongst other problems. But also the Wicket KendoAutoComplete backend code makes the same assumption and selects the always the first item with the given label.

**To Reproduce**
1. Configure Wikidata as a KB
2. Try to link some word to `Chicago`
3. There are multiple items in Wikidata with this label. Try to select the first one, then try to select the second one.
4. You should note that after selecting the second one, the linked item is actually still the first one.

**Expected behavior**
The frontend should send the IRI of the concept back to the server and the server should use that to make the selection.

**Please complete the following information:**
 - Version and build ID: 0.14.0

**Additional context**
- Upstream issue: https://github.com/sebfz1/wicket-jquery-ui/issues/313

PR #1560, opened 2019-12-11
#1559 - Problems if multiple entities have the same label

reckart:
**What's in the PR**
- Fixes issue of items with the same label being confused by frontend and backend by using IRI as an identifier
- Improve performance of linking an entity by not fetching the choices again if we still have them from filling the dropdown
- Avoid sending redundant data to the UI
- Add area with entity description below the entity linking field

**How to test manually**
* See issue description

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-layer-docmetadata/pom.xml
+++ b/inception-layer-docmetadata/pom.xml
@@ -92,10 +92,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.wicket</groupId>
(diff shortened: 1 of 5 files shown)


jcklie:
@reckart Build fails

reckart:
Has issues - the current state breaks e.g. search on the KB page. I'm looking into it again.

Issue #1572, opened 2019-12-17
Safeguard CAS against concurrent accesses


reckart:
**Describe the refactoring action**
Wrap the CAS in a dynamic proxy which checks that a CAS is not accessed from a different thread.

**Expected benefit**
Catch cases where unsafe accesses to CAS objects are made.

PR #1573, opened 2019-12-17
#1572 - Safeguard CAS against concurrent accesses

reckart:
**What's in the PR**
- Added dealing with the CAS wrapper
- Fixed concurrent access caused by the indexing service

**How to test manually**
* Use the application
* check for ClassCastExceptions because which mean that the proxy may need to be removed via `getRealCas()`
* check for exceptions due to cross-thread access to the CAS

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Depends on https://github.com/webanno/webanno/pull/1570

reckart:

PR #1588, opened 2019-12-31
Feature/1587 disable fact linking support

reckart:
**What's in the PR**
- Exclude fact linking documentation from the user's guide
- Disable fact layer project initializer
- Keep all files and in particular keep the SubjectObjectFeatureSupport and PropertyFeatureSupport such that existing projects with the fact layer do not break

**How to test manually**
* Create a new project and check that fact layer is no longer there
* Open an old project and see that the fact layer is still there and also that it can be used for annotation
* Check that fact linking is not in documentation

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:

reckart:

Issue #1597, opened 2020-01-15
Several format examples cannot be used copy-paste


reckart:
**Describe the bug**
Several format examples cannot be used copy-paste because the tab characters used in the examples are auto-converted to spaces and the corresponding readers do not like the spaces.

**Expected behavior**
It should be possible to copy-paste an example from the documentation and import it, e.g. for CoNLL formats.

See also: https://github.com/webanno/webanno/issues/1587

PR #1598, opened 2020-01-15
#1597 - Several format examples cannot be used copy-paste

reckart:
**What's in the PR**
- Disable auto-conversion of tabs to spaces

**How to test manually**
* Copy VRT example to a file and then try importing it

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-app-webapp/pom.xml
+++ b/inception-app-webapp/pom.xml
@@ -650,6 +650,7 @@
         <plugin>
           <groupId>org.asciidoctor</groupId>
           <artifactId>asciidoctor-maven-plugin</artifactId>
+          <version>${asciidoctor.plugin.version}</version>
           <configuration>
             <sourceDirectory>target/${project.artifactId}-${project.version}-documentation/META-INF/asciidoc</sourceDirectory>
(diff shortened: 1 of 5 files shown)


reckart:
How did you build? On the command line or in Eclipse?

UWinch:
I build on the console. I'm not sure what the problem might be, could you have a look at it again?

reckart:
Ok, I found the problem. The `GenerateDocumentation` code tried to configure the `coderay` source-highlighter, but it actually didn't because it tried to do it as an attribute instead of an option. And therefore `tabsize` worked there because coderay wasn't turned on.

In the maven build, coderay was turned on and it does not seem to support `tabsize` meaning that the tabs were expanded.

I now removed coderay everywhere.

Mind that most format samples are sources from WebAnno, so we'll only get them once we upgrade to a WebAnno version which has them (i.e. > 4.0.0-beta-9). One which *is* now copyable is the "VRT" example which is sourced from INCEpTION.

reckart:
Any objections against merging this one?

Issue #1633, opened 2020-03-12
Exception when session expires


reckart:
**Describe the bug**
There is an exception when the session expires. The stack trace suggests it comes from the changes related to the curation-on-annotation page.

**To Reproduce**
1. Log in
2. Wait until the session expires

**Expected behavior**
No error.

**Screenshots**
<details>
<summary>Stacktrace</summary>

```
2020-03-12 16:03:08 [http-nio-8080-exec-1] ERROR [anonymousUser] [/inception-app-webapp] - Session event listener threw exception
java.lang.NullPointerException: null
	at de.tudarmstadt.ukp.clarin.webanno.security.UserDaoImpl.getRoles(UserDaoImpl.java:212) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.security.UserDaoImpl.isAdministrator(UserDaoImpl.java:178) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.security.UserDaoImpl$$FastClassBySpringCGLIB$$ef901b5c.invoke(<generated>) ~[classes/:?]
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218) ~[spring-core-5.2.3.RELEASE.jar:5.2.3.RELEASE]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.invokeJoinpoint(CglibAopProxy.java:769) ~[spring-aop-5.2.3.RELEASE.jar:5.2.3.RELEASE]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:163) ~[spring-aop-5.2.3.RELEASE.jar:5.2.3.RELEASE]
	at org.springframework.aop.framework.CglibAopProxy$CglibMethodInvocation.proceed(CglibAopProxy.java:747) ~[spring-aop-5.2.3.RELEASE.jar:5.2.3.RELEASE]
	at org.springframework.transaction.interceptor.TransactionAspectSupport.invokeWithinTransaction(TransactionAspectSupport.java:366) ~[spring-tx-5.2.3.RELEASE.jar:5.2.3.RELEASE]
	at org.springframework.transaction.interceptor.TransactionInterceptor.invoke(TransactionInterceptor.java:99) ~[spring-tx-5.2.3.RELEASE.jar:5.2.3.RELEASE]
[...]

PR #1635, opened 2020-03-14
#1633 Exception when session expires

UWinch:
**What's in the PR**
* Fixes NPE when storing curation settings for a user.

**How to test manually**
* see Issue

### DIFF
--- a/inception-curation/src/main/java/de/tudarmstadt/ukp/inception/curation/CurationServiceImpl.java
+++ b/inception-curation/src/main/java/de/tudarmstadt/ukp/inception/curation/CurationServiceImpl.java
@@ -325,6 +325,10 @@ public class CurationServiceImpl
     private void storeCurationSettings(String aUsername)
     {
         User currentUser = userRegistry.get(aUsername);
+        if (currentUser == null) {
+            return;
+        }
+        
         for (Project project : projectService.listAccessibleProjects(currentUser)) {
             Long projectId = project.getId();
             Set<String> usernames = null;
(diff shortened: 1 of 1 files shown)


UWinch:
I wanted to save on database accesses. I thought flush is necessary to actually write through to the database. Do you think I should change something before you accept the PR?

reckart:
Saving on DB accesses is ok if they happen often (I'm just as we speak in the process of resolving some performance issues due to DB hammering). But in this case, how often does the user change  the curation settings? Almost never probably. 

So it wouldn't hurt saving the settings immediately. You can still keep them in memory (in the annotator state) to avoid having to read them from the DB on every action. 

I was wondering if you wanted to save only on session end in case a user logs in using two different sessions simultaneously (e.g. two browsers) - but that's something generally discouraged with INCEpTION / WebAnno anyway because it can break all kinds of things.

The `@Transactional` annotation takes care of saving stuff to the DB, so the explicit flush is not needed in most cases.

I'm a bit undecided on merging it without change. Well, if it works, should be ok, right? But I wanted to make you aware of some design issues in the code. You can have another look at it or we merge it as running code (note, I didn't check (yet) if it actually works) and book any remaining comments under technical dept to be addressed later.

UWinch:
I understand your points. To be honest, I would like to merge it now and come back to the issues when I work on the next features for this (e.g. aggregating user annotations) because I would like to work on something different for a time. I would make a note on this in the associated issue check list.

reckart:
Ok, we don't need to do a big refactoring now. It would be good if we could remove the `.flush()` call though because I think it is not needed at all. 

The fix you added was checking if the user existed in the DB and if not bailing out. So I guess what might have happened here was that for testing purposes, I deleted a user directly from the DB and then the session of this user timed out. Probably this would never happen under normal circumstances because the application does not (yet) allow to delete users.

Anyway, maybe you could add a comment to the check you added, e.g. something like `If the user no longer exists, we do not store the curation settings either`.

UWinch:
@reckart ok to merge now?

Issue #1602, opened 2020-01-22
NEL only looks for first word and not entire span


JackDapid:
**Describe the bug**
Having a NEL running on a Token-level base NE layer only looks in Knowledge Base for the first word

**To Reproduce**
- Have an active named entity layer
Validation: Always
Granularity Level: Token-level
Overlap: Any
- Add a [Named entity@identifier] Named Entity Linker
Feature: identifier

**Expected behavior**
The NEL should look foremost for the entire NE tag: Human Rights Office and not only the first word 'Human'

**Screenshots**
![image](https://user-images.githubusercontent.com/8346230/72896840-e21caf80-3d20-11ea-95de-019140c6fa9f.png)

**Please complete the following information:**
 - Version and build ID: a44b4c76b19fcc3d0a58e75a82aafdd96a564500
 - OS: OSX client, CentOS server
 - Browser: Chrome

reckart:
Hm, indeed. In particular since the NEL recommender only kicks in if there already is a named entity annotation, it should really take the entire covered text into account.

Sounds like it should not be too difficult to fix.

jcklie:
The logic in the named entity linker is really difficult to understand. Why are we predicting explicitly for tokens and not the whole named entity?

reckart:
I have no idea... maybe to ensure that annotation boundaries match token boundaries and that they don't go across sentences? ... but actually the behaviors of the NamedEntity layer should already enforce that...

PR #1639, opened 2020-03-16
#1602 - NEL only looks for first word and not entire span

jcklie:
**What's in the PR**
- Use annotation boundaries
- Fix empty bean jackson serialization
- Allow overlapping suggestions for different features

**How to test manually**
* See bug report

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
+++ b/inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
@@ -21,15 +21,8 @@ import static de.tudarmstadt.ukp.clarin.webanno.api.annotation.util.WebAnnoCasUt
 import static de.tudarmstadt.ukp.clarin.webanno.api.annotation.util.WebAnnoCasUtil.selectSentences;
 import static org.apache.uima.fit.util.CasUtil.getType;
 import static org.apache.uima.fit.util.CasUtil.indexCovered;
-import static org.apache.uima.fit.util.CasUtil.selectCovered;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import org.apache.commons.lang3.tuple.ImmutablePair;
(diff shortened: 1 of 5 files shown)


inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
```diff
@@ -21,28 +21,21 @@
...
-
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
+
+import java.util.*;
```
reckart: No * imports

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinkerTraits.java
```diff
@@ -19,6 +19,9 @@
 
 import java.io.Serializable;
 
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
```
reckart: Dependency needs to be added to the POM - build seems to fail because of this.

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
```diff
@@ -1470,7 +1470,7 @@ public void calculateVisibility(CAS aCas, String aUser, AnnotationLayer aLayer,
                     for (AnnotationFS annotation : annotations.get(oi.getB())) {
                         String label = annotation.getFeatureValueAsString(feat);
                         for (AnnotationSuggestion suggestion : group) {
-                            if (!aLayer.isAllowStacking()
+                            if ((!aLayer.isAllowStacking() && label != null)
```
reckart: `label != null` appears again in a conjunction in the next line. This redundancy can be removed by joining the two conditions.

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinkerTraits.java
jcklie: Done

inception-concept-linking/src/main/java/de/tudarmstadt/ukp/inception/conceptlinking/recommender/NamedEntityLinker.java
jcklie: Done

inception-recommendation/src/main/java/de/tudarmstadt/ukp/inception/recommendation/service/RecommendationServiceImpl.java
jcklie: Done

jcklie:
@reckart The tests fail because the overlapping code does not want to show recommendations for annotations that have null labels. I like the behaviour though and this PR needs it I think, What do you think?

reckart:
@jcklie I adjusted the visibility logic once more. I was getting recommendations with empty labels on existing annotations with a label. Accepting such a recommendation would be a no-op. Such no-ops should be hidden.

reckart:
Can you please double-check if it makes sense to you as well and works as you would expect?

jcklie:
@reckart I tested it with the named linker and NER stringmatcher opennlp with annotations without feature and with feature and it worked. Should the empty label be there for the last Nicolas?

image

jcklie:
@reckart After more thinking, I would leave the visibility now as is, I think it was like that before anyway. If you are ok with it, please approve.

reckart:
Fine by me

Issue #1533, opened 2019-11-26
Periodical commit of MTAS index


reckart:
**Describe the refactoring action**
Instead of committing after indexing each and every document, commit the indices periodically, e.g. using a timer. See [here](https://stackoverflow.com/questions/32269632/writing-to-lucene-index-one-document-at-a-time-slows-down-over-time) for further info.

**Expected benefit**
The indexing should become much faster.

PR #1663, opened 2020-04-13
#1533 - Periodical commit of MTAS index

reckart:
**What's in the PR**
- Schedule a delayed commit when a document is indexed (cancel any already scheduled commit if necessary) - this allows bundling commits over time
- Display a message when triggering a search but the search is not available due to a running indexing process
- Improve performances MtasUimaParser by avoiding DB lookups inside the class
- Improved internal search APIs
- Cleaned up code and removed various redundancies
- Use a time-bounded cache for active indexes which expunges and closes indexes after some time
- Close indexes on shutdown

**How to test manually**
* Use the application
* Use in particular the search functionality
* Mind that e.g. changing the layer/feature triggers a re-indexing of the project
* Depending on the number of documents, indexing may be faster or slower - this can be exploited e.g. to test how search behaves while indexing is still in progress

**Automatic testing**
* [ ] PR includes unit tests (some tests were slightly updated due to changed API, but no new tests)

**Documentation**


jcklie:
Search still works, I did not manage to test searching during reindex.

Issue #1669, opened 2020-04-18
Curation won't start properly


LisaEggert:
**Describe the bug**

When I try to load an annotated text (approx. 3200 lines and annotation on 16 layers) the curation mode won't start. By now I have waited for about 2 hours - without any progress.

We have had this problem with an earlier version (it must have been the one before 0.12.0) and was fixed within version 0.12.0. 
Maybe the bug "remigareted" to version we use (0.14.3).
The short answer could be that we simply have to update our installation ;)

Thank you and stay healthy!

**To Reproduce**
1. Go to 'Curation'
2. Click on 'file to open it'
3. See error: Won't open in curation-mode

**Expected behavior**
That file opens in curation mode within a manageable time.

**Screenshots**

**Please complete the following information:**
 - Version and build ID:  INCEpTION -- 0.14.3 (2020-02-05 14:21:11, build 7f794ab99fd2695ac9d5ed831b7e3ea92b511be5) 
 - OS: Windows
 - Browser: firefox

**Additional context**

reckart:
Upstream issue: https://github.com/webanno/webanno/issues/1340

PR #1678, opened 2020-04-28
#1669 - Curation won't start properly

reckart:
**What's in the PR**
- Switch to newer WebAnno version where the slow and buggy "invisible annotations" field has been removed from the curation page

**How to test manually**
* Make an annotation which crosses two sentences
* Finish the document so you can open it in the curation page
* Configure the page to show only one sentence at a time
* Observe that there is no panel "invisible annotations on other pages" where the invisible cross-sentence annotation is mentioned

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Would anybody care to review this PR?

Issue #1675, opened 2020-04-23
Documentation for TEI P5 format is wrong


reckart:
**Describe the bug**
The documentation claims that TEI P5 supports custom layers, but it does not. The documentation needs to be updated and expanded.

**To Reproduce**
See: https://inception-project.github.io//releases/0.15.2/docs/user-guide.html#sect_formats_tei

**Screenshots**

**Please complete the following information:**
 - Version and build ID: 0.15.2

PR #1682, opened 2020-04-28
#1675 Documentation for TEI P5 format is wrong

UWinch:
Corrected the info on TEI format.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


inception-doc/src/main/resources/META-INF/asciidoc/user-guide/formats-tei.adoc
```diff
@@ -28,7 +28,7 @@ For more details about the supported element types, see the link:https://dkpro.g
 | link:https://www.tei-c.org/release/doc/tei-p5-doc/en/html/index.html[TEI P5]
 | yes
 | yes
-| yes
-| UIMA XMI CAS
+| no
+| Lemma, POS (xpos), named entities
```
reckart, line 32: I think for named entities only the "value" feature is supported - not the identifier feature. Since you mention xpos for the POS layer, maybe that should be mentioned here as well.

UWinch: you are right, changed it

Issue #1677, opened 2020-04-28
New workload management page


reckart:
Added a new page for workload management and monitoring. This page should eventually be able to replace the existing monitoring page.

This page contains:

**- A label** with the caption: "Default # of annotations per document"

**- A textbox right after the label,** where the administrator of the project is able to enter a number which defines how often each of the documents in the project should be "finished". When the number is reached, the document will no longer be offered to other annotators.
(Right now, there is no logic behind the textbox)

**- A label** with the caption: "Select an advanced filter"

**- A dropdown** giving the possibility to use a filter for the table

**- A textbox** enabling to enter a specific value matching the filter criteria from the dropdown

**- A  sortable table containing four columns:**

**1. Document name** (self explaining)

**2. Finished** (How often a document was "finished" by a curator, maximum number is the "default # of annotations per document" as described above)

**3. In Progress** (How many curators are working on the document right now) 

**4. Metadata** (Contains only an image, upon clicking a new panel opens containing further metadata of the document, further refinement meetings will clarify which metdata content will be needed to be displayed)
Right now, the page looks like the following:

reckart:
@jcklie Hasn't this already been merged? Then the milestone is 0.17.0, isn't it?

reckart:
And I guess it can also be closed?

PR #1686, opened 2020-05-03
#1677 - New-workload-management-page

christophscheunemann:
**What's in the PR**
* New Monitoring page established in new module "inception-ui-workload-management-and-monitoring"

**How to test manually**
* As Administrator of a Project, simply go to the Monitoring page


inception-app-webapp/pom.xml
```diff
@@ -500,6 +504,12 @@
 	    <groupId>org.webjars.bower</groupId>
 	    <artifactId>jquery.scrollTo</artifactId>
 	</dependency>
+      <dependency>
+          <groupId>de.tudarmstadt.ukp.inception.app</groupId>
```
UWinch, line 508: I think you added the same dependency twice

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/Monitoring.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static de.tudarmstadt.ukp.clarin.webanno.api.WebAnnoConst.PAGE_PARAM_PROJECT_ID;
+
+public class Monitoring extends ApplicationPageBase
+{
+    //Default annotations label and text box
+    public static final String DEFAULTDOCUMENTSLABEL = "default # of documents:";
+    public static final TextField DEFAULTDOCUMENTSTEXTFIELD = new TextField("defaultDocumentsTextField");
```
UWinch, line 55: Please init. wicket components in the constructor because of possible problems with the wicket lifecycle problems. You can make it `private final`.

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/Monitoring.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.wicket.util.string.StringValue;
+import javax.persistence.NoResultException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static de.tudarmstadt.ukp.clarin.webanno.api.WebAnnoConst.PAGE_PARAM_PROJECT_ID;
+
+public class Monitoring extends ApplicationPageBase
```
UWinch, line 51: You can mount pages at a readable url using `@MountPath("monitoring.html")`.

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/Monitoring.java
```diff
@@ -0,0 +1,166 @@
...
+            return Optional.of(projectService.getProject(aProject.toLong()));
+        }
+        catch (NoResultException e)
+        {
+            return Optional.empty();
+        }
+    }
+
+    //Used for creating the menu
+    private List<MenuItem> getMenuItems()
```
UWinch, line 161: Please don't add the left handside menu from the dashboard to the page to stay consistent with other pages. You might want to test that you can use the back button/ dashboard link to navigate back to the dashboard project view.

christophscheunemann:
Thanks for the feedback, everything is now corrected according to your comments. I will make a new commit in the evening

inception-ui-workload-management-and-monitoring/pom.xml
```diff
@@ -0,0 +1,145 @@
+<!--
+  Copyright 2017
```
reckart, line 2: Wrong year :) Applies to all files really.

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.java
```diff
@@ -0,0 +1,158 @@
...
+import org.apache.wicket.markup.html.panel.Panel;
+import org.apache.wicket.model.ResourceModel;
+import org.apache.wicket.request.mapper.parameter.PageParameters;
+import org.apache.wicket.spring.injection.annot.SpringBean;
+import org.apache.wicket.util.string.StringValue;
+
+import org.wicketstuff.annotation.mount.MountPath;
+
+
+@MountPath("MonitoringPage.html")
```
reckart, line 51: We'd use lowercase and not add page. Suggestion: `workload.html`

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.java
```diff
@@ -0,0 +1,158 @@
...
+public class MonitoringPage extends ApplicationPageBase
+{
+
+    //Layout Panel
+    private final Panel table;
+
+    //Default annotations label and textbox
+    private final NumberTextField DEFAULTDOCUMENTSNUMBERTEXTFIELD;
```
reckart, line 59: Please use `_` to separate words in all-capitals identifiers.

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package de.tudarmstadt.ukp.inception.app.ui.monitoring.page;
```
reckart, line 20: I'm getting checkstyle errors in this file because of wrong order of imports. When you build on the command line with Maven, you should be able to see the checkstyle errors as well. Otherwise, you can install a Checkstyle plugin into your IDE as described in the dev setup guide.

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.java
```diff
@@ -0,0 +1,158 @@
...
+
+    //Layout Panel
+    private final Panel table;
+
+    //Default annotations label and textbox
+    private final NumberTextField DEFAULTDOCUMENTSNUMBERTEXTFIELD;
+
+
+    //Columns in the datatable
+    public static final String DOCUMENT = "Document";
```
reckart, line 63: Strings should go into a properties file (cf. https://ci.apache.org/projects/wicket/guide/8.x/single.html#_internationalization_with_wicket)

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.properties
```diff
@@ -0,0 +1,21 @@
+# Copyright 2012
+# Ubiquitous Knowledge Processing (UKP) Lab and FG Language Technology
+# Technische Universit�t Darmstadt
```
reckart, line 3: Bad encoding?

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/page/MonitoringPage.properties
```diff
@@ -0,0 +1,21 @@
+# Copyright 2012
+# Ubiquitous Knowledge Processing (UKP) Lab and FG Language Technology
```
reckart, line 2: Wrong attribution

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/support/DataProvider.java
```diff
@@ -0,0 +1,114 @@
...
+    @Override
+    public Iterator<SourceDocument> iterator(long first, long count)
+    {
+        List<SourceDocument> newList = data;
+
+        //Sorting, check for which column was clicked and return new sorting accordingly
+        Collections.sort(newList, (o1, o2) ->
+        {
+            int dir = getSort().isAscending() ? 1 : -1;
+            if ("Document".equals(getSort().getProperty()))
```
reckart, line 71: Please avoid use of "magic strings". Also, this seems like an internationalizable string - so it probably shouldn't be used for comparison here?

inception-ui-workload-management-and-monitoring/src/main/java/de/tudarmstadt/ukp/inception/app/ui/monitoring/support/MetaImageColumn.java
```diff
@@ -0,0 +1,90 @@
...
+import org.apache.wicket.markup.html.panel.Panel;
+import org.apache.wicket.markup.repeater.Item;
+import org.apache.wicket.model.IModel;
+import org.apache.wicket.request.resource.PackageResourceReference;
+import org.apache.wicket.request.resource.ResourceReference;
+
+public class MetaImageColumn extends AbstractColumn
+{
+
+    private static final ResourceReference ICON_FINISHED = new PackageResourceReference(
```
reckart, line 40: Please add information about origin and license for any media files to the NOTICE file in the module root. For more information, check other modules which come with a NOTICE file. Licenses must be liberal (i.e. no proprietary or strong-copyleft licenses).

pom.xml
```diff
@@ -131,6 +133,9 @@
     <!-- Curation -->
     
     <module>inception-curation</module>
+      <module>untitled</module>
```
reckart, line 136: untitled?

inception-ui-workload-management-and-monitoring/pom.xml
```diff
@@ -0,0 +1,145 @@
...
+        <artifactId>uimaj-core</artifactId>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.commons</groupId>
+        <artifactId>commons-lang3</artifactId>
+      </dependency>
+      <dependency>
+        <groupId>org.jsoup</groupId>
+        <artifactId>jsoup</artifactId>
+        <version>1.11.3</version>
```
reckart, line 73: pom files should not declare versions. All versions should come from the dependenciesManagement section in the pom in the root of the project.

reckart:
The new page does not appear in the left sidebar for the project dashboard for me. When I enter the `MonitoringPage.html` URL manually into the browser, I get an error:

```
Unexpected RuntimeException

Last cause: The component(s) below failed to render. Possible reasons could be that:
	1) you have added a component in code but forgot to reference it in the markup (thus the component will never be rendered),
	2) if your components were added in a parent container then make sure the markup for the child container includes them in <wicket:extend>.

1. [ModalWindow [Component id = modalPanel, page = de.tudarmstadt.ukp.inception.app.ui.monitoring.page.MonitoringPage, path = dataTable:body:rows:1:cells:4:modalPanel, type = org.apache.wicket.extensions.ajax.markup.html.modal.ModalWindow, isVisible = true, isVersioned = false], children =  [ModalPanel [Component id = content]]]
2. [ModalWindow [Component id = modalPanel, page = de.tudarmstadt.ukp.inception.app.ui.monitoring.page.MonitoringPage, path = dataTable:body:rows:2:cells:4:modalPanel, type = org.apache.wicket.extensions.ajax.markup.html.modal.ModalWindow, isVisible = true, isVersioned = false], children =  [ModalPanel [Component id = content]]]
``

File: inception-app-webapp/src/main/java/de/tudarmstadt/ukp/inception/app/menu/MonitoringPageMenuItem.java [hunk 16]
```diff
@@ -29,7 +29,7 @@
 import de.tudarmstadt.ukp.clarin.webanno.security.UserDao;
 import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
 import de.tudarmstadt.ukp.clarin.webanno.ui.core.menu.MenuItem;
-import de.tudarmstadt.ukp.clarin.webanno.ui.monitoring.page.MonitoringPage;
+import de.tudarmstadt.ukp.inception.app.ui.monitoring.page.MonitoringPage;
```
reckart, line 32: Instead of changing the existing MonitoringPageMenuItem, please add a new WorkloadManagementPageMenuItem - I would like to be able to access both at the moment.

reckart:

reckart:
Branch does not build: 

> 17:20:35 Caused by: hudson.remoting.ProxyException: org.apache.maven.project.ProjectBuildingException: Some problems were encountered while processing the POMs:
17:20:35 [ERROR] 'dependencies.dependency.version' for org.jsoup:jsoup:jar is missing. @ de.tudarmstadt.ukp.inception.app:inception-ui-workload-management-and-monitoring:[unknown-version], /home/svc_jenkins/workspace/INCEpTION (GitHub) (PR builder)/git/inception-ui-workload-management-and-monitoring/pom.xml, line 70, column 19

reckart:

reckart:

reckart:

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.html
```diff
@@ -0,0 +1,87 @@
...
+              <div class="form-group form-row" style="flex: 0.4;" wicket:enclosure="filterTextfield">
+                <div class="col-sm-5" style="flex: 0.4;">
+                  <label class="col-form-label">
+                    <select wicket:id="filterDropDown"/>
+                  </label>
+                  <label class="col-form-label">
+                    <wicket:message key="unused"/>
+                  </label>
+                  <label class="col-form-label">
+                    <input wicket:id="Search" type="submit"/>
```
UWinch, line 70: For buttons you can use `class="btn btn-primary"` or `btn-secondary` to match the layout of other buttons in the application.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,456 @@
...
+
+        //Headers of the table
+        List<String> headers = new ArrayList<>();
+        headers.add(getString("Document"));
+        headers.add(getString("Finished"));
+        headers.add(getString("Processed"));
+        headers.add(getString("Metadata"));
+
+        //Data Provider for the table
+        dataProvider = new DataProvider(documentList.getObject(),
```
UWinch, line 225: I think you don't need the `documentList` at all as a instance variable, but can pass the info directly to your DataProvider `dataProvider = new DataProvider(documentService.
                listSourceDocuments(currentProject.getObject()),
            headers, documentService.
            listAnnotationDocuments(currentProject.getObject()));`

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,456 @@
...
+
+
+
+    //Constructor with a project
+    public MonitoringPage(final PageParameters aPageParameters)
+    {
+        super(aPageParameters);
+
+        //top form initialize
+        workload = new MonitoringForm("workload");
```
UWinch, line 155: I think you don't need the class MonitoingForm at all (this conveniently also gets rid of the name-label problem ;) ). You don't use the label or the stored project model as far as I can see. `workload = new Form<>("workload");` should work.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,456 @@
...
+    private @SpringBean DocumentService documentService;
+
+
+    //Default constructor, no project selected (only when workload.html
+    // put directly in the browser without any parameters)
+    public MonitoringPage() {
+        super();
+
+        //top form initialize
+        workload = new MonitoringForm("workload");
```
UWinch, line 133: `workload = new Form<>("workload");` should suffice, see other comment.

reckart:
Thanks for the PR. I have a few comments.

Screenshot 2020-06-23 at 11 22 45

The most important part of the page from my perspective is the status queue / table. Yet, the screen is dominated by the setting and search panels and when I first opened the page, I did in fact not even notice the table at the bottom of the screen. I could imagine several ways of making these smaller or partially or even fully hiding them by default.

I actually see two scrollbars for the table (see screenshot)

It is possible to sort the table by clicking on the header (e.g. Document), but there is no indication in the header saying what the current sort order is (nice to have).

The "finished", "Processed" and "Metadata" columns have the same width as the "Document" column - but they only contain small numbers or an icon - while the document title can be long in comparison. Why not make the document column take the available space and leave the other three columns at their minimum size?

The "Document name" search field does not seem to work. I have a document in the project called "(20012) Ranke.txt", but no matter whether I search for `ranke`, `Ranke`, `.*Ranke.*` or even `%Ranke%` - I never find it.

Not sure how to use "User name" field in the search form. I was trying this out on a very large project with like 45k documents and several hundred users and was foolish enough to open the old monitoring page which sent my system into re-indexing hell... -- good thing is that the new UI seems to be much faster! --  anyway, so I didn't test the user field yet. But I was wondering: how can I see which effect the search for a user has if the table doesn't show user information? Maybe user-information shows up when I start searching by user? Can I enter multiple users to search for or can I search by some kind of username pattern? Not sure.

Also not sure what the "document creation" time searches for. As a project manager, I could imagine to search for documents that were last touched by an annotator or that were touched in a particular time frame -- but why would I care to search by the document creation time?

I assume by "unused documents", you mean documents that were not finished by any annotator yet? Or do you mean documents that were not started by any annotator yet?

The use of the bootstrap "card" styles look better for the area containing the table than for the two settings panels above which look odd compared to other "card"-style UI elements in the application.

inception-workload-dynamic/NOTICE.txt
```diff
@@ -0,0 +1,4 @@
+The following icon sets are licensed under a Creative Commons Attribution 3.0 License.
+Source: http://www.fatcow.com/free-icons
+
+book_open.png
```
reckart, line 4: I'd suggest we use font-awesome instead of icons. We use font-awesome in most places in the UI now.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.html
```diff
@@ -0,0 +1,183 @@
...
+ - Unless required by applicable law or agreed to in writing, software
+ - distributed under the License is distributed on an "AS IS" BASIS,
+ - WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ - See the License for the specific language governing permissions and
+ - limitations under the License.
+-->
+<html xmlns:wicket="http://wicket.apache.org" xmlns="http://www.w3.org/1999/html" lang="">
+<head>
+  <wicket:head>
+    <style type="text/css">
```
reckart, line 22: Do you need the page-specific CSS here? If yes, why?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.html
```diff
@@ -0,0 +1,183 @@
+<!DOCTYPE html>
```
reckart, line 1: Let's rename it from "MonitoringPage" to "DynamicWorkloadManagementPage" or something like that - we already have the "MonitoringPage" - it is a bit confusing.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,524 @@
...
+    public MonitoringPage(final PageParameters aPageParameters)
+    {
+        super(aPageParameters);
+
+        //top form initialize
+        workload = new Form("workload");
+        workload.setOutputMarkupId(true);
+
+        //Get current Project
+        StringValue projectParameter = aPageParameters.get(PAGE_PARAM_PROJECT_ID);
```
reckart, line 138: We do not need to support the case where the project is not yet on this new page. 
If no project is set or the user has no permission to access it, the user should be returned to the default page.
Otherwise, the page can proceed with setting itself up normally.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,524 @@
...
+
+
+
+        //Get value for the project and set it accordingly
+        //TODO get correct value after upstream change
+        defaultNumberDocumentsTextField.setDefaultModel(new CompoundPropertyModel<Integer>(6));
+        defaultNumberDocumentsTextField.setConvertEmptyInputStringToNull(false);
+
+        //add AJAX event handler on changing input value
+        defaultNumberDocumentsTextField.add(new OnChangeAjaxBehavior()
```
reckart, line 223: I think this can be done using our `LambdaAjaxFormComponentUpdatingBehavior`.

reckart: Parsing a string and setting the model value should not be necessary.
If you use the component in a form with a submit button, then saving the value into the model (and converting it) is done by Wicket automatically.
If you use the component outside a form, then just adding the `LambdaAjaxFormComponentUpdatingBehavior` should do the trick. e.g.:

```
            // Just update the selection
            repositoryCombo.add(new LambdaAjaxFormComponentUpdatingBehavior("change"));
```

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/MonitoringPage.java
```diff
@@ -0,0 +1,524 @@
...
+            @Override
+            protected void onUpdate(AjaxRequestTarget ajaxRequestTarget) {
+                ajaxRequestTarget.add(workload);
+            }
+        });
+
+        TextField<String> documentFilterTextField = new TextField("documentFilter",
+                PropertyModel.of(dataProvider, "filter.documentName"), String.class);
+
+        documentFilterTextField.add(new AjaxFormComponentUpdatingBehavior("onchange") {
```
reckart, line 258: Use our `LambdaAjaxFormComponentUpdatingBehavior` since it doesn't require the need to use inner classes and make the code less verbose.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/ImagePanel.java
```diff
@@ -0,0 +1,41 @@
...
+
+package de.tudarmstadt.ukp.inception.workload.dynamic.support;
+
+import org.apache.wicket.markup.html.image.Image;
+import org.apache.wicket.markup.html.panel.Panel;
+import org.apache.wicket.request.resource.ResourceReference;
+
+
+
+public class ImagePanel extends Panel
```
reckart, line 27: When using font-awesome instead of the icon, we can remove this here.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/ModalPanel.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2020
```
reckart, line 2: The class name is not saying very much about the purpose of the class. Maybe something like `WorkloadItemDetailDialogContent`?

pom.xml
```diff
@@ -171,7 +172,7 @@
       <name>jcenter</name>
       <url>https://jcenter.bintray.com</url>
       <snapshots>
-        <enabled>false</enabled>
+        <enabled>true</enabled>
```
reckart, line 175: We don't want to get snapshots from remote repos. If you need a snapshot version of WebAnno, best check it out locally.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/DynamicWorkloadManagementPage.java
```diff
@@ -150,12 +150,12 @@ public DynamicWorkloadManagementPage(final PageParameters aPageParameters)
...
-        columns.add(new LambdaColumn<>(new ResourceModel("Document"), SourceDocument::getName));
-        columns.add(new LambdaColumn<>(new ResourceModel("Finished"), _doc -> dataProvider
+        columns.add(new LambdaColumn<>(new ResourceModel("Document"), getString("Document"), SourceDocument::getName));
+        columns.add(new LambdaColumn<>(new ResourceModel("Finished"), getString("Finished"), _doc -> dataProvider
                 .getFinishedAmountForDocument((SourceDocument) _doc)));
-        columns.add(new LambdaColumn<>(new ResourceModel("Processed"), _doc -> dataProvider
+        columns.add(new LambdaColumn<>(new ResourceModel("Processed"), getString("Processed"), _doc -> dataProvider
                 .getInProgressAmountForDocument((SourceDocument) _doc)));
-        columns.add(new LambdaColumn<>(new ResourceModel("Users"), _doc -> dataProvider
+        columns.add(new LambdaColumn<>(new ResourceModel("Users"), getString("Users"), _doc -> dataProvider
```
reckart, line 158: Why do you add the `getString()` here again?

christophscheunemann: Otherwise the columns are not sortable and initially we wanted them to be sortable

reckart: Ok. But by `getString()` - this method is used to obtain an internationalized string for a given resource bundle key. Why does the sort property depend on the language of the UI? That said, `getString()` must not be used by code which is called from the constructor.

reckart:

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/DynamicWorkloadManagementPage.java
christophscheunemann: I used getString() here, because the sorting in "DataProvider" depends on the column headers strings. And the headers are all internationalzed strings. But if getString() must not be called here, I can also change that by using headers(0), headers(1) ... and so on if you prefer

reckart: As it stands, that won't work either since `headers` is also defined in terms of `getString()`:

```
        List<String> headers = new ArrayList<>();
        headers.add(getString("Document"));
```

One option could be that you create an enum which defines the columns of the table. Btw: `DataProvider` is a very generic name. Would it be possible to use a slightly more self-explanatory name? What data does it provide?

christophscheunemann: Enum would be a nice solution, sure I'll change it to that.
Ok how about "TableContentProvider"?

reckart: That still doesn't say what it provides. It sounds like a generic re-usable class - or rather like a high-level interface. But is that what it is? Maybe something like `AnnotationQueueOverviewDataProvider`(given... the "queue" isn't quite there yet).

jcklie:
@christophscheunemann This pull request is still marked as work in progress. Is it ready for merge?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadPropertiesImpl.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.stereotype.Component;
+
+@Component
+@ConfigurationProperties("workload")
+public class WorkloadPropertiesImpl implements WorkloadProperties
+{
+    private boolean isActive;
+
+    @Override
+    public boolean isWorkloadManagerActive() {
```
jcklie, line 31: Please follow dkpro style, also prefix method args with `a`, e.g. `aActive`. Also, why do you pass a boxed Boolean?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/Filter.java
```diff
@@ -0,0 +1,85 @@
...
+ * limitations under the License.
+ */
+
+package de.tudarmstadt.ukp.inception.workload.dynamic.support;
+
+import java.io.Serializable;
+import java.util.Date;
+
+//Helper class for the Filter
+public class Filter implements Serializable
```
jcklie, line 25: DKpro style

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/TableContentProvider.java
```diff
@@ -0,0 +1,359 @@
...
+   
+    private List<SourceDocument> shownDocuments;
+    private Filter filter;
+
+    private int defaultAnnotations;
+
+    public TableContentProvider(
+        List<SourceDocument> aData,
+        List<String> headers, List<AnnotationDocument> aAllAnnotationDocuments)
+    {
```
jcklie, line 63: Dkpro style, no need for this then

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/TableContentProvider.java
```diff
@@ -0,0 +1,359 @@
...
+                }
+            }
+            else {
+                return 0;
+            }
+        });
+
+
+        //Reset
+        this.shownDocuments = new ArrayList<>();
```
jcklie, line 131: shownDocuments.clear()

christophscheunemann:
Dkpro style issues at some methods fixed, thanks for the hint!
Normally, I use new Arraylist<> instead of clear() due to memory reallocation instead of nulling values, but I changed it to clear() now.
This PR should now be ready to merge

jcklie:
@christophscheunemann  Too many files with unapproved license

jcklie:
@christophscheunemann Still fails, please build on the command line and check which files do not have a license header. The error message tells you where the file listing sources with errors are.

Issue #1715, opened 2020-05-26
AL sidebar is missing


UWinch:
**Describe the bug**
The active learning sidebar is missing.

**To Reproduce**
Open the annotation page, the active learning sidebar is missing.

**Please complete the following information:**
 - Version and build ID: 0.16.0
 - OS: ubuntu
 - Browser: chrome

PR #1716, opened 2020-05-26
#1715 AL sidebar is missing

UWinch:
Active Learning sidebar was no longer shown.

**What's in the PR**
* Fixed AL autoconfiguration to be conditional on recommendersidebar factory instead of the sidebar itself.
* Made RecommendationMetrics conditional on recommenders enabled.

**How to test manually**
* Open the annotation page and check that the AL sidebar is there
* Check e.g. with jconsole that metric on enabled recommenders is calculated
* Test with settings `recommender.enabled=false`-> neither recommendersidebar nor AL sidebar should be shown
* Test with settings `recommender.active-learning.enabled=false`-> recommendersidebar should be shown but not AL sidebar


reckart:
Why is the recommender sidebar a prerequisite for the AL sidebar? Does it not rather depend on the AL service and the sidebar itself?

UWinch:
> Why is the recommender sidebar a prerequisite for the AL sidebar? Does it not rather depend on the AL service and the sidebar itself?

You implemented it here https://github.com/inception-project/inception/pull/1401/files#diff-b8ffb6f8f914b63f944d6a55e59c2719. I assume it's for the case that recommenders are disabled via config switch. In this case AL would also not be possible.

UWinch:
Can be reviewed/merged.

reckart:
@UWinch from my side you can merge it once Jenkins approves.

Issue #1756, opened 2020-07-12
New DB entry for workflow


christophscheunemann:
A new database entry for the workload and workflow manager would be highly encouraged, due to the reason that otherwise the changes between the managers within the application will not be permanently saved. Each project shall have its own entry on which managers to use.

PR #1757, opened 2020-07-12
#1756 - New DB entry for workflow

christophscheunemann:
**What's in the PR**
Extension for the DB. New entries for each project, which are saving which manager is used (workload and workflow).

- Issue: #1756


inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+@Table(name = "workloadandworkflow")
+public class WorkloadAndWorkflow implements Serializable
+{
+    private static final long serialVersionUID = -9087395004474377523L;
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OnDelete(action = OnDeleteAction.CASCADE)
```
jcklie, line 50: Doesnt this cascade mean that if we delete the workflow settings, we delete the project?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private String workloadmanager;
+
+    private String workflow;
+
+    private int defaultnumberofannotations;
+
+    public WorkloadAndWorkflow()
+    {
+        workflowmanager = "default";
```
jcklie, line 69: Please use constants

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+    @OnDelete(action = OnDeleteAction.CASCADE)
+    @ManyToOne
+    @JoinColumn(name = "project", nullable = false)
+    private Project project;
+
+    @Lob
+    @Column(length = 64000)
+    private String traits;
+
+    private String workflowmanager;
```
jcklie, line 59: I would rename it to workflowManagerType or so

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+
+    private String workloadmanager;
```
jcklie, line 61: I would rename it to workloadManagerType or so

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+        workflowmanager = "default";
+        workloadmanager = "default";
+        workflow = "default";
+        defaultnumberofannotations = 6;
+    }
+
+    public WorkloadAndWorkflow(Project aProject, String aWorkflowManager,
+        String aWorkloadManager, String aWorkflow, int aDefaultNumberOfAnnotations)
+    {
+        super();
```
jcklie, line 78: Do you need the super?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    public int getDefaultnumberofannotations()
+    {
+        return defaultnumberofannotations;
+    }
+
+    public void setDefaultnumberofannotations(int aDefaultnumberofannotations)
+    {
+        this.defaultnumberofannotations = aDefaultnumberofannotations;
```
jcklie, line 133: Not this pls

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+    private String workflow;
+
+    private int defaultnumberofannotations;
+
+    public WorkloadAndWorkflow()
+    {
+        workflowmanager = "default";
+        workloadmanager = "default";
+        workflow = "default";
+        defaultnumberofannotations = 6;
```
jcklie, line 72: I find 6 a lot, I would rather say 2 or 3

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowEngineFactoryImplBase.java
```diff
@@ -0,0 +1,74 @@
...
+
+public abstract class WorkloadAndWorkflowEngineFactoryImplBase<T>
+    implements WorkloadAndWorkflowEngineFactory<T>
+{
+    private Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    @Override
+    public T createTraits()
+    {
+        return null;
```
jcklie, line 37: Why null?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowEngineFactoryImplBase.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public T readTraits(WorkloadAndWorkflow aWorkloadAndWorkflow)
+    {
+        if (aWorkloadAndWorkflow.getTraits() == null) {
+            return createTraits();
```
jcklie, line 44: createTraits returns null, so why bother calling it?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+import org.hibernate.annotations.OnDeleteAction;
+
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+
+/**
+ * A persistence object for the workflow and workload properties of each project.
+ */
+@Entity
+@Table(name = "workloadandworkflow")
+public class WorkloadAndWorkflow implements Serializable
```
jcklie, line 42: I do not like the name, I would maybe rename it to WorkloadAssignment or WorkScheduler or so

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+import de.tudarmstadt.ukp.clarin.webanno.api.ProjectService;
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+
+@Component(WorkloadAndWorkflowService.SERVICE_NAME)
+public class WorkloadAndWorkflowServiceImplBase implements WorkloadAndWorkflowService
+{
+    private final EntityManager entityManager;
+    private final ProjectService projectService;
+
+    public WorkloadAndWorkflowServiceImplBase(
```
jcklie, line 36: Why no autowire?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+        List<String> result = entityManager.createQuery(
+            "SELECT workflowmanager " +
+                "FROM WorkloadAndWorkflow " +
+                "WHERE project = :projectID", String.class)
+            .setParameter("projectID", aProject).getResultList();
+
+        if (result.isEmpty()) {
+            entityManager.persist(new WorkloadAndWorkflow(
+                aProject,"Default workload manager","Default monitoring page","detault",6));
+            return "default";
```
jcklie, line 56: Please use a constant

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        entityManager = aEntityManager;
+        projectService = aProjectService;
+    }
+
+    @Override
+    @Transactional
+    public String getWorkflowManager(Project aProject)
+    {
+        List<String> result = entityManager.createQuery(
```
jcklie, line 47: I would use `query.getSingleResult()`

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+            "SELECT workflowmanager " +
+                "FROM WorkloadAndWorkflow " +
+                "WHERE project = :projectID", String.class)
+            .setParameter("projectID", aProject).getResultList();
+
+        if (result.isEmpty()) {
+            entityManager.persist(new WorkloadAndWorkflow(
+                aProject,"Default workload manager","Default monitoring page","detault",6));
```
jcklie, line 55: detault
please use constants and or default constructors

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        List<String> result = entityManager.createQuery(
+            "SELECT workloadmanager " +
+            "FROM WorkloadAndWorkflow " +
+            "WHERE project = :projectID", String.class)
+            .setParameter("projectID", aProject).getResultList();
+
+        if (result.isEmpty()) {
+            entityManager.persist(new WorkloadAndWorkflow(
+                aProject,"Default workload manager","Default monitoring page","detault",6));
```
jcklie, line 74: Please use constants, you really detault/6 all over the place

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+
+    @Override
+    @Transactional
+    public String getWorkloadManager(Project aProject)
+    {
+        List<String> result = entityManager.createQuery(
+            "SELECT workloadmanager " +
+            "FROM WorkloadAndWorkflow " +
+            "WHERE project = :projectID", String.class)
+            .setParameter("projectID", aProject).getResultList();
```
jcklie, line 70: I would use query.getSingleResult()

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+            return result.get(0);
+        }
+    }
+
+    @Override
+    @Transactional
+    public Integer getDefaultAnnotations(Project aProject)
+    {
+        List<Integer> result = entityManager.createQuery(
+            "SELECT defaultnumberofannotations " +
```
jcklie, line 86: This also wants camelCase

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+                aProject,"Default workload manager","Default monitoring page","detault",6));
+            return "default";
+        } else {
+            return result.get(0);
+        }
+    }
+
+    @Override
+    @Transactional
+    public Integer getDefaultAnnotations(Project aProject)
```
jcklie, line 83: I would call that `getDefaultNumberOfAnnotations`

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+    {
+        List<Integer> result = entityManager.createQuery(
+            "SELECT defaultnumberofannotations " +
+                "FROM WorkloadAndWorkflow " +
+                "WHERE project = :projectID", Integer.class)
+            .setParameter("projectID", aProject).getResultList();
+
+        if (result.isEmpty()) {
+            entityManager.persist(new WorkloadAndWorkflow(
+                aProject,"Default workload manager","Default monitoring page","detault",6));
```
jcklie, line 93: Please use constants

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+    @Transactional
+    public String getWorkflowType(Project aProject)
+    {
+        List<String> result = entityManager.createQuery(
+            "SELECT workflow " +
+                "FROM WorkloadAndWorkflow " +
+                "WHERE project = :projectID", String.class)
+            .setParameter("projectID", aProject).getResultList();
+
+        if (result.isEmpty()) {
```
jcklie, line 110: I would use query.getSingleResult()

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+            entityManager.persist(new WorkloadAndWorkflow(
+                aProject,"Default workload manager","Default monitoring page","detault",6));
```
jcklie, line 112: Use constants defined in a single place

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,166 @@
...
+            "UPDATE WorkloadAndWorkflow " +
+                "SET workflow = :workflow " +
+                "WHERE project = :projectID")
+            .setParameter("workflow", aType)
+            .setParameter("projectID", aProject).executeUpdate();
+    }
+
+    @Override
+    @Transactional
+    public void setDefaultAnnotations(int aDefaultAnnotationsNumber, Project aProject)
```
jcklie, line 157: setDefaultNumberOfAnnotations

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflow.java
```diff
@@ -0,0 +1,184 @@
...
+
+    public WorkloadAndWorkflow(Project aProject, String aWorkflowManager,
+        String aWorkloadManager, String aWorkflow, int aDefaultNumberOfAnnotations)
+    {
+        super();
+        project = aProject;
+        workflowmanager = aWorkflowManager;
+        workloadmanager = aWorkloadManager;
+        workflow = aWorkflow;
+        defaultnumberofannotations = aDefaultNumberOfAnnotations;
```
jcklie, line 83: Why do you call it defaultnumberofannotations (camel case missing btw) and not numberOfAnnotations or minNumberOfAnnotations? Default implies for me not setting this value will give the default, but why would we have that per project and not globally?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/settings/WorkflowAndMonitoringPanel.java
```diff
@@ -110,17 +112,16 @@ private void actionConfirm(AjaxRequestTarget aTarget, Form<?> aForm)
         aTarget.addChildren(getPage(), IFeedback.class);
 
         if (monitoringChoices.getDefaultModelObjectAsString().equals(monitoring.get(0))) {
-            workloadProperties.setActive(false);
+            workloadAndWorkflowService.setWorkloadManager("Default monitoring page", project);
         } else {
-            workloadProperties.setActive(true);
+            workloadAndWorkflowService.setWorkloadManager("Workload monitoring page", project);
```
jcklie, line 117: I do not like these strings, please use constants somewhere

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/WorkloadPageMenuItem.java
```diff
@@ -94,10 +91,11 @@ public boolean applies()
 
         // Visible if the current user is a curator or project admin
         User user = userRepo.getCurrentUser();
+
         return (projectService.isCurator(project, user)
             || projectService.isProjectAdmin(project, user))
             && WebAnnoConst.PROJECT_TYPE_ANNOTATION.equals(project.getMode())
-            && workloadProperties.isWorkloadManagerActive();
+            && !workloadAndWorkflowService.getWorkloadManager(project).equals("Default monitoring page");
```
jcklie, line 98: I do not like literal strings, please use constants

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/WorkloadAndWorkflowServiceImplBase.java
jcklie: I would define a default constructor once or at least a function here or in the DAO so that you do not duplicate the default wf manager code everywhere

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkflowState.java
```diff
@@ -15,19 +15,31 @@
...
-/**
- * @deprecated This needs to be replaced by a proper extension point for workload management
- *             strategies as soon as possible.
- */
-@Deprecated
-public interface WorkloadProperties
+public enum WorkflowState implements PersistentEnum
 {
-    boolean isWorkloadManagerActive();
+    DYNAMIC_WORKFLOW("DYNAMIC_WORKFLOW"),
```
jcklie, line 24: Is it possible to move enum somewhere else, e.g. the central/default workflow place? The dynamic workflow module should not define the default one. Also, I think @reckart  does not like enums but would prefer string constants.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkflowState.java
```diff
@@ -15,19 +15,31 @@
...
+    WorkflowState(String aID)
+    {
+        id = aID;
+    }
 
-    void setActive(boolean aActive);
+    @Override
+    public String getId()
+    {
+        return null;
```
jcklie, line 37: why null

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkflowType.java
```diff
@@ -15,24 +15,31 @@
...
-import org.springframework.stereotype.Component;
+import de.tudarmstadt.ukp.clarin.webanno.support.PersistentEnum;
 
-@Component
-@ConfigurationProperties("defaultannotations")
-public class DefaultAnnotationsPropertiesImpl implements DefaultAnnotationsProperties {
+public enum WorkflowType implements PersistentEnum
+{
+    RANDOMIZED_WORKFLOW_TYPE("RANDOMIZED_WORKFLOW_TYPE"),
+    DEFAULT_WORKFLOW_TYPE("DEFAULT_WORKFLOW_TYPE");
```
jcklie, line 25: Is it possible to move enum somewhere else, e.g. the central/default workflow place? The dynamic workflow module should not define the default one. Also, I think @reckart  does not like enums but would prefer string constants because new modules can be added without needing to add another enum entry.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkflowType.java
```diff
@@ -15,24 +15,31 @@
...
+    {
+        id = aID;
+    }
 
     @Override
-    public int getNumberOfDefaultAnnotations() {
-        return numberOfDefaultAnnotations;
+    public String getId()
+    {
+        return null;
```
jcklie, line 37: Why null?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkloadState.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.workload.dynamic.manager.enums;
+
+import de.tudarmstadt.ukp.clarin.webanno.support.PersistentEnum;
+
+public enum WorkloadState implements PersistentEnum
+{
+    WORKLOAD_MONITORING("WORKLOAD_MONITORING"),
+    DEFAULT_MONITORING("DEFAULT_MONITORING");
```
jcklie, line 25: Is it possible to move enum somewhere else, e.g. the central/default workflow place? The dynamic workflow module should not define the default one. Also, I think @reckart  does not like enums but would prefer string constants.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/WorkloadPageMenuItem.java
```diff
@@ -94,10 +92,12 @@ public boolean applies()
 
         // Visible if the current user is a curator or project admin
         User user = userRepo.getCurrentUser();
+
         return (projectService.isCurator(project, user)
             || projectService.isProjectAdmin(project, user))
             && WebAnnoConst.PROJECT_TYPE_ANNOTATION.equals(project.getMode())
-            && workloadProperties.isWorkloadManagerActive();
+            && !workloadAndWorkflowService.getWorkloadManager(project).equals(
```
jcklie, line 99: We normally compare CONSTAT.equals(foo) because it is more likely for foo to be null than CONSTANT

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/WorkloadPageMenuItem.java
reckart: Internationalizable strings should never be used for logic anyway. See e.g. how the `ActionBarExtensionPointImpl`and  `FeatureSupportRegistryImpl` implement an extension points and how they use the `ExtensionPoint_ImplBase` base class. Then also how e.g. `ActionBarExtension` plugs into that extension point and how it uses the `Extension` base class which defines a `getId()` method. Please use these base classes (`ExtensionPoint_ImplBase` and `Extension` unless there is a good reason not to).

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/manager/enums/WorkflowState.java
reckart: We use DB-manageable enums for states. Cf. `SourceDocumentState` and `SourceDocumentStateType`.

reckart: But... this does not seem to be a state. It seems to be a plugin ID. That needs do be done differently because we cannot have a single place where all the plugins are hard-coded. That would go against the idea of having plugins.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/api/WorkloadConst.java
```diff
@@ -0,0 +1,16 @@
+package de.tudarmstadt.ukp.inception.workload.dynamic.api;
+
+public class WorkloadConst
```
jcklie, line 3: I still think that the default constants should not be in the dynamic workload module. @reckart  What do you think?

christophscheunemann: Ahh ok, I thought you just wanted them to be more on "top" instead of inside the "manager" package, but yes maybe something similar to what it is in webanno? (de.tudarmstadt.ukp.clarin.webanno.model)

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/settings/WorkflowAndMonitoringPanel.java
```diff
@@ -33,57 +38,61 @@
...
+
         //Basic form
         Form<Void> form = new Form<>("form");
 
         //Add two possibilities to select for the workflow manager
         workflow = new ArrayList<>();
-        workflow.add("Default workflow manager");
-        workflow.add("Dynamic workflow manager");
+        workflow.add(DEFAULT_WORKFLOW);
+        workflow.add(DYNAMIC_WORKFLOW);
```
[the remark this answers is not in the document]
reckart: The available workflow strategies should be fetched from the Workflow extension point and be displayed here. They should not be hard-coded.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/model/WorkloadAssignment.java
```diff
@@ -0,0 +1,184 @@
...
+import javax.persistence.Table;
+
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+
+/**
+ * A persistence object for the workflow and workload properties of each project.
+ */
+@Entity
+@Table(name = "workload_assignment")
+public class WorkloadAssignment implements Serializable
```
reckart, line 43: already commented on the choice of name earlier. IMHO the table should only contain an primary key, the project ID, the extension point ID and the traits.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/settings/WorkflowAndMonitoringPanel.java
```diff
@@ -110,17 +119,16 @@ private void actionConfirm(AjaxRequestTarget aTarget, Form<?> aForm)
         aTarget.addChildren(getPage(), IFeedback.class);
 
         if (monitoringChoices.getDefaultModelObjectAsString().equals(monitoring.get(0))) {
-            workloadProperties.setActive(false);
+            workloadAndWorkflowService.setWorkloadManager(DEFAULT_MONITORING, project);
```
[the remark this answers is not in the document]
reckart: Why are the choices (internationalizable) Strings anyway and not actual instances of `WorkloadManager` fetched from the registry? Or alternatively a helper type like `WorkloadManagerType` analogous to e.g. the `LayerType` we have for the layer registry:

**de.tudarmstadt.ukp.clarin.webanno.ui.project.layers.LayerDetailForm**:
```
        add(layerTypeSelect = new BootstrapSelect<LayerType>("type") { ...
```

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/DynamicWorkloadManagementPage.java
```diff
@@ -94,14 +92,12 @@
...
     private @SpringBean UserDao userRepository;
     private @SpringBean ProjectService projectService;
     private @SpringBean MenuItemRegistry menuItemService;
     private @SpringBean DocumentService documentService;
-    private @SpringBean WorkloadProperties workloadProperties;
-    private @SpringBean DefaultAnnotationsProperties defaultAnnotations;
+    private @SpringBean WorkloadAndWorkflowService workloadAndWorkflowService;
 
     //Default constructor, no project selected (only when workload.html
     // put directly in the browser without any parameters)
```
reckart, line 103: We use line comments **in** methods. We use JavaDoc to describe what methods do - if what they do is not obvious.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/page/workload/DynamicWorkloadManagementPage.java
```diff
@@ -239,11 +235,9 @@ private void actionShowInfoDialog(AjaxRequestTarget aTarget, IModel<SourceDocume
         if (aProjectParam == null || aProjectParam.isEmpty()) {
```
reckart, line 235: We do not use separator comments like `--- Helper methods ---`. Such methods are private - that's sufficient.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/model/WorkloadAndWorkflowServiceImplBase.java
```diff
@@ -0,0 +1,171 @@
...
+        entityManager.createQuery(
+            "UPDATE WorkloadAssignment " +
+                "SET numberOfAnnotations = :aDefaultAnnotationsNumber " +
+                "WHERE project = :projectID")
+            .setParameter("aDefaultAnnotationsNumber", aDefaultAnnotationsNumber)
+            .setParameter("projectID", aProject).executeUpdate();
+    }
+
+    @Transactional
+    public void createDefaultEntry(Project aProject)
```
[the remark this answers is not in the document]
christophscheunemann: Methods with @Transactional must be overridable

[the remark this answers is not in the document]
reckart: Methods that are only called from other transactional methods do not need to be transactional themselves.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/Workload_Manager.java
```diff
@@ -0,0 +1,108 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @ManyToOne
+    @JoinColumn(name = "project", nullable = false)
+    private Project project;
+
+    @Column(columnDefinition = "VARCHAR(50)")
+    private String extensionPointID;
```
reckart, line 52: workloadType?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/Workload_Manager.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @ManyToOne
+    @JoinColumn(name = "project", nullable = false)
+    private Project project;
+
+    @Column(columnDefinition = "VARCHAR(50)")
```
reckart, line 51: Why only 50 characters?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/Workload_Manager.java
```diff
@@ -0,0 +1,108 @@
...
+    public String getTraits()
+    {
+        return traits;
+    }
+
+    public void setTraits(String aTraits)
+    {
+        traits = aTraits;
+    }
+
```
reckart, line 107: Entity beans should always have an equals/hashCode method which does include the semantic primary key fields but NOT the generated ID field. In this case, the semantic primary key would be the project and the workload type.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/registry/WorkloadRegistryImpl.java
```diff
@@ -0,0 +1,66 @@
...
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.inception.workload.extension.WorkloadExtension;
+import de.tudarmstadt.ukp.inception.workload.model.WorkloadManagementService;
+
+@Component
+public class WorkloadRegistryImpl
+    implements WorkloadRegistry
+{
+    private List<WorkloadExtension> extensions;
+    private @Autowired WorkloadManagementService workloadManagementService;
```
reckart, line 35: Please do not mix constructor injection and field injection unless there is a pressing reason for it. The service should also be injected through the constructor here.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/registry/WorkloadRegistryImpl.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public WorkloadRegistryImpl(
+        @Lazy @Autowired(required = false) List<WorkloadExtension> aExtensions)
+    {
```
reckart, line 39: Please make us of the `ExtensionPoint_ImplBase` base class for this registry unless there is a strong reason not to do it (and if so, please let me know the reason).

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/ProjectWorkloadSettingsPanelFactory.java
```diff
@@ -51,6 +43,6 @@ public String getLabel()
     @Override
     public Panel createSettingsPanel(String aID, final IModel<Project> aProjectModel)
     {
-        return new WorkflowAndMonitoringPanel(aID, aProjectModel);
+        return new WorkloadPanel(aID, aProjectModel);
```
reckart, line 46: `WorkloadSettingsPanel`?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadPanel.java
```diff
@@ -27,63 +27,55 @@
...
-        workflowChoices.setInline(true);
-        workflowChoices.setOutputMarkupId(true);
-        //Set default for the group
-        workflowChoices.setModel(new Model<>(workflowProperties.getProperty()));
-
-        //add them to the form
-        form.add(workflowChoices);
-
         //Add two possibilities to select from the monitoring manager
         monitoring = new ArrayList<>();
```
reckart, line 62: The list of available managers must be obtained from the registry. It cannot be hard-coded here.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadPanel.java
```diff
@@ -27,63 +27,55 @@
...
         monitoringChoices = new BootstrapRadioChoice<>("monitoringRadios",
             new Model<>(getString("defaultMonitoring")), monitoring);
         monitoringChoices.setInline(true);
         monitoringChoices.setOutputMarkupId(true);
 
         //Set default value for the group
-        monitoringChoices.setModel(new Model<>(workloadProperties.getProperty()));
-
+        switch (workloadRegistry.getSelectedExtension(project)) {
+        case("StaticWorkloadExtension"):
```
reckart, line 73: There must not be any concrete list of plugin(IDs) here - everything must be obtained generically through the registry.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadPanel.java
```diff
@@ -27,63 +27,55 @@
...
-        form.add(workflowChoices);
-
         //Add two possibilities to select from the monitoring manager
         monitoring = new ArrayList<>();
-        monitoring.add("Default monitoring page");
-        monitoring.add("Workload monitoring page");
+        monitoring.add(getString("defaultMonitoring"));
+        monitoring.add(getString("workload"));
 
         monitoringChoices = new BootstrapRadioChoice<>("monitoringRadios",
```
reckart, line 66: workloadStrategy?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadPanel.java
```diff
@@ -27,63 +27,55 @@
...
     private final List<String> monitoring;
     private final BootstrapRadioChoice<String> monitoringChoices;
+    private final Project project;
+
+    private @SpringBean WorkloadManagementService workloadManagementService;
+    private @SpringBean WorkloadRegistry workloadRegistry;
 
-    public WorkflowAndMonitoringPanel(String aID, IModel<Project> aProject) {
+    @Autowired
+    public WorkloadPanel(String aID, IModel<Project> aProject)
```
reckart, line 52: I would expect there to be a selection of the workload type and depending on which workload is selected, a traits panel which shows the editor for the traits of the selected workload type. 

I would expect the workload type selection widget uses list of available workload managers from the workload manager registry as its choice model - similar to how it is done in the WebAnno FeatureDetailForm lines 133-183 there the `featureType` dropdown is constructed.

I would also expect that there is a `WorkoadManager` entity bean used as a model here somewhere. I.e. one in which the selected workload type is saved and where also the traits are saved by the nested traits panel.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadPanel.java
```diff
@@ -96,29 +88,16 @@ public WorkflowAndMonitoringPanel(String aID, IModel<Project> aProject) {
...
 
-        if (monitoringChoices.getDefaultModelObjectAsString().equals(monitoring.get(0))) {
-            workloadProperties.setActive(false);
-        } else {
-            workloadProperties.setActive(true);
-        }
-
-        if (workflowChoices.getDefaultModelObjectAsString().equals(workflow.get(0))) {
-            workflowProperties.setActive(false);
+        if (monitoringChoices.getDefaultModelObjectAsString().equals(getString("workload"))) {
```
reckart, line 95: This way of obtaining the extensions goes away once the workload selection widget is constructed properly - see previous comment.

inception-workload/src/main/resources/de/tudarmstadt/ukp/inception/workload/model/db-changelog.xml
```diff
@@ -0,0 +1,32 @@
...
+      </column>
+        <column name="extensionPointID" type="VARCHAR(50)">
+            <constraints nullable="false" />
+        </column>
+        <column name="traits" type="CLOB(64000)"></column>
+    </createTable>
+    <createIndex indexName="FK_project_id" tableName="workload_manager">
+      <column name="project" />
+    </createIndex>
+    <addForeignKeyConstraint constraintName="FK_project_id"
```
reckart, line 26: Assuming we only want to have a single workflow manager per project, there should probably be a unique constraint over the `project` column here, right? That would hard-prevent that a second workflow manager configuration for a project would be created.

inception-workload/src/main/resources/de/tudarmstadt/ukp/inception/workload/model/db-changelog.xml
```diff
@@ -0,0 +1,32 @@
...
+      <column autoIncrement="true" name="id" type="BIGINT">
+        <constraints primaryKey="true" />
+      </column>
+      <column name="project" type="BIGINT">
+        <constraints nullable="false" />
+      </column>
+        <column name="extensionPointID" type="VARCHAR(50)">
+            <constraints nullable="false" />
+        </column>
+        <column name="traits" type="CLOB(64000)"></column>
```
reckart, line 21: Please compare your changelog file to the others in the project. We use "LONGTEXT" for traits elsewhere.

reckart (dismissed):

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/WorkloadManagementServiceImplBase.java
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            String result = entityManager.createQuery(
+                "SELECT extensionPointID " +
+                    "FROM Workload_Manager " +
+                    "WHERE project = :projectID", String.class)
+                .setParameter("projectID", aProject).getSingleResult();
+            return result;
+        } catch (NoResultException e) {
+            createDefaultEntry(aProject,"StaticWorkloadExtension",null);
+            return "StaticWorkloadExtension";
```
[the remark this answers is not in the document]
christophscheunemann: Yes that sounds good to me

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/Workload_Manager.java
christophscheunemann: Should be enough for anything to come. Of cource, i can put a VARCHAR(255) however I do not think this might be necessary.

reckart: Normally, we use VARCHAR(255). I'd tend to stick with that although we have a bit of trouble with it with MySQL:

> The difference between utf8 and utf8mb4 charset is the bytes requires to store each characters. utf8 requires 3 bytes, while utf8mb4 requires 4 bytes. This means using utf8mb4 charset in a table with innodb engine with innodb_large_prefix disabled, at most 191 characters in a string column must be used.

jcklie:
@christophscheunemann Build failed due to Javadoc

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/extension/WorkloadExtension.java
```diff
@@ -15,16 +15,16 @@
...
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.clarin.webanno.support.extensionpoint.Extension;
 
-    void setNumberOfDefaultAnnotations(int aAnnotations);
+public interface WorkloadExtension extends Extension<Project>
+{
+    @Override
+    default boolean accepts(Project project)
+    {
+        return true;
```
jcklie, line 28: Why do you use a default?

jcklie (approved):
I left some comments, otherwise it looks ok.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/extension/WorkloadExtension.java
christophscheunemann: Otherwise I have to implement accepts() in all the other ExtensionPoints, and they all would have a "return = true" in them, so I thought it would be smoother to implement it only once in the interface

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/workflow/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -76,8 +73,8 @@ public int getPriority()
     }
 
     @Override
-    public boolean accepts(AnnotationPageBase aPage) {
-        return workflowProperties.isWorkflowManagerActive();
+    public boolean accepts (AnnotationPageBase aPage) {
+        return false;
```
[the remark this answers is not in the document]
christophscheunemann: Was left to be changed in #1777, but I fixed it now for this PR too

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/workflow/DynamicWorkflowActionBarExtension.java
```diff
@@ -61,9 +59,7 @@ public int getPriority()
     @Override
     public boolean accepts(AnnotationPageBase aPage)
     {
-        // New dynamic workflow only used when the new workflow manager selected in the settings.
-        // Otherwise use the default one and skip this
-        return workflowProperties.isWorkflowManagerActive();
+        return false;
```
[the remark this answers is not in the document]
christophscheunemann: Same as for the NavigationActionBarExtension, I fixed it now instead of in #1777

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/config/DynamicWorkloadManagerAutoConfiguration.java
```diff
@@ -20,32 +20,20 @@
...
-import de.tudarmstadt.ukp.inception.workload.dynamic.page.workload.WorkloadPageMenuItem;
-
-//@AutoConfigureAfter(WorkloadServiceAutoConfiguration.class)
+import de.tudarmstadt.ukp.inception.workload.dynamic.extension.DynamicWorkloadExtension;
 
 @Configuration
-@ConditionalOnProperty(prefix = "workload.dynamic", name = "enabled", havingValue = "true")
+@Order(300)
+@ConditionalOnProperty(prefix = "workload.dynamic", name = "enabled", havingValue = "true",
+    matchIfMissing = true)
```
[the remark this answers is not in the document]
christophscheunemann: Sorry I missed that, ofc its not enabled by default

jcklie (commented):
When I try to load the dynamic workload extension via `workload.dynamic.enabled=true` in my config file then it does not load. Does it work for you?

1. It would be nice if only the static workload manager is enabled, then the workload tab goes away
2. The labels for the workload strategies are ugly with the underscore
3, It would be nice if you would either add a short description for the selected strategy or add a (i) information icon that links to the docu that has to be written

jcklie (changes requested):
When I select the dynamic workflow manager and save, leave the tab and come back then it is not saved. Also, there is none selected when I open the workload tab, the static one should be selected

christophscheunemann:
@jcklie Fixed the link, forgot about it sorry, was only for testing, Fixed your other request with the saving (indeed looks a lot better now thanks). Build issue havn't been on local mvn verify, however fixed the issue shown in the log

jcklie (changes requested):
The default is now dynamic for all projects for me, that should be static

christophscheunemann:
Done, static now default

jcklie (changes requested):
When I activate the dynamic workflow via application properties, select a project and choose dynamic, then disable dynamic in the application properties and restart, then the project still has dynamic enabled. I think that is not intended.

christophscheunemann:
@jcklie Bug fixed

jcklie:
That is not really a fix, if I add a new workflow provider then it breaks.

christophscheunemann:
@reckart A lot better solution with your idea using the registry, however: Reload is neeeded, because the change is not "registered" on first loading.
Putting it in the "WorkloadPageMenuItem" instead of the "DashboadMenu" will not help as the monitoring page wouldn't be rendered then, same the other way around.

jcklie:
@christophscheunemann Does the dynamic workload is found via autoconfiguration when you enable it via the property? For me it does not...

christophscheunemann:
@jcklie, yes indeed with workload.dynamic.enabled=true in the application.properties its found here. Strange .. I can do a screen share tomorrow to show you what issue I meant for Richard if you like and it still does not work for you

reckart:
Alright, I hope to have addressed now most of the things I had previously ask to be addressed and additionally a few that I just noted while digging deeper into it (e.g. that the settings were in the dynamic module instead of the workload module). I didn't look into the traits and traits UI because we hardly have traits right now. Actually we have the "number of annotators" setting for the dynamic strategy, but properly representing this as a trait then needs to be done in the next PR.

@christophscheunemann can you please familiarize yourself with the changes I made here - you'll need to face them when merging the branch into your follow-up branch.

@UWinch care you give it a bit of a test?

christophscheunemann:
@reckart thanks for the changes, looks very good. Just made a quickfix for the jenkins build error

jcklie (approved):
It works for me, disabling also works after using dynamic

reckart (approved):

Issue #1747, opened 2020-06-23
Curation sidebar causes very large projects to hang


reckart:
**Describe the bug**
Curation sidebar causes very large projects to hang.

**To Reproduce**
1. Import the "tiny" project
2. Activate the new curation sidebar plugin
3. Open the annotation page (no need to actually open the curation sidebar)
4. Finish a document (I think that's what triggered it)
5. ... or switch to another document

From the screenshot taken in VisualVM, it looks like the DB is being hammered.

**Expected behavior**
Curation sidebar should not hammer the DB - in particular if it has never even been opened and curation mode isn't active.

**Screenshots**
![Screenshot 2020-06-23 at 11 58 00](https://user-images.githubusercontent.com/1410238/85390297-ec014380-b548-11ea-8963-4434cccca750.png)

**Please complete the following information:**
 - Version and build ID: 0.16.1 (I guess - I found this on the workflow page branch actually)

PR #1762, opened 2020-07-17
#1747 Curation sidebar causes very large projects to hang

UWinch:
**What's in the PR**
Don't overwrite `onConfigure` for visibility of components, use behaviours instead to avoid calling `getModelObject` which loads the components' model each time.

**How to test manually**
* Have the curation sidebar enabled with `curation.sidebar.enabled=true` in the settings.properties file
* Go to annotation page
* Open and close the curation sidebar once
* Switch to the next document 

**Automatic testing**
* [x] PR updates unit tests


reckart:
Screenshot 2020-06-23 at 11 58 00

If you check the profiling screenshot in the initial comment, you can see that a problem is caused by `listUsers` calling `userRepository.getCurrentUser()` and `hasFinishedDoc` for each user. So `userRepository.getCurrentUser()` does a DB access. Then also, each time `hasFinishedDoc` is called, two database accesses are done using `existsAnnotationDocument` and then `getAnnotationDocument`. So let's say we have 300 annotators on a project, then we make 900 database accesses here.

Btw. I can see that there are many cases where `getCurrentUser()` is called just to get the current users username. Maybe we should have a `getCurrentUsername()` which would only need to access the security context and not access the DB at all... that might speed of a few things.

UWinch:

jcklie:
@UWinch Did you fix the looped DB access?

UWinch:
I changed the loop in `CurationSidebar` line 351 to line 338 which calls one db query now via ` curationService.listFinishedUsers` .

jcklie:
Works for me

Issue #1771, opened 2020-07-24
Hide Agreement menu item if not enough users


UWinch:
**Is your feature request related to a problem? Please describe.**
Same as https://github.com/webanno/webanno/issues/1735, but for Inception.

PR #1772, opened 2020-07-24
#1771 Hide Agreement menu item if not enough users

UWinch:
Same as https://github.com/webanno/webanno/issues/1735

**What's in the PR**
* check in agreement menuitem if there are more than one annotators in the project

**How to test manually**
* Be the only user in a project
* Open the dashboard, you should not see the agreement menu item
* Add another user to the project, you should now see the agreement item


jcklie:
@UWinch Build fails

UWinch:
That's because of an upstream change to `ProjectServiceImpl`, should be fixed when https://github.com/inception-project/inception/pull/1762 is merged.

Issue #1776, opened 2020-07-28
Advanced workload and workflow


christophscheunemann:
**Is your feature request related to a problem? Please describe.**
New features for the workload page, to give the project manager more flexibility and possibilities.
New features will be
1. Change workflow on how documents will be given for annotation (ordered / random) from the settings dropdown in the workload page
2. Actual logic implementation for the default number of annotations

PR #1777, opened 2020-07-28
#1776-Advanced-workload-and-workflow

christophscheunemann:
**What's in the PR**
Implementation of #1776. Advanced features for the workload and workflow

**How to test manually**
Go to the workload page and check the new features


reckart:
@christophscheunemann lots of conflicts to resolve now that we merged the other branch :)

christophscheunemann:
@reckart Thanks for the merge! Unfortunately I am not allowed to hit the "resolve conflicts" button for this PR, can you grant me the required rights please?

jcklie:
You need to merge on the commandline

christophscheunemann:
Ahh ok, last time it was also here possible, but thanks

jcklie:
The conflics are now too complicated

reckart:
Why are there JAR files and hidden binaries files in the PR?

christophscheunemann:
@reckart /lib with jar files removed
@jcklie Merge conflicts resolved
@UWinch Fixed issue #1813, can be closed

UWinch:
Can you please open a new PR just for the bugfix? Then we can quickly merge that one.

christophscheunemann:
@UWinch yes sure

reckart:
What is the file ".java.security.swp" in this PR?

inception-app-webapp/src/main/resources/application.properties
```diff
@@ -65,4 +65,4 @@ wicket.stuff.htmlcompressor.features.removeComments=true
 wicket.stuff.htmlcompressor.features.removeIntertagSpaces=true
 wicket.stuff.htmlcompressor.features.preserveLineBreaks=true
 wicket.stuff.htmlcompressor.features.compressJavaScript=false
-wicket.stuff.htmlcompressor.features.compressCss=false
+wicket.stuff.htmlcompressor.features.compressCss=false
```
reckart, line 68: Please undo whitespace-only changes

inception-workload-dynamic/pom.xml
```diff
@@ -27,104 +27,120 @@
...
-      </dependency>
-      <dependency>
-        <groupId>org.springframework.boot</groupId>
-        <artifactId>spring-boot-autoconfigure</artifactId>
-      </dependency>
-      <dependency>
-        <groupId>com.googlecode.wicket-jquery-ui</groupId>
-        <artifactId>wicket-kendo-ui</artifactId>
-      </dependency>
+        <dependency>
```
reckart, line 30: Please undo whitespace-only changes

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowActionBarExtension.java
```diff
@@ -24,26 +24,29 @@
...
 import de.tudarmstadt.ukp.clarin.webanno.ui.annotation.DefaultWorkflowActionBarExtension;
+import de.tudarmstadt.ukp.inception.workload.dynamic.model.DynamicWorkflowManagementService;
 import de.tudarmstadt.ukp.inception.workload.model.WorkloadManagementService;
 
-@Component
 public class DynamicWorkflowActionBarExtension
     implements ActionBarExtension
 {
     private final WorkloadManagementService workloadManagementService;
+    private final DynamicWorkflowManagementService dynamicWorkflowManagementService;
```
reckart, line 38: As far as I can tell, we should not have the `DynamicWorkflowManagementService` since we already have the `WorkloadManagementService`. They are redundant. The specific settings go into the traits field of the `WorkloadManager` entities.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/workflow/WorkflowManagerExtension.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.workload.dynamic.workflow;
+
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.clarin.webanno.support.extensionpoint.Extension;
+
+public interface WorkflowManagerExtension extends Extension<Project>
```
reckart, line 23: We already have the `WorkloadManagerExtension`, so this seems to be redundant.

reckart (changes requested):
It seems there is quite a bit of redundant code now (see line comments) - mind, I didn't comment all the redundant stuff. Could you please remove the redundancy dropping the new redundante files added in this PR. That will facilitate reviewing the rest.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -44,37 +51,42 @@
...
 
-public class DynamicAnnotatorWorkflowActionBarItemGroup extends Panel
+public class DynamicAnnotatorWorkflowActionBarItemGroup
+    extends Panel
 {
     private static final long serialVersionUID = -292514874000914541L;
 
     private @SpringBean DocumentService documentService;
     private @SpringBean ProjectService projectService;
     private @PersistenceContext EntityManager entityManager;
```
reckart, line 63: There should be no entity managers in UI components. Please remove.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +101,91 @@ public String getStateClass()
...
                     return;
                 }
             }
 
-            //Check if no new document has been selected, return to homepage
-            if (getAnnotationPage().getModelObject().getDocument().equals(document)) {
-                getAnnotationPage().setResponsePage(getAnnotationPage().getApplication().
-                    getHomePage());
-                getSession().info("There are no more documents to annotate available for you. Please contact your project supervisor.");
+            if ((workloadManagementService.getTraits(project).split(",")[0]).equals("randomized")) {
```
reckart, line 132: Traits are by convention JSON objects and we use a common pattern for accessing them consisting of the methods `readTraits` and `writeTraits` (cf. e.g. the LayerSupport_ImplBase or FeatureSupport and its implementations). And we do not create the JSON manually, but rather from a `XXXTraits` (c.f. StringFeatureTraits) class using Jackson. I believe your case, these methods should go into the DynamicWorkloadExtension and MatrixWorkloadExtension classes. The readTraits method take a database entity with a traits field, read the field, parse it into an extension-specific data object, and returns it. The writeTraits does it the other way around. The database entity should be obtained from the DAO class - which in your case would be WorkloadManagementService.getOrCreateWorkloadManagerConfiguration() and after a change, it should be written using the corresponding write method in the DAO, so in your case WorkloadManagementService.setWorkloadManagerConfiguration()

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +97,86 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
-                    Optional<AjaxRequestTarget> target = RequestCycle.get().
-                        find(AjaxRequestTarget.class);
-                    aPage.actionLoadDocument(target.orElse(null));
-                    return;
+            if ((workloadManagementService.getTraits(project).split(",")[0]).equals("randomized")) {
+                // Go through all documents in a random order and check if there
+                // is a Annotation document with the state NEW
+                String query = "FROM SourceDocument " + "WHERE project = :project "
+                        + "ORDER BY rand()";
+                for (SourceDocument doc : entityManager.createQuery(query, SourceDocument.class)
```
reckart, line 127: See comment on entitymanager and traits before.

reckart (changes requested):

christophscheunemann:
@reckart Curators no more use the new workload as requested
@jcklie Inital implementation for your upcoming Curriculum Annotation, added TODO in DynamicWorkflowDocumentNavigationActionBarExtension and DynamicAnnotatorWorkflowActionBarItemGroup for the logic on the document selection. Selecting of the feature possible in the workload page

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
-            if (getAnnotationPage().getModelObject().getDocument().equals(document)) {
-                getAnnotationPage().setResponsePage(getAnnotationPage().getApplication().
-                    getHomePage());
-                getSession().info("There are no more documents to annotate available for you. Please contact your project supervisor.");
+            WorkloadManager currentWorkload = workloadManagementService
+                    .getOrCreateWorkloadManagerConfiguration(project);
+
+            WorkloadTraits traits = dynamicWorkloadExtension.readTraits(currentWorkload);
+            switch (traits.getWorkloadType()) {
+            case ("Randomized workflow"):
```
reckart, line 138: As explained numerous times before: there must not be any internationalizable strings used in logic.

If you want to support different workflow types as an extension to the dynamic workflow manager, then the dynamic workflow manager must define an extension mechanism in a similar way as we defined an extension mechanism for workflows in general.

@jcklie Could you please keep an eye on this?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                // Go through all documents in a random order and check if there
+                // is a Annotation document with the state NEW
+                List<SourceDocument> randomList = documentService.listSourceDocuments(project);
+                Collections.shuffle(randomList);
+                for (SourceDocument doc : randomList) {
+                    if (getUsersWorkingOnTheDocument(doc) < traits
+                            .getDefaultNumberOfAnnotations()) {
+                        if (documentService.listAnnotatableDocuments(project, user).get(doc) == null
+                                || (documentService.listAnnotatableDocuments(project, user).get(doc)
+                                        .getState().equals(NEW))) {
```
reckart, line 148: Never do `variable.equals(CONSTANT)`, always do `CONSTANT.equals(variable)` - the latter is null-safe, the former is not.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                            getAnnotationPage().getModelObject().setDocument(doc,
+                                    documentService.listSourceDocuments(project));
+                            Optional<AjaxRequestTarget> target = RequestCycle.get()
```
reckart, line 151: You have `_target` - why fetch the target from the request cycle?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
 
-            //Check if no new document has been selected, return to homepage
-            if (getAnnotationPage().getModelObject().getDocument().equals(document)) {
-                getAnnotationPage().setResponsePage(getAnnotationPage().getApplication().
-                    getHomePage());
-                getSession().info("There are no more documents to annotate available for you. Please contact your project supervisor.");
+            WorkloadManager currentWorkload = workloadManagementService
+                    .getOrCreateWorkloadManagerConfiguration(project);
+
+            WorkloadTraits traits = dynamicWorkloadExtension.readTraits(currentWorkload);
```
reckart, line 136: Instead of having a generic `WorkflowTraits` for all workflow types, there should be a `DynamicWorkflowTraits` and a `MatrixWorkflowTraits`, each with the settings required by **the particular workflow strategy**.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+            switch (traits.getWorkloadType()) {
+            case ("Randomized workflow"):
+                // Go through all documents in a random order and check if there
+                // is a Annotation document with the state NEW
+                List<SourceDocument> randomList = documentService.listSourceDocuments(project);
+                Collections.shuffle(randomList);
+                for (SourceDocument doc : randomList) {
+                    if (getUsersWorkingOnTheDocument(doc) < traits
+                            .getDefaultNumberOfAnnotations()) {
+                        if (documentService.listAnnotatableDocuments(project, user).get(doc) == null
```
reckart, line 146: Instead of having nested IF clauses, use a AND conjunction.

Calling `listAnnotatableDocuments` in a loop is **terribly slow** and definitely not a good idea. The user and project at this point in the code is fixed, so you can call this once outside the loop...

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+
+            WorkloadTraits traits = dynamicWorkloadExtension.readTraits(currentWorkload);
+            switch (traits.getWorkloadType()) {
+            case ("Randomized workflow"):
+                // Go through all documents in a random order and check if there
+                // is a Annotation document with the state NEW
+                List<SourceDocument> randomList = documentService.listSourceDocuments(project);
+                Collections.shuffle(randomList);
+                for (SourceDocument doc : randomList) {
+                    if (getUsersWorkingOnTheDocument(doc) < traits
```
reckart, line 144: I have the feeling calling this in a loop will also be terribly slow.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
-                getSession().info("There are no more documents to annotate available for you. Please contact your project supervisor.");
+            WorkloadManager currentWorkload = workloadManagementService
+                    .getOrCreateWorkloadManagerConfiguration(project);
+
+            WorkloadTraits traits = dynamicWorkloadExtension.readTraits(currentWorkload);
+            switch (traits.getWorkloadType()) {
+            case ("Randomized workflow"):
+                // Go through all documents in a random order and check if there
+                // is a Annotation document with the state NEW
+                List<SourceDocument> randomList = documentService.listSourceDocuments(project);
```
reckart, line 141: Do we really have to go through **all** the source documents?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                            getAnnotationPage().getModelObject().setDocument(doc,
+                                    documentService.listSourceDocuments(project));
+                            Optional<AjaxRequestTarget> target = RequestCycle.get()
+                                    .find(AjaxRequestTarget.class);
+                            getAnnotationPage().actionLoadDocument(target.orElse(null));
+                            return;
+                        }
+                    }
+                }
+                //No documents left
```
reckart, line 158: The comment doesn't explain nothing.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                    // First check if too many users are already working on the document
+                    if (((getUsersWorkingOnTheDocument(entry.getKey())
+                            + 1) <= dynamicWorkloadExtension.readTraits(currentWorkload)
+                                    .getDefaultNumberOfAnnotations())) {
+                        // Now check if there either is no annotation document yet created or its
+                        // state is NEW
+                        if (entry.getValue() == null || entry.getValue().getState().equals(NEW)) {
+                            getAnnotationPage().getModelObject().setDocument(entry.getKey(),
+                                    documentService.listSourceDocuments(project));
+                            Optional<AjaxRequestTarget> target = RequestCycle.get()
```
reckart, line 184: See other comment on _target.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                // Default, simply go through the list and return the first document
+                for (Map.Entry<SourceDocument, AnnotationDocument> entry : documentService
+                        .listAnnotatableDocuments(project, user).entrySet()) {
+                    // First check if too many users are already working on the document
+                    if (((getUsersWorkingOnTheDocument(entry.getKey())
+                            + 1) <= dynamicWorkloadExtension.readTraits(currentWorkload)
+                                    .getDefaultNumberOfAnnotations())) {
+                        // Now check if there either is no annotation document yet created or its
+                        // state is NEW
+                        if (entry.getValue() == null || entry.getValue().getState().equals(NEW)) {
```
reckart, line 181: See other comment on usage of constants.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                            return;
+                        }
+                    }
+                }
+                //No documents left
+                noDocumentsLeft();
+                break;
+
+
+            case ("Curriculum annotation"):
```
reckart, line 163: This should not be here at all...

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+    public int getUsersWorkingOnTheDocument(SourceDocument aDocument)
+    {
+        return (int) documentService
+                .listAnnotationDocuments(getAnnotationPage().getModelObject().getProject()).stream()
+                .filter(d -> d.getDocument().equals(aDocument) && !d.getState().equals(NEW)
+                        && !d.getState().equals(IGNORE))
+                .count();
+    }
+
+    public void noDocumentsLeft()
```
reckart, line 208: Maybe rename to `redirectUserToHomePage`?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -89,48 +102,115 @@ public String getStateClass()
...
+                noDocumentsLeft();
+                break;
             }
         });
         finishDocumentDialog.show(aTarget);
     }
+
+    public int getUsersWorkingOnTheDocument(SourceDocument aDocument)
+    {
+        return (int) documentService
```
reckart, line 201: Instead of fetching all stuff from the DB and then filtering on it, implement a new query method which does the filtering already on the SQL level. If necessary open a corresponding PR on WebAnno.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +105,106 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
     @Override
     public void onInitialize(AnnotationPageBase aPage)
     {
         User user = aPage.getModelObject().getUser();
         Project project = aPage.getModelObject().getProject();
-        //Check if there is a document in progress and return this one
-        for (AnnotationDocument annotationDocument:
-            documentService.listAnnotationDocuments(project,user)) {
-            //There was one in progress, load it
+        annotationDocumentList = documentService.listAnnotationDocuments(project, user);
```
reckart, line 114: This is slow. Better implement a method in the document service which already filters by state at the SQL level.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +105,106 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
-                        getState().equals(NEW)) {
-                    //This document had the state NEW, load it
-                    aPage.getModelObject().setDocument(doc, documentService.
-                        listSourceDocuments(project));
-
-                    Optional<AjaxRequestTarget> target = RequestCycle.get().
-                        find(AjaxRequestTarget.class);
-                    aPage.actionLoadDocument(target.orElse(null));
-                    return;
+            WorkloadManager currentWorkload = workloadManagementService
```
reckart, line 130: See comments I made on DynamicAnnotatorWorkflowActionBarItemGroup.java  - they apply here as well.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -250,10 +265,10 @@ private void actionShowInfoDialog(AjaxRequestTarget aTarget, IModel<SourceDocume
     public List<String> listUsersInProgressForDocument(SourceDocument aDocument)
     {
         List<String> result = new ArrayList<>();
-        for (AnnotationDocument anno: documentService.
-            listAnnotationDocuments(currentProject.getObject())) {
+        for (AnnotationDocument anno : documentService
```
reckart, line 268: Slow - filtering should happen at SQL level in a backend service bean, e.g. documentService.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -229,18 +243,19 @@ private void actionShowInfoDialog(AjaxRequestTarget aTarget, IModel<SourceDocume
...
             return Optional.empty();
         }
     }
 
     public List<String> listUsersFinishedForDocument(SourceDocument aDocument)
     {
         List<String> result = new ArrayList<>();
-        for (AnnotationDocument anno: documentService.
-            listAnnotationDocuments(currentProject.getObject())) {
+        for (AnnotationDocument anno : documentService
```
reckart, line 255: Slow - filtering should happen at SQL level in a backend service bean, e.g. documentService.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -263,56 +278,54 @@ private void actionShowInfoDialog(AjaxRequestTarget aTarget, IModel<SourceDocume
...
+        // Input dates
+        dateFrom = new AjaxDatePicker("from", PropertyModel.of(aProv, "filter.from"), "MM/dd/yyyy");
+        dateTo = new AjaxDatePicker("to", PropertyModel.of(aProv, "filter.to"), "MM/dd/yyyy");
 
         searchForm.add(dateFrom);
         searchForm.add(dateTo);
 
         // Date choices
-        // FIXME: This should use an Enum value as model - and I18N should be used to map the 
+        // FIXME: This should use an Enum value as model - and I18N should be used to map the
```
reckart, line 306: This looks like a case for the EnumChoiceRenderer

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -342,33 +356,253 @@ protected void onUpdate(AjaxRequestTarget ajaxRequestTarget) {
...
-        dateChoices.add(enabledWhen(() ->
-            unused.getValue().equals("false")));
-        userFilterTextField.add(enabledWhen(() ->
-            unused.getValue().equals("false")));
-        documentFilterTextField.add(enabledWhen(() ->
-            unused.getValue().equals("false")));
+        // Condition for filter inputs to be enabled
+        dateTo.add(enabledWhen(
+            () -> !dateChoices.getDefaultModelObjectAsString().equals(dateChoice.get(0))
+                 && unused.getValue().equals("false")));
```
reckart, line 372: Comparing a boolean value against the string value "false" seems like a terrible hack. There must be a better approach to this.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -342,33 +356,253 @@ protected void onUpdate(AjaxRequestTarget ajaxRequestTarget) {
...
+        // Finally, add the confirm button at the end
+        Button save = new LambdaAjaxButton<>("save", this::actionConfirm);
+
+        settingsForm.add(defaultNumberDocumentsTextField);
+        settingsForm.add(save);
+
+        workflowChoices = new BootstrapSelect<>("workloadStrategy");
+        workflowChoices.setDefaultModel(Model.of(traits.getWorkloadType()));
+        List<String> choices = new ArrayList<>();
+        choices.add(getString("def"));
```
reckart, line 411: Not good. Looks like you want to have an extensible list here, so you have to introduce an extension point where these options can be plugged in...

@jcklie Can you please have an eye on this?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/WorkloadManagementServiceImplBase.java
```diff
@@ -33,7 +33,8 @@
  * {@link WorkloadManagementAutoConfiguration#workloadManagementService}
  * </p>
  */
-public class WorkloadManagementServiceImplBase implements WorkloadManagementService
+public class WorkloadManagementServiceImplBase
```
reckart, line 36: This should be `WorkloadManagementServiceImpl`

inception-workload-matrix/src/main/java/de/tudarmstadt/ukp/inception/workload/matrix/MatrixWorkloadExtension.java
```diff
@@ -29,19 +36,57 @@
...
-    
+
     @Override
     public String getLabel()
     {
         return "Static assignment";
     }
+
+    @Override
+    public WorkloadTraits readTraits(WorkloadManager aWorkloadManager)
```
reckart, line 59: This should return a MatrixWorkloadTraits

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -384,4 +618,92 @@ private void actionReset(AjaxRequestTarget aTarget, Form<?> aForm)
...
+            Form<List<SourceDocument>> aForm)
+    {
+        aAjaxRequestTarget.addChildren(getPage(), IFeedback.class);
+
+        if (aForm.getModelObject() == null) {
+            error("Please add documents you want to assign.");
+        }
+        else {
+
+            for (SourceDocument sourceDocument : aForm.getModelObject()) {
```
reckart, line 678: The body of this loop looks very much like backend code that should go into a service bean...

reckart (changes requested):
General notes: I suggested in some comments to add query methods to documentService and to open a corresponding PR in WebAnno. It may be simpler thought to add these methods to the WorkloadManagementService for the time being.

Left various comments in the code. As there are still many places that need to be adjusted, I didn't look at all of them in the highest detail.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
christophscheunemann: Hm I mean it should be random from a list. And it is only the source documents, not the annotation documents for a project. What exactly do you prefer here?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
christophscheunemann: I implemented it the same way as it was done in webanno (e.g. UserSelectionPanel). But I can try to change that ofc

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
reckart: As you put it in the comment "go through all the source documents and check if there is an annotation document in state new" - so why not instead "find annotation documents in state new which are annotatable by the current user" - if possible, then preferably in a single SQL query which would be performed in a backend service and not in the frontend code.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
reckart: I guess the `UserSelectionPanel` is probably one of the oldest pieces of code we have - I don't know how often it was refactored and in particular with respect to performance/scalability. The workflow changes we make here have the goal of supporting a scalable workflow with lots of users and lots of documents, so we need to keep in mind that DB accesses e.g. over source/annotation documents need to be efficient as well.

christophscheunemann: Hehe ok, sure I'll change that

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
christophscheunemann: Alright good idea thanks I'll do it that way

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/WorkloadManagementService.java
```diff
@@ -17,12 +17,32 @@
...
+ * Interface for all required DB calls. Short explanation given in the
+ * Interface implementation class WorkloadManagementServiceImpl
+ */
 public interface WorkloadManagementService
 {
     WorkloadManager getOrCreateWorkloadManagerConfiguration(Project aProject);
 
-    void setWorkloadManagerConfiguration(String workloadType, Project aProject);
-    void setTraits(String aTraits, Project aProject);
+    void setWorkloadManagerConfiguration(String aWorkloadType, Project aProject);
```
reckart, line 36: Normally, we don't have methods for setting individual DB columns in the services. We work on the entities and then persist the entities.

So in this case, we would use `getOrCreateWorkloadManagerConfiguration()`  to obtain a `WorkloadManager` entity, then we would set the workload type and/or traits on that entity, and then we would eventually save the entity.

Typically, we load the entity into a form. Then the user makes modifications on the form. And when the user presses save, we persist the entity that is model of the form.

So in the service beans, we usually have the typically only have methods for loading, listing and saving entities but none that manipulate entities.

Could you please change this so that setWorkloadMangagerConfiguration goes away and we work only with the WorkloadManager entity and methods for loading and saving it?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/AnnotationQueueOverviewDataProvider.java
```diff
@@ -162,39 +172,38 @@ public void detach()
         List<SourceDocument> dateList = new ArrayList<>();
         List<SourceDocument> unusedList = new ArrayList<>();
 
-        //Avoid error in one specific case
+        // Avoid error in one specific case
         if (filter.getSelected() == null) {
-            filter.setSelected("false");
+            filter.setSelected(Boolean.FALSE.toString());
```
reckart, line 177: Why does this have to be a string? Why is it not a boolean field in the Filter class?

reckart (changes requested):

reckart:
@christophscheunemann in terms of the code, it looks pretty good now. I have only a few minor comments.

We still need to look at the actual functionality.

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/model/WorkloadManagementService.java
christophscheunemann: Ahh ok, thanks for the information and description. Sure no problem. I changed it now to your likings.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -87,50 +107,99 @@ public String getStateClass()
...
+            // Get all documents for which the state is NEW, or which have not been created yet.
+            List<SourceDocument> sourceDocuments = workloadManagementService
+                    .getAnnotationDocumentListForUser(project, user);
+
+            // Switch for all workflow types which are available. If a new one is created.
+            // simply add a new case here.
+            switch (dynamicWorkloadExtension.readTraits(currentWorkload).getWorkflowType()) {
+
+            // Go through all documents in a random order
+            case (RANDOMIZED_WORKFLOW):
```
reckart, line 159: Instead of doing a switch here, there should be a method in the `WorkflowExtension` interface like `SourceDocument nextDocument(List<SourceDocument> candidates)` and that method should be used to fetch the next document. The specific implementations like the `RandomizedWorkflowExtension` implement this method accordingly. The menu bar code should not have to know anything about the selection process except that it calls the `nextDocument` method.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -87,50 +107,99 @@ public String getStateClass()
...
     }
+
+    private void loadNewDocument(List<SourceDocument> aSourceDocuments, Project aProject,
+            WorkloadManager aCurrentWorkload, AnnotationPageBase aPage, AjaxRequestTarget aTarget)
+    {
+        // Go through all documents of the list
+        for (SourceDocument doc : aSourceDocuments) {
+            // Check if there are less annotators working on the selected document than
+            // the default number of annotation set by the project manager
+            if ((workloadManagementService.getAmountOfUsersWorkingOnADocument(doc, aProject)
```
reckart, line 183: "number of users" - not "amount"

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -87,50 +107,99 @@ public String getStateClass()
...
+                    + 1) <= (dynamicWorkloadExtension.readTraits(aCurrentWorkload)
```
reckart, line 184: not `+1` and then `<` instead of `<=`

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicAnnotatorWorkflowActionBarItemGroup.java
```diff
@@ -87,50 +107,99 @@ public String getStateClass()
...
+        // No documents left, return to homepage and show corresponding message
+        redirectUSerToHomePage();
+
+    }
+
+    private void redirectUSerToHomePage()
+    {
+        // Nothing left, so returning to homepage and showing hint
+        getAnnotationPage().setResponsePage(getAnnotationPage().getApplication().getHomePage());
+        getAnnotationPage().getSession().info(
```
reckart, line 202: I'd set the info before the redirect. Might not make a difference in practice though.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicDocumentNavigator.java
```diff
@@ -18,10 +18,14 @@
 
 package de.tudarmstadt.ukp.inception.workload.dynamic.annotation;
 
+import java.io.Serializable;
+
 import org.apache.wicket.markup.html.panel.Panel;
 
-public class DynamicDocumentNavigator extends Panel
+public class DynamicDocumentNavigator extends Panel implements Serializable
```
reckart, line 25: Panel already implements serializable. Why implement it again?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowActionBarExtension.java
```diff
@@ -66,15 +66,16 @@ public boolean accepts(AnnotationPageBase aPage)
             return false;
         }
 
+        //Curator are excluded from the feature
```
reckart, line 69: Please remember to document this in the user documentation.

UWinch (commented):
Thanks, I really like the look of the workload page now. I had a few problems when trying it though:
* As a user i'm always getting the `no more available documents` message when trying to annotate. I used the setting 2 annotators per document, default workflow. Explicitly assigning the user works.
* I'm getting a partial update error when trying to finish a document.

Some UX issues I experienced:
* As a new user I do not know what the different workflows are (*randomized*, *external*, *default*). can you add an explanation, maybe as a mouse over or a link to the documentation?
* When I'm assigning a user to a specific document, the window closes after selecting the document before I can save by clicking *assign*
* Assigning a document to a user does not immediately update the table (the annotators column of the respective document), so I'm not sure if the assignment really worked.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -73,16 +95,18 @@ public int getPriority()
...
         if (aPage.getModelObject().getProject() == null) {
             return false;
         }
-
-        return DYNAMIC_WORKLOAD_MANAGER_EXTENSION_ID.equals(workloadManagementService.
-            getOrCreateWorkloadManagerConfiguration(aPage.getModelObject().getProject())
-            .getType());
+        // Curator are excluded from the feature
+        return DYNAMIC_WORKLOAD_MANAGER_EXTENSION_ID
+                .equals(workloadManagementService.loadOrCreateWorkloadManagerConfiguration(
```
reckart, line 106: You can fetch the `aPage.getModelObject()` once, store it in a local variable and then use it. Makes sense if you use a value more than once.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +115,84 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
+        // Get all documents for which the state is NEW, or which have not been created yet.
+        List<SourceDocument> sourceDocuments = workloadManagementService
+                .getAnnotationDocumentListForUser(project, user);
+
+        // Switch for all workflow types which are available. If a new one is created
+        // simply add a new case here.
+        switch (dynamicWorkloadExtension.readTraits(currentWorkload).getWorkflowType()) {
+
+        // Go through all documents in a random order
+        case (RANDOMIZED_WORKFLOW):
```
reckart, line 155: See comment on other menu bar extension regarding document selection.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +115,84 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
+
+    private void loadNewDocument(List<SourceDocument> aSourceDocuments, Project aProject,
+                                 WorkloadManager aCurrentWorkload, AnnotationPageBase aPage,
+                                 Optional<AjaxRequestTarget> aTarget)
+    {
+        // Go through all documents of the list
+        for (SourceDocument doc : aSourceDocuments) {
+            // Check if there are less annotators working on the selected document than
+            // the default number of annotation set by the project manager
+            if ((workloadManagementService.getAmountOfUsersWorkingOnADocument(doc, aProject)
```
reckart, line 176: Looks like duplicated code. Maybe move to a utility class or to a base class for workflow menubar extensions?

reckart: Actually, the logic that checks the number of people working on a document against the traits should be moved into the `nextDocument` method that I described earlier. The code here should basically be something similar to:

```
aPage.getModelObject().setDocument(workflowExtension.nextDocument(workflowManager, candidates),
                         documentService.listSourceDocuments(aProject));
                 aPage.actionLoadDocument(aTarget.orElse(null));
```

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/annotation/DynamicWorkflowDocumentNavigationActionBarExtension.java
```diff
@@ -91,53 +115,84 @@ public Panel createActionBarItem(String aId, AnnotationPageBase aPage)
...
+                            .getDefaultNumberOfAnnotations())) {
+                // This was the case, so load the document and return
+                aPage.getModelObject().setDocument(doc,
+                        documentService.listSourceDocuments(aProject));
+                aPage.actionLoadDocument(aTarget.orElse(null));
+                return;
+            }
         }
+        // No documents left, return to homepage and show corresponding message
+        redirectUSerToHomePage(aPage);
```
reckart, line 187: Typo `S`

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -37,162 +43,190 @@
...
+    private Form<Collection<SourceDocument>> userAssignDocumentForm;
+    private Form<AnnotationDocument> userResetDocumentForm;
+
+    // Current Project
+    private final IModel<Project> currentProject = new Model<>();
 
-    private DataTable<SourceDocument, String> table;
-    private ModalWindow infoDialog;
+    // Date provider
+    AnnotationQueueOverviewDataProvider dataProvider;
```
reckart, line 125: "private"?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -37,162 +43,190 @@
...
-        //Error, user is returned to home page, nothing else to do
-        //getSession required to show the message at the homepage
+        // Error, user is returned to home page, nothing else to do
+        // getSession required to show the message at the homepage
         setResponsePage(getApplication().getHomePage());
         getSession().error("No project selected.");
     }
 
+    /**
+     * Constructor, create the whole page
```
reckart, line 169: JavaDoc comments that state the obvious can be omitted.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.java
```diff
@@ -342,31 +320,282 @@ protected void onUpdate(AjaxRequestTarget ajaxRequestTarget) {
...
+        userSelection.setChoices(this::getUsersForCurrentProject);
+        userSelection.setOutputMarkupId(true);
+
+        // Add AjaxUpdating Behavior to the dropdown
+        userSelection.add(new AjaxFormComponentUpdatingBehavior("change")
+        {
+            @Override
+            protected void onUpdate(AjaxRequestTarget aTarget)
+            {
+                aTarget.add(userSelection);
```
reckart, line 426: You can add multiple targets on a single `add` call.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/DateSelection.java
```diff
@@ -0,0 +1,32 @@
...
+
+public enum DateSelection
+{
+    from("FROM"),
+    until("UNTIL"),
+    between("BETWEEN");
+
+    private final String id;
+
+    DateSelection(String aId)
```
reckart, line 28: What is the "id" used for? It it is a private variable without any getter. Do we need it?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/support/TableHeader.java
```diff
@@ -0,0 +1,35 @@
...
+    Document("DOCUMENT"),
+    Finished("FINISHED"),
+    Processed("PROCESSED"),
+    Annotators("ANNOTATORS"),
+    Updated("UPDATED"),
+    Actions("ACTIONS");
+
+    private final String id;
+
+    TableHeader(String aId)
```
reckart, line 31: id is a private variable without a getter. Do we need it?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadSettingsPanel.java
```diff
@@ -28,67 +30,97 @@
...
+        workloadStrategy
+                .setChoiceRenderer(new LambdaChoiceRenderer<>(WorkloadManagerType::getUiName));
+        workloadStrategy.setModel(LoadableDetachableModel.of(this::getWorkloadManager));
         workloadStrategy.setRequired(true);
         workloadStrategy.setNullValid(false);
         workloadStrategy.setChoices(workloadManagerExtensionPoint.getTypes());
-        workloadStrategy.setModel(LoadableDetachableModel.of(this::getWorkloadManager));
 
-        //add them to the form
+        // add them to the form
```
reckart, line 81: Comments stating the obvious can also be omitted ;)

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadSettingsPanel.java
```diff
@@ -28,67 +30,97 @@
...
+        // Finally, add the confirm button at the end
         Button confirm = new LambdaAjaxButton<>("save", this::actionConfirm);
         form.add(confirm);
 
         add(form);
     }
-    
+
+    /**
+     * This method returns the current WorkloadManagerType.
```
reckart, line 92: `@returns current {@link WorkloadManager}`

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadSettingsPanel.java
```diff
@@ -28,67 +30,97 @@
...
         aTarget.addChildren(getPage(), IFeedback.class);
 
-        workloadManagementService.setWorkloadManagerConfiguration(
-                workloadStrategy.getModelObject().getWorkloadManagerExtensionId(), project);
-
-        success("Workload settings changed");
+        // Either traits are already available for the project
+        WorkloadManager manager = workloadManagementService
+                .loadOrCreateWorkloadManagerConfiguration(project);
+        if (manager != null) {
```
reckart, line 112: You have a `loadOrCreate` method above. Can it actually return `null`?

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/WorkloadSettingsPanel.java
```diff
@@ -28,67 +30,97 @@
...
+            manager.setType(workloadStrategy.getModelObject().getWorkloadManagerExtensionId());
+            workloadManagementService.saveConfiguration(manager);
+        }
+        // or they are not available (first time after project creation or "older" projects)
+        else {
+            DynamicWorkloadTrait trait = new DynamicWorkloadTrait();
```
reckart, line 118: New traits objects should be obtained from the `WorkloadExtension` for the selected type of workload. You should not have to call their constructors directly. Typically, this happens implicitly when you call `workloadManagerExtension.readTraits(workloadManager)` on workload manager that does not have any traits yet - the `readTraits()` would then create the traits object appropriate for the respective `workloadManagerExtension` implementation (e.g. the DynamicWorkloadExtension would create a DynamicWorkloadTraits`(typo "Trait" -> "Traits").

inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/workflow/types/ExternalWorkflowExtension.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.workload.workflow.types;
+
+import de.tudarmstadt.ukp.inception.workload.workflow.WorkflowExtension;
+
+/**
+ * Curriculum workflow extension type
```
reckart, line 23: Curriculum -> External

reckart (changes requested):
Lots of improvements 👍 Still, I have a few more comments.

christophscheunemann:
@reckart  implemented your requested changes
@UWinch issue with "no more available documents" should be fixed, everything works fine here. Please let me know when you still encounter the issue. Furthermore:
1. Update error in the annotation page fixed
2. Added link for workflow strategies documentation (PR #1836 contains documentation)
3. Table now updated accordingly to changes made
4. Still missing: Window closing issue when assigning multiple documents to a user, I'm working on it, its more difficult to fix than I expected

UWinch:
Thanks, the assignment functionality works great. I tested different number of annotators per document in random and default workflow mode, as well as switching back to the static workflow.

I still think you should fix this: 
>     4\. Still missing: Window closing issue when assigning multiple documents to a user, I'm working on it, its more difficult to fix than I expected

Two other issues I had:

- On the workload page, there is the column *processed*. I'm not sure what it indicates, it seems to change when a user starts on the document, as well as when it is assigned to someone. Should it really change when a user is assigned or only when the user started working on it?
- When I accidentally assign a user to a document, I cannot remove them anymore. I would like a function to un-assign them. If you agree, it might be best to open a new issue for this.

reckart:
@UWinch Would you think that it would be reasonable to merge the PR and open separate bug for the window-closing issue - or better wait for it to be fixed. How annoying/critical is it?

christophscheunemann:
@reckart The "processed column" and also the "accidently assigning" issue @UWinch mentioned I have just fixed (not committed though), however I'm still working on the issue you just have mentioned

UWinch:
> @UWinch Would you think that it would be reasonable to merge the PR and open separate bug for the window-closing issue - or better wait for it to be fixed. How annoying/critical is it?

I would like it to be fixed in this PR. If I want to assign documents to users, I might not realize that it did not succeed because of this bug.

reckart:
@UWinch are we good now with the dropdown?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -110,33 +138,144 @@ <h1>
...
+                              && !$('li').is(e.target)
+                              && dropDownMenuButton.has(e.target).length === 0
+                              && $('form.ddform').has(e.target).length === 0
+                              && $('.open').has(e.target).length === 0) {
+                              $('form.ddform').hide();
+                          }
+                      });
+                  </script>
+                <form wicket:id="userForm" class="shadow-lg pt-0 pb-0 ddform" role="menu"
+                  style="width: 1000px; height: 435px;">
```
reckart, line 168: Hard-coding the with to 1000 might not be a good idea, e.g. for people that have small screens. Can we use a relative width?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -110,33 +138,144 @@ <h1>
...
-                    <wicket:message key="settings" />
+                  <script>
+                      let dropDownMenuButton = $('button.dd-toggle');
+
+                      dropDownMenuButton.on('click', function () {
+                          $('form.ddform').show();
+                      });
+
+                      $('body').on('click', function (e) {
+                          if (!dropDownMenuButton.is(e.target)
```
reckart, line 157: Is this really necessary? Please at least comment what you are doing here and why this is necessary.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -17,84 +17,112 @@
...
+      <div class="card-header">
+        <wicket:message key="documentStatus"/>
+          <div class="actions">
+            <span class="dropdown" aria-haspopup="true" aria-expanded="false">
+              <button class="btn btn-action btn-secondary dropdown-toggle flex-content"
+                type="button" data-toggle="dropdown">
+                <i class="fas fa-filter"></i>&nbsp;
+                <wicket:message key="Filter"/>
+              </button>
+              <span class="dropdown-menu shadow-lg pt-0 pb-0" role="menu" style="min-width: 800px;">
```
reckart, line 85: I believe 800px as min-width is too generous. How about using 400px as min-width and allowing to grow relative to the screen size?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -17,84 +17,112 @@
...
+        }
+
+        .form-control.k-widget {
+            display: inline-block;
+            height: 100px;
+            overflow-y: scroll;
+        }
+
+        .k-floatwrap {
+            height: 1000px;
```
reckart, line 41: IMHO hard-coded 1000px width is too much.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -17,84 +17,112 @@
...
-}
+        .table-striped > tbody > tr:nth-child(2n+1) > td, .table-striped > tbody > tr:nth-child(2n+1) > th {
+            background-color: lightgray;
+        }
 
-.table-hover tbody tr:hover td, .table-hover tbody tr:hover th {
- background-color: burlywood;
-}
-</style>
+        .table-hover tbody tr:hover td, .table-hover tbody tr:hover th {
```
reckart, line 30: Let's stick to standard colors defined by the overall theme. Why override the color here?

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -17,84 +17,112 @@
...
-.page-content {
- overflow: auto;
-}
+    <style type="text/css">
+        .page-content {
+            overflow: auto;
+        }
 
-.table-striped>tbody>tr:nth-child(2n+1)>td, .table-striped>tbody>tr:nth-child(2n+1)>th
- {
```
reckart, line 28: why do we need to override the Bootstrap striped table color? https://getbootstrap.com/docs/4.5/content/tables/#striped-rows

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
```diff
@@ -17,84 +17,112 @@
...
+            display: inline-block;
+            height: 100px;
+            overflow-y: scroll;
+        }
+
+        .k-floatwrap {
+            height: 1000px;
+        }
+
+        .ddform {
```
reckart, line 44: Is it really necessary to really extensively re-define this class locally here? If so, why?

reckart (changes requested):

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
christophscheunemann: At least on my screen:
I can not see the difference in the default color theme. The difference is that minor, that it all simply looks the same (nearly white). 
Same for the hovering. 
I found the colors I picked quite good to work with.
However, ofc I can simply remove the override to stick to the default theme.

reckart: The choice of color is one of taste. But the issue of overwriting colors or other styles locally is one of maintainability and consistency. The more we override style locally, the more effort we have when upgrading to e.g. a newer version of Bootstrap (I felt that because I did that earlier this year...).

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
christophscheunemann: I tried a lot. If it gets even with some adjustments below 800px, the complete layout would not look good (missing words in the dropdown etc.)
Is 800px fine?
And as it is only a dropdown, there is no need to grow if there is more space, because especially when we have a lot of space, it might be good to see more from the background (the table)

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
christophscheunemann: Tried here also a lot and adjusted some variables. But same as with the comment above, I cannot go below 600px here. Growth also not required

UWinch:
@christophscheunemann Great work with the dropdown, thanks :) Unfortunately, I also found the same problem existed when filling in the dates on the filter-dropdown. 

Could you also test the time-filtering functionality again, please? I tested with an older imported project and it seems like I could not filter by time-periods and the document-updated column in the table seemed to not correctly update to the new date. I would be ok if you wanted to merge this PR first before looking at that though.

A minor thing: it would be more convenient if the list of documents that I can un-assign only included the actually assigned documents and they were ordered alphabetically. This way it would be easier to find the document to un-assign.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
reckart: We can go with 800px for the moment and if necessary see if/which other options there may be in a future UI update. Thanks!.

inception-workload-dynamic/src/main/java/de/tudarmstadt/ukp/inception/workload/dynamic/management/DynamicWorkloadManagementPage.html
reckart: We can go with 600px for the moment and if necessary see if/which other options there may be in a future UI update. Thanks!.

christophscheunemann:
@UWinch  
- Fixed issue for datepickers
- Dropdown with documents are now all sorted
- Un-assign should now only be possible for documents that are assigned before (should no more be available in the dropdown to select)
- Time filtering and value in the table is fixed. The issue was, that the previous date that I used (document.getUpdated()) does only change its value when a document is assigned for the first time, instead of when "working on the document and making changes" what we actually were intended to be seen. The value in the table should now always be the latest update made in the document

reckart:
When I access the workload page, I see this message in the console:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by net.sf.cglib.core.ReflectUtils$1 (file:/.../.m2/repository-ukp/cglib/cglib/3.3.0/cglib-3.3.0.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of net.sf.cglib.core.ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

I believe this points to a Spring bean being injected into a Wicket page as a class instead of as an interface. I.e. something like `private @SpringBean MyBeanImpl bean;` instead of `private @SpringBean MyBean bean`.

reckart:
If I see it correctly, the **Annotators** dropdown offers two functionalities: 1) assignment of documents to annotators, 2) changing the status of a document for a given annotator. There are two different confirmation buttons. This makes me believe that these two functionalities should be in separate dropdowns. Or the process of assigning / setting the document status should be conflated.

What does it mean to "unassign" a document from a user? I thought it means that the user will not be able to annotate this document, but it looks like what it does is to set the state of the document for that user back to "New" - which should not happen unless the document is actually reset and all annotations are discarded - but that should not happen without a proper confirmation dialog because it is a destructive action.

"Assigning" a document to a user seems to set the state of that document for the user to "in progress". But the document should only be "in progress" if the user has actually worked on it, no? On the other hand, in the context of the dynamic assignment, "in progress" is probably synonymous with "assigned" because at the moment that an annotator starts working on it, that annotator "grabs" it and may thereby make it unavailable for other annotators to "grab" it as well. In the matrix workload assignment, however, I would say a document  "assigned" to an annotator unless it is "locked" for that annotator.

Maybe a single multi-select for choosing the documents and then a single dropdown for assigning a status would be sufficient?

A multi-select for users might also be nice instead of the dropdown.

There does not seem to set the state "locked" on a document for a user.

In the "Filter" dropdown, I would find it convenient if the dropdown closed when I press either "search" or "reset" because otherwise, I don't see the results properly until I manually close the dropdown.

The filter does not allow to filter by document or annotation state (e.g. show only annotations that user John as finished or show me all documents that at least the number of required annotators per document have finished).

I'd say these are all not a blocks for merging, but we should do another round of brain storming about this. 

There seems to be a minor bug: 
* go to the workload page with a manager user
* go to the annotation page in a different browser with an annotator user
* mark a document as finished
* reload the workload page as the manager user
* the finished and in progress counts have not change for the document the annotators has marked as finished
* only when going back to the dashboard and then returning to the workload page, the numbers are refreshed

Idea: it would be cool if the progress on a document would be indicated as a bar instead of numbers. E.g. if you have configured that each document should be annotated by 6 annotators, then the bar could be sub-divided into 6 units. All units are gray by default. For every "in-progress" user, one box is turned to yellow. For every "finished" user, one box is turned to green. If more than the 6 requested users have finished the document there could be a "+1", "+4", etc. behind the document indicating how many extra users have finished it. This could e.g. happen if you reduce the number of required annotators during the project runtime.

I believe the error that goes to the console should still be investigated and fixed and then I'd say we probably should merge this branch - if you agree, @UWinch.

For the other things mentioned in this comment, we should open separate issues and discuss them separately.

reckart:
Sorting by the "in progress" column does not seem to work.

reckart:
"Only unused documents" -> "Only unassigned documents"?

reckart:
"from" - "after"; "until" -> "before"? - maybe in the order "before", "after", "between"?

reckart:
Since the "user names" field in the filter dropdown is not a multi-select for users, might be nice if it would allow wildcards ("anno*") or even a regular expression ("jo.*|ja.*", "anno1[0-9]")

reckart:
Since the states are a small set, we could consider having a quick-filter option for them, similar to the quick-filter option for project roles we have on the project overview page.

reckart:
The first column in the table (name) should be left-aligned, not centered.
The numeric columns should probably be right-aligned.
The annotators column should probably also be left-aligned.

reckart:
The "updated" column could show like "today", "1 day ago", etc. and only start showing dates after a week has passed or so.

christophscheunemann:
@UWinch annotation workflow works fine here (clicking finish automatically selects a new document). Haven't found an issue

reckart (approved):

Issue #1803, opened 2020-09-10
LAPPSGrid NER tests fail


UWinch:
**Describe the bug**
The unit tests fail due to: "The client did not produce a request within the time that the server was prepared to wait". We had this bug before when the service was not reachable.

<details>

```

de.tudarmstadt.ukp.inception.recommendation.api.recommender.RecommendationException: Cannot predict
	at de.tudarmstadt.ukp.inception.recommendation.imls.lapps.LappsGridRecommender.predict(LappsGridRecommender.java:110)
	at de.tudarmstadt.ukp.inception.recommendation.imls.lapps.LappsGridRecommenderConformityTest.predict(LappsGridRecommenderConformityTest.java:133)
	at de.tudarmstadt.ukp.inception.recommendation.imls.lapps.LappsGridRecommenderConformityTest.testNerConformity(LappsGridRecommenderConformityTest.java:94)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at junitparams.internal.InvokeParameterisedMethod.evaluate(InvokeParameterisedMethod.java:234)
	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)
	at junitparams.internal.ParameterisedTestMethodRunner.runMethodInvoker(ParameterisedTestMethodRunner.java:47)
	at junitparams.internal.ParameterisedTestMethodRunner.runTestMethod(ParameterisedTestMethodRunner.java:40)
	at junitparams.internal.ParameterisedTestClassRunner.runParameterisedTest(ParameterisedTestClassRunner.java:146)
	at junitparams.JUnitParamsRunner.runChild(JUnitParamsRunner.java:446)
[...]

PR #1804, opened 2020-09-10
#1803 LAPPSGrid NER tests fail

UWinch:
Getting response *unauthorized* from service, but service is open to all users to my knowledge (no username/password). So, ignore tests when we get this status.

**What's in the PR**
* Updated lappsgrid tests to ignore tests when http response is 401

**How to test**
* Run unit tests


jcklie (approved):
Thanks.

Issue #1812, opened 2020-09-28
Tutorial fails on setting tabs


UWinch:
**Describe the bug**
The order of tabs in the settings changed again, fix the tutorial again. I gave the tab a name last time, check if this still works.

PR #1815, opened 2020-09-28
#1812 Tutorial fails on setting tabs

UWinch:
Tutorial failed because of change from previous tab order of the setting tabs. Naming the recommender tab that we want to select here does not work, because they are represented as links at the time. We will have to keep identifying the correct tab by order.

**What's in the PR**
* Change back to previous tab order.

**How to test manually**
* go through the tutorial and check that everything works


inception-workload/src/main/java/de/tudarmstadt/ukp/inception/workload/settings/ProjectWorkloadSettingsPanelFactory.java
```diff
@@ -33,7 +33,7 @@
  * {@link WorkloadManagementAutoConfiguration#projectWorkloadSettingsPanelFactory}.
  * </p>
  */
-@Order(300)
+@Order(600)
```
UWinch, line 36: @christophscheunemann is there a reason why the order of the tabs in the settings panel is `300` for the workload panel? then I will not change it back here ;)

reckart:
The availability of tabs may change depending on which features the user has enabled/disabled in the settings.properties. Maybe we could use the tab labels for the moment (although these might at some point change as well when we do more I18N)?

reckart (approved):

reckart:

Issue #1860, opened 2020-11-24
Layers without features are not indexed for search


reckart:
**Describe the bug**
If a layer does not have an indexable feature, then the entire layer is not indexed.

**To Reproduce**
1. Create a new layer `Entityrelation` on the span layer `Named entity`
2. Create two `Named entity`annotations and connect them using an `Entityrelation`
3. Try searching in the search sidebar for `<Entityrelation/>`
4. No result.

**Expected behavior**
The entity relation annotation should be found.

**Please complete the following information:**
 - Version and build ID: 0.17.1

PR #1863, opened 2020-11-24
#1860 - Layers without features are not indexed for search

reckart:
**What's in the PR**
- Also index layers that do not have any features
- Improve trace log
- Fix issue in the search sidebar that it was displaying "1-10 of 1" results - now does "1-1 of 1".

**How to test manually**
* See issue description

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


UWinch (approved):
Searching layers without features works, great! Tested with relation and span layer.

PR #1871, opened 2020-11-29
Feature/1870 improve roundtrip times on annotation page

reckart:
**What's in the PR**
- Using the username instead of the User object reduces database access for checking if a user is viewing another users annotations
- Store the available layers in the annotator state. If the layers change, the change in the type system will require the user to reload the annotation editor anyway, so we do not need to listen to any events for refreshing the list. Getting the layers from the state is way faster than getting them from the DB every time.
- Maintain cache of tags within a tagset which is used primarily for the feature editors and which is flushed when a tag is created or deleted.
- Remove previous hack of maintaining a transient field in the Tag entity to mark it as reordered. Instead have a proper ReorderableTag class for this.

**How to test manually**
* No specific test procedure

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:

Issue #1765, opened 2020-07-20
Curation Sidebar: Should be able to change user even when finished


UWinch:
**Is your feature request related to a problem? Please describe.**
Using the curation sidebar, you can change between current user and the curation user. However, when you finish a document as one of the users you should still be able to switch to the other user. Examples are cases where annotators also do curation.

**Describe the solution you'd like**
Enable the settings button even when document is finished, maybe move it to the top of the sidebar.

lfoppiano:
I was going to open the same issue. Once there aren't document for curation it's not possible to switch back. 

Probably, leaving the settings button enabled in the panel should solve the problem. 

Also, it might happen that the user doesn't notice that he/she is editing the curation document instead of his/her annotation document. Not sure if (and how) this can be avoided, though.

PR #1872, opened 2020-11-29
#1765 Curation Sidebar: Should be able to change user even when finished

UWinch:
**What's in the PR**
- [x] always allow to change settings
- [x] update message that doc is finished to include username
- [x] fix user in annotator state not being up to date with curator
- [x] should also be able to show my annotation user's annotations during curation with CURATION_USER
- [x] fix that CURATION_USER document can be finished but this does not update to document state (as seen on monitoring page)

**How to test manually**
Checkout webanno branch for https://github.com/webanno/webanno/issues/1822
* Enable the curation sidebar with `curation.sidebar.enabled=true` in your settings.properties
* Be an annotator and curator
* Annotate a document as yourself and finish it
* You should be able to switch to the CURATION_USER via the curation sidebar to curate it

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart (approved):
Did testing in the UI. Looks good. Good attention to details like using a placeholder to put the user name into the message that the document has been locked. Thanks! 👍

Issue #1951, opened 2021-01-21
Custom error page


reckart:
**Is your feature request related to a problem? Please describe.**
When there is an error, there is a quite unhelpful and rather ugly error page - either generated by Wicket if there is a problem in the application code or from Spring if there is a problem with serving stuff or starting up in the first place.

**Describe the solution you'd like**
A nice and helpful error page.

PR #1952, opened 2021-01-21
#1951 - Custom error page

reckart:
**What's in the PR**
- Replace the Spring Whitelabel error page with a nice page that uses the application design and provides a bit of information that may be helpful when reporting an error.

**How to test manually**
* For a 404 error, try e.g. accessing `/assets/blah`
* Try accessing "/whoops/test" after logging in for an error with a stack trace (make sure to be running in Wicket development mode, otherwise the test page is not mounted)

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
WDTY folks? Should we merge it?

PR #1961, opened 2021-01-24
Feature/46 letor entity linking

reckart:
**What's in the PR**
* Refactoring of the entity linking code we used to hook up an external entity linking ranker
* *NOTE:* there is still quite a bit that would need to be done before this code be merged...

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
I assume nothing more will happen here. We can always re-open.

PR #1962, opened 2021-01-26
Feature/1876 sharing project links

UWinch:
When using e.g. SSO, I need to wait for project managers to add me. If a project calls for annotations, I want to immediately start annotating.

**What's in the PR**

* add check for invite-link to dashboard page
* add UI-elements for creating, removing, showing, copying link in settings tab

**How to test manually**
* As a manager of a project, go to the project settings, share-project tab and click the **New Link** button to create an invite link
* Copy the invite link
* As a user that is not yet part of the project use this link to access inception
* Log in and be automatically added to the project and directed to its dashboard page

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


inception-sharing/src/main/java/de/tudarmstadt/ukp/inception/sharing/ProjectInvite.java
```diff
@@ -0,0 +1,80 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "project_invite")
+public class ProjectInvite
+{
+    @Id
+    @Column(nullable = false)
```
reckart, line 30: I am surprised it works without adding `@GeneratedValue(strategy = GenerationType.IDENTITY)` as e.g. in `SourceDocument`.

inception-sharing/src/main/java/de/tudarmstadt/ukp/inception/sharing/panel/InviteLinkPanel.java
```diff
@@ -0,0 +1,155 @@
...
+        linkMainContainer = new WebMarkupContainer("linkMainContainer");
+        linkMainContainer.setOutputMarkupId(true);
+        add(linkMainContainer);
+
+        inviteLinkContainer = createInviteLinkContainer();
+        inviteLinkContainer.add(visibleWhen(() -> linkField.getModelObject() != null));
+        inviteLinkContainer.setOutputMarkupId(true);
+        linkMainContainer.add(inviteLinkContainer);
+
+        AjaxLink<Void> shareBtn = new AjaxLink<>("shareProject")
```
reckart, line 69: LambdaAjaxLink - and pull the aTarget.add out into the `shareProject(target)` method which we would normally name `actionShareProject(...)` to highlight it is a user-action handler.

inception-sharing/src/main/java/de/tudarmstadt/ukp/inception/sharing/panel/InviteLinkPanel.java
```diff
@@ -0,0 +1,155 @@
...
+
+        linkField = new TextField<>("linkText", LoadableDetachableModel.of(this::getInviteLink));
+        linkContainer.add(linkField);
+
+        Button copyBtn = new Button("copy");
+        ClipboardJsBehavior clipboardBehavior = new ClipboardJsBehavior();
+        clipboardBehavior.setTarget(linkField);
+        copyBtn.add(clipboardBehavior);
+        linkContainer.add(copyBtn);
+        AjaxLink<Void> regenBtn = new AjaxLink<>("regen")
```
reckart, line 98: LambdaAjaxLink

inception-sharing/src/main/resources/de/tudarmstadt/ukp/inception/sharing/db-changelog.xml
```diff
@@ -0,0 +1,42 @@
...
+      </not>
+    </preConditions>
+    <createTable tableName="project_invite">
+      <column name="projectId" type="BIGINT">
+        <constraints primaryKey="true"/>
+      </column>
+      <column name="inviteId" type="VARCHAR(255)">
+        <constraints nullable="false"/>
+      </column>
+      <column name="expirationDate" type="BIGINT">
```
reckart, line 36: Shouldn't this be a DATE or TIMESTAMP?

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/dashlet/CurrentProjectDashlet.html
```diff
@@ -30,8 +30,8 @@
   </wicket:head>
   <wicket:panel>
     <h1><div wicket:id="name" class="block-ajax-editable-label"/></h1>
-
     <p><wicket:container wicket:id="description"/></p>
+    <div wicket:id="inviteLinkPanel"></div>
```
reckart, line 34: As discussed offline, right now we should make this panel a project settings panel because there we have an extension point we can easily plug the new functionality into. On the dashboard, we'd first have to create extensibility.

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/project/ProjectDashboardPage.java
```diff
@@ -51,7 +56,7 @@
 /**
  * Project dashboard page
  */
-@MountPath(value = "/project/${" + PAGE_PARAM_PROJECT_ID + "}")
+@MountPath(value = "/project/${" + PAGE_PARAM_PROJECT_ID + "}/#{" + PAGE_PARAM_INVITE_ID + "}")
```
reckart, line 59: Hm, why use a HTML anchor tag (`#`) here instead of a query parameter (`?`)?

reckart (changes requested):

inception-sharing/src/main/java/de/tudarmstadt/ukp/inception/sharing/ProjectInvite.java
UWinch: It's not a generated value, it's the (unique) project ID.

reckart: The primary key of your table is simultaneously a foreign key? Uh - I guess that might work, but it feels very very odd. I think the project should be a separate column (i.e a field of type Project) with a corresponding foreign key to the project table. That's how we usually do it.

inception-ui-core/src/main/java/de/tudarmstadt/ukp/inception/ui/core/dashboard/project/ProjectDashboardPage.java
UWinch: The `#` is wicket's way to mark an optional part of the url, an example invite url would be something like `https://blabla.com/inception/projects/2/NMshwsuwhdw_`

UWinch:
@reckart You can have a look at it again if you want. I still want to add some tests and the build has some problems at the moment.

jcklie:

reckart:
@UWinch which problem does the build have? Jenkins seems to be green.

jcklie:
I fixed it, it was a jenkins docker PR config error

UWinch:
@jcklie Thanks for fixing it!

inception-sharing/src/main/java/de/tudarmstadt/ukp/inception/sharing/InviteServiceImpl.java
```diff
@@ -0,0 +1,134 @@
...
+
+import org.apache.commons.lang3.Validate;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.stereotype.Component;
+import org.springframework.transaction.annotation.Transactional;
+
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+
+@Component
+@ConditionalOnProperty(prefix = "sharing", name = "enabled", havingValue = "true")
```
jcklie, line 37: I do not like the property name here, we at least need a higher level when we want to group related properties, For me it is not share but invite links, so I would have named it something like that.

jcklie (approved):
It works! Some small remarks:
- I would like to see how long the link is valid and be able to extend its duration
- I would like a popup when you join a project that you joined this project

inception-sharing/pom.xml
```diff
@@ -86,5 +111,19 @@
   		<artifactId>wicketstuff-clipboardjs</artifactId>
   		<version>${wicketstuff-clipboardjs.version}</version>
   	</dependency>
+  	<dependency>
+  		<groupId>junit</groupId>
+  		<artifactId>junit</artifactId>
+  	</dependency>
```
reckart, line 117: scope test for junit?

reckart:
> I would like to see how long the link is valid and be able to extend its duration
> I would like a popup when you join a project that you joined this project

I'd suggest to open new issues for these ideas so they do not block the release.

reckart:
@UWinch to improve modularity, I have moved the code which handles the invite out of the project dashboard and into a separate page where the user may explicitly accept the invite. As part of the invite, a feedback message is generated which then appears as the user is redirected to the dashboard page after accepting the invite. This should address @jcklie request for the popup notifying the user that the invite was accepted.

I also fixed a bug that would remove any permissions that are not "annotator" from a user accepting an invite. A manager trying this could have easily locked herself out of the project. The invite now adds the "annotator" permission for the user to the project if the user didn't already have it - any existing permissions are retained.

reckart (approved):

Issue #1966, opened 2021-01-29
No images in guides


UWinch:
**Describe the bug**
The images can no longer be displayed in the guides.

**To Reproduce**
Build the documentation with Maven (it works with the test class in inception-doc) and see that there are no images in the documentation (instead the image name is displayed).

**Additional context**
In development, but not in stable branch.

PR #1967, opened 2021-01-29
#1966 No images in guides

UWinch:
Asciidoctor images directory attribute was not used and therefore no images showed up in the documentation.

**What's in the PR**
* Moved imagesDir into attributes

**How to test manually**
* Build with Maven and look at the generated documentation. There should be pictures.


jcklie:
Does this also fix the manual generation of the documenation via the junit test or was that not affected?

UWinch:
The test was not affected.

reckart (approved):

Issue #1975, opened 2021-02-02
Allow disabling SSL certificate checks for individual SPARQL repositories


reckart:
**Is your feature request related to a problem? Please describe.**
It would be nice if it would be possible to disable SSL certificate checks on a per-SPARQL repo basis.

**Describe the solution you'd like**
A checkbox next to the URL by which the SSL certificate checking can be disabled.

**Describe alternatives you've considered**
* Disable checks globally
* Not allowing to disable checks and instead require self-signed certificates to be imported  into a global trust store

**Additional context**
See also https://github.com/eclipse/rdf4j/issues/2793

PR #1976, opened 2021-02-02
#1975 - Allow disabling ssl certificate checks for individual sparql repositories

reckart:
**What's in the PR**
- Added a HTTP client variation which uses a thread-local variable to check if SSL verification should be performed
- Added a checkbox to the UI and a boolean column to the DB
- Hook up SSL certificate check skipping in most unit tests

NOTE: the solution is rather hacky. Also disabling the SSL checks in the tests is probably a bad idea / unnecessary. It seemed necessary at first to me because some tests did get skipped due to the SSL checks failing but it turned out that it was due a bug in Java 11.0.1 which got fixed in later revisions.

**How to test manually**
* Try connecting to a SPARQL repository with a self-signed or invalid certificate

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
_Jenkins, can you test this please?_

Issue #1971, opened 2021-02-01
Cannot annotate properties only concept feature with wiki-data


UWinch:
**Describe the bug**
I cannot annotate concept features that are linked to wiki-data remote knowledge base and configured to **properties only**. This works for local knowledge bases (as long as they have labels configured). Tested with another SPARQL endpoint remote KB (GND), the same error occurs. 
A TimeoutException occurs during the SPARQL query.

<details>

```
2021-02-01 17:14:59 ERROR [admin] ConceptFeatureEditor - An error occurred while retrieving entity candidates
org.eclipse.rdf4j.query.QueryEvaluationException: SPARQL-QUERY: queryStr=SELECT DISTINCT ?lc ?dc ?subj
WHERE { { { { OPTIONAL { ?pLabel <http://www.wikidata.org/prop/direct/P1647>* <http://www.w3.org/2000/01/rdf-schema#label> . }
{ ?subj ?pLabel ?lc .
FILTER ( ( REGEX( ?lc, ".*academic.*", "i" ) && LANGMATCHES( LANG( ?lc ), "en" ) ) || ( REGEX( ?lc, ".*academic.*", "i" ) && LANGMATCHES( LANG( ?lc ), "" ) ) ) } UNION { ?subj ?pLabel ?lc .
FILTER ( ( REGEX( ?lc, ".*Illinois.*", "i" ) && LANGMATCHES( LANG( ?lc ), "en" ) ) || ( REGEX( ?lc, ".*Illinois.*", "i" ) && LANGMATCHES( LANG( ?lc ), "" ) ) ) } }
FILTER EXISTS { { ?subj <http://www.wikidata.org/prop/direct/P31>/<http://www.wikidata.org/prop/direct/P279>* <http://www.wikidata.org/entity/Q18616576> . } UNION { [ ] <http://www.wikidata.org/prop/direct/P1647> ?subj . } UNION { ?subj <http://www.wikidata.org/prop/direct/P1647> [ ] . } } } }
OPTIONAL { { ?subj <http://www.w3.org/2000/01/rdf-schema#comment> ?dc .
FILTER ( LANGMATCHES( LANG( ?dc ), "en" ) ) } UNION { ?subj <http://www.w3.org/2000/01/rdf-schema#comment> ?dc .
FILTER ( LANGMATCHES( LANG( ?dc ), "" ) ) } } }
LIMIT 2000

java.util.concurrent.TimeoutException
[...]

PR #1992, opened 2021-02-08
#1971 Cannot annotate properties only concept feature with wiki-data

UWinch:
**What's in the PR**
* Moved query for label to secondary pattern for query with regex in *matchAnyOf...* if Fulltextsearch is not available. This is still slow, but we get a result.

**How to test manually**
* Add wikidata as your knowledge base
* Configure a layer with a concept feature, restrict the feature to properties
* Annotate a text span and try setting the feature by searching for something
 

**Automatic testing**
* [x] PR includes unit tests


reckart (approved):
Nice one :)

Issue #1965, opened 2021-01-29
Extend invite settings panel


UWinch:
**Is your feature request related to a problem? Please describe.**
Follow up to https://github.com/inception-project/inception/pull/1962. I would like to see how long the link is valid and be able to extend its duration.

PR #1998, opened 2021-02-09
#1965 Extend invite settings panel

UWinch:
**What's in the PR**
* The expiration date is now shown 
* It can be extended

**How to test manually**
* Enable invite links in settings.properties with `sharing.invites.enabled=true`
* Visit the settings **Share project** tab
* Configure invite link
* Extend its duration
* Compare shown expiration dates

**Automatic testing**
* [x] PR includes unit tests


reckart:
The PR allows extending the invitation for another year. Wouldn't it make more sense to allow the user to actually set the expiry date of the invitation? Maybe the user just wants to extend by a day or even to reduce the time the invite is valid by setting some end-date next week (say before a workshop).

UWinch:
I thought this would be the easiest option for the user, just clicking a link instead of selecting a date even though you might not be sure about it. If you implement the selection for the extension, you should also provide it when you first enable sharing and then you're again in a situation when you have to decide on options from the beginning as a user.

However, if you have a strong opinion on this, I can change it.

reckart:
Maybe we can have both? One-click extension by a fixed amount of time a well as the ability to explicitly edit the end date?

jcklie:
If I set the expiration for a year and then click the time picker, then I would expect the date I set previously to be selected. If I set dates before now, then it clears the link but shows a popup saying that I set the expiration date.

reckart (approved):
Played around with extending and trying to set different dates - looks nice :)

PR #2000, opened 2021-02-11
Feature/1856 support regex based recommendations

HerrKrishna:
**What's in the PR**
* This is a first version just to let you have a look if this is going into the right direction.
* I will add unit tests next

**How to test manually**
* Setup a new Regex Recommender for the Named Entity Layer, and value feature
* Check the "Always active" box.
* Select the document that I have attached. It contains lemma annotations.
* Add a Named Entity annotation
* Enter a regex
* Now you should get recommendations for all the lemmas that fit your regex.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

[Bundestag_08-1.zip](https://github.com/inception-project/inception/files/5966112/Bundestag_08-1.zip)


reckart:
Thanks for the PR. I have had a look at the code and it looks quite interesting. You appear to have the need to show a popup as part of accepting an annotation and that currently seems to work only if INCEpTION is run as a standalone version in a graphical desktop environment. This requirement is something I need to think about - i.e. if/how this can be handled in such a way that it works in a purely browser-based environment.

Issue #2027, opened 2021-02-22
PDFs do not load with reverse proxy + https


reckart:
**Describe the bug**
```
The second error, "Mixed Content: The page at'https://morbo.ukp.informatik.tu-darmstadt.de /annotation.html?4&p=549#!p=549&d=2240' was loaded over HTTPS, but requested an insecure resource'http://morbo.ukp.informatik.tu-darmstadt.de/resources/pdfanno/index.html ?pdf=http://morbo.ukp.informatik.tu-darmstadt.de/annotation.html?4-1.0-centerArea-editor-vis&p=549&pdftxt=http://morbo.ukp.informatik.tu-darmstadt.de /annotation.html?4-1.1-centerArea-editor-vis&p=549&api=http://morbo.ukp.informatik.tu-darmstadt.de/annotation.html?4-1.2-centerArea-editor-vis&p=549'. This request has been blocked; the content must be served over HTTPS."
```

**To Reproduce**
1. Upload a PDF and try viewing it

**Expected behavior**
Should work.

**Please complete the following information:**
 - Version and build ID: 0.18.1

PR #2036, opened 2021-02-24
#2027 - PDFs do not load with reverse proxy + https

reckart:
**What's in the PR**
- Calculate the callback URLs relative to the URL of the iframe in which pdf.js is loaded

**How to test manually**
* Run INCEpTION behind a https reverse proxy which forwards to http
* Check if simply opening a PDF for annotation on Windows works

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
@jcklie Would be cool if you could try if this works on Windows - reason being that I am using the `Path` object to help constructing the relative URL and that works differently on Windows than on macOS/Linux.

jcklie:
Loads for me. Is it possible to make the pdf editor default when importing PDF? I find it strange that I have to select it the first time I want it.

jcklie (approved):

reckart:
> Loads for me. Is it possible to make the pdf editor default when importing PDF? I find it strange that I have to select it the first time I want it.

Not currently possible. We'd need to add a hook for that into editor registry or such.

Issue #2048, opened 2021-03-04
Change status of annotation documents from API


ofersabo:
We work with many annotators and multiple projects on a relatively large scale.
Often, at least a few documents in a project are annotated in a way that contradicts the annotation guidelines.
Currently,  we can't check the validation of these annotations from INCEPTION. 
I don't think this is possible to validate this from the constraint feature. 

Therefore, we must download these annotations and check them on our servers with our own script. 
The issue arises when we find documents that are not valid. 
We need to re-open these documents after those were closed by the annotators when they finished.

Doing this manually is too much labor. 

Given a document ID within a project, change the file status.
for example, close or open the file for annotation.

reckart:
Since you ask, INCEpTION indeed does have a mechanism that can invoke validators when the user tries to mark a document as finished. In principle, it would be possible to implement a custom validator in Java and hook it in there. See e.g. the `SpanCrossSentenceBehavior` class, in particular the `onValidate(...)` method.

PR #2054, opened 2021-03-06
#2048 - Change status of annotation documents from API

reckart:
**What's in the PR**
- Added new endpoint for setting the annotation document state
- Switch from Springfox to Springdoc

**How to test manually**
* Enable remote API
* Access swagger-ui.html
* Try to use the new endpoint to change the state of an annotation document

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


ofersabo:
@reckart Any idea when would this be available in the docker version? How often do you push the changes to a new image?

reckart:
The docker image is updated as part of the normal release cycle. So when 0.19.0 is released, there will be a docker image for it. So far, the interval between feature releases has been about 4-8 weeks.

ofersabo:
I'm not sure this is related here, cause I'm not sure who stands behind the pycaprio package. 
Anyway, do you happen to know whether this change would also be reflected in the pycaprio SDK? or we would need to create a python wrapper to call this API functionality?

reckart:
That is a question for @JavierLuna

JavierLuna:
Hi!
I developed `pycaprio` while I was working in Savana and I no longer work there. As far as I know, they do not have plans to continue maintaining the project, but I'd like to do it myself.

I've just [forked pycaprio](https://github.com/JavierLuna/pycaprio) and will add support for this changes soon. I've also contacted the people at Savana to check wether they give me the ownership of the PyPi project. This way I don't have to publish a `pycaprio2` and you won't need to change the dependency name.

I'll get back to you whenever I hear back from them. Thank you for using the package!

JavierLuna:
Hi, sorry for the late update.

I've started conversations with my former company for an ownership transfer of the `pycaprio` PyPi package. However I have the impression that it might take a while.

I've just upload the new functionality in the [0.2.1 release](https://github.com/JavierLuna/pycaprio/tree/0.2.1). I've also modified the installation instructions to show how to install from a Github repo+tag.

Issue #1289, opened 2019-07-09
Debounce scheduler job submission


jcklie:
**Is your feature request related to a problem? Please describe.**
Right now, if a job is scheduled twice in quick succession to due e.g. bursting events, it can happen that the scheduler starts a job and adds the job to the queue. It should be possible that the job is not added to the queue if the submission times are really close.

**Describe the solution you'd like**
Add a start date to each job and check it when scheduling jobs.

PR #2149, opened 2021-03-30
#1289 - Debounce scheduler job submission

reckart:
**What's in the PR**
- Add DebouncingTask and ability in the schedule to support debouncing
- Also support tasks that are not associated with a specific user (i.e. SYSTEM tasks)

**How to test manually**
* No specific test procedure

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/DebouncingTask.java
+++ b/inception/inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/DebouncingTask.java
@@ -17,6 +17,8 @@
  */
 package de.tudarmstadt.ukp.inception.scheduling;
 
+import java.time.Duration;
+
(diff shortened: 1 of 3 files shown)


inception/inception-scheduling/src/main/java/de/tudarmstadt/ukp/inception/scheduling/DebouncingTask.java
```diff
@@ -0,0 +1,51 @@
...
+
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+public abstract class DebouncingTask
+    extends Task
+{
+    private final long runnableAfter;
+
+    public DebouncingTask(Project aProject, String aTrigger, long aDebouncePeriod)
```
jcklie, line 28: It would be good to document in which units this is, maybe use java.time.Duration

Issue #2176, opened 2021-04-07
Make long relations nicer without arcs


jcklie:
**Is your feature request related to a problem? Please describe.**
Relations spanning more than a row are sometimes not looking good, especially if there are many of these.

**Describe the solution you'd like**
Several things are thinkable:

1. Use numbers at source and target 
2. Use same color something around source and target, e.g. underline

Hide the arcs, on hover, highlight both as always

reckart:
I believe #4826 solves this issue here as well.

PR #2218, opened 2021-04-21
#2176 - Make long relations nicer without arcs

jcklie:
**What's in the PR**
- Option to not render relations

Highlighting of the other span when hovering/rendering of this via equivalenve sets is still missing.

**How to test manually**
* ...

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

**Todo**
- [ ] Make adding equivalence sets fast
- [ ] Add hover for same equivalence sets in BRAT


inception/inception-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/adapter/SpanAdapter.java
```diff
@@ -77,6 +81,13 @@ public SpanAdapter(LayerSupportRegistry aLayerSupportRegistry,
         }
     }
 
+    @Override
+    public void initialize(AnnotationSchemaService aSchemaService)
```
reckart, line 85: This method is called in order to create default features in the annotation schema when a new layer is being created. This is *not* meant to be used to initialize the adapter itself. It is called only by `LayerDetailForm.actionSave`. You cannot reliably obtain an `annotationSchemaService` in this way.

jcklie:
I do not like the name of that strategy but I do not know a better name for it.

reckart:
Looking repeatedly for attached relations via the method offered by the annotation service is quite slow - not something I think is a good idea to do during rendering. It is ok in the annotation sidebar where we call it only once, but in a loop as during rendering it doesn't scale well. Searching the `VDocument` would probably be quite a bit faster.

jcklie:
> Looking repeatedly for attached relations via the method offered by the annotation service is quite slow - not something I think is a good idea to do during rendering. It is ok in the annotation sidebar where we call it only once, but in a loop as during rendering it doesn't scale well. Searching the `VDocument` would probably be quite a bit faster.

I do not know how to get the relations from the vdoc, as the spans are rendered first and then the relations.

Issue #2243, opened 2021-05-01
Directly select tags for small tagsets


reckart:
**Is your feature request related to a problem? Please describe.**
Selecting a label usually requires typing or opening a  drop down box to select a label.

**Describe the solution you'd like**
To avoid having to type or click too much, it would be nice if there was an option to configure a tagset-based feature editor simply as a radio group.

PR #2244, opened 2021-05-01
#2243 - Directly select tags for small tagsets

reckart:
**What's in the PR**
- Add radio-group-based editor
- Add option to select the preferred editor in the feature traits

**How to test manually**
* Use with tagset <= 5 tags
* Manually configure which editor to use in the feature settings
* Also try with tagsets that have or have not descriptions

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/feature/StringFeatureSupport.java
+++ b/inception/inception-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/feature/StringFeatureSupport.java
@@ -191,7 +191,8 @@ public class StringFeatureSupport
         FeatureState featureState = aFeatureStateModel.getObject();
 
         // For really small tagsets where tag creation is not supported, use a radio group
-        if (!featureState.feature.getTagset().isCreateTag() && featureState.tagset.size() <= 3) {
+        if (!featureState.feature.getTagset().isCreateTag()
+                && featureState.tagset.size() < properties.getComboBoxThreshold()) {
             return RADIOGROUP;
         }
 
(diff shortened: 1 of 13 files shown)


jcklie:
I would expect that clicking the radio button updates the brat, but it does not? 

image

I also used Ortography as a layer with 4 tags, but it was an autocomplete by default whereas this issue says <= 5 gets radio.

jcklie:
It would also be nice if clicking on the selected tag would remove/reset the label to none.

reckart:
You need to make sure the tagset does not have "anntotator can add new tags" enabled.

Yes, clicking should indeed update.

reckart:
> It would also be nice if clicking on the selected tag would remove/reset the label to none.

Remember, this is a radio-group, not a multiple-choice component. The selection behavior is determined by the radio group as implemented in the browser. Clicking means selecting, not unselecting. There is the "x" in the feature header to unselect.

jcklie:
> > It would also be nice if clicking on the selected tag would remove/reset the label to none.
> 
> Remember, this is a radio-group, not a multiple-choice component. The selection behavior is determined by the radio group as implemented in the browser. Clicking means selecting, not unselecting. There is the "x" in the feature header to unselect.

I know, but it does not look like radio buttons right now so I wish more functionality

jcklie (approved):
Works for me

jcklie:
You should not be able to select radio buttons if "add new tags" is true for the tagset/feature

jcklie:
Maybe make tagsets default closed

reckart:
For the inclined reader: the "operation" feature we used here to test is by default not shown in the label of orthography annotations. For other layers+features (e.g. Named entity / value) it works.

Issue #2111, opened 2021-03-18
Replace okhttp in external recommender with Java HTTP client


jcklie:
**Describe the refactoring action**
Replace okhttp in external recommender with Java HTTP client

**Expected benefit**
Less dependencies.

PR #2270, opened 2021-05-08
#2111 - Replace okhttp in external recommender with Java HTTP client

reckart:
**What's in the PR**
- Use Java HTTP client in external recommender
- Switch to mockwebserver3-junit5 dependency
- Switch to a different Matomo/Piwik client library which does not use okhttp
- Remove kotlin Jackson module

**How to test manually**
* Check that there is no warning about a Jackson Kotlin module during startup
* Try setting up a WebLicht recommender and check if it is producing any results
 
**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


inception/inception-imls-external/src/main/java/de/tudarmstadt/ukp/inception/recommendation/imls/external/ExternalRecommender.java
```diff
@@ -129,26 +128,26 @@ public void train(RecommenderContext aContext, List<CAS> aCasses) throws Recomme
...
-            else if (!response.isSuccessful()) {
-                int code = response.code();
-                String responseBody = getResponseBody(response);
-                String msg = format("Request was not successful: [%d] - [%s]", code, responseBody);
-                throw new RecommendationException(msg);
-            }
-
-            aContext.put(KEY_TRAINING_COMPLETE, true);
+        HttpRequest request = HttpRequest.newBuilder() //
+                .uri(URI.create(traits.getRemoteUrl()).resolve("train")) //
```
jcklie, line 132: Uri::resolve is dangerous I think if the last char is not a slash:

```
import java.net.URI;
public class Main{

     public static void main(String []args){
        System.out.println(URI.create("http://example.com/foo/bar/baz").resolve("train"));
     }
}
```

`> http://example.com/foo/bar/train`

reckart: Looks like `URI.create()` normalizes paths and there is also a handy `appendIfMissing` method in `StringUtils` - so I'll use that to ensure the URL ends in a slash before resolving.

Issue #2295, opened 2021-05-21
Links to user guide do not work


UWinch:
**Describe the bug**
Clicking on question marks in the application should open the user guide on a relevant topic. This does no longer work for at least some of them in the settings tab or monitoring, agreement page.

**To Reproduce**
1. Click on a question mark (e.g. agreement page)

**Please complete the following information:**
 - Version and build ID: 0.19.3, 0.20.x
 - OS: linux
 - Browser: chrome

PR #2296, opened 2021-05-21
#2295 Links to user guide do not work

UWinch:
**What's in the PR**

Adapt help links on pages to new url scheme

**How to test manually**
* Click on the blue-white question marks, a related section in the user guide should open.


inception/inception-ui-project/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/project/layers/ProjectLayersPanel.properties
```diff
@@ -15,10 +15,10 @@
 # limitations under the License.
 
 help.label=Help
-layers.help.link=doc/user-guide.html#sect_projects_layers
-properties.help.link=doc/user-guide.html#sect_projects_layers_properties
-behaviours.help.link=doc/user-guide.html#sect_projects_layers_behaviours
-features.help.link=doc/user-guide.html#sect_projects_layers_features
+layers.help.link=../../../doc/user-guide.html#sect_projects_layers
```
reckart, line 18: This works when accessing the layer settings through the project dashboard with URLs of this form

```
http://localhost:8080/inception-app-webapp/p/394/settings/layers
```

But it doesn't work when accessing the settings via the Administrator -> Projects path which has these URLs:

```
http://localhost:8080/inception-app-webapp/admin/projects
```

reckart (commented):
I think we may have to go for the more sophisticated approach of introducing a `HelpLink` component because in particular for the project settings panels, we have two paths of accessing them and they have a different "depth" in their URLs.

Issue #2281, opened 2021-05-18
Experimental annotation API integration into BRAT


christophscheunemann:
**Integration of the new API into BRAT**
Brat Editor shall be fully functional using the new annotation API

reckart:
Brat editor is now mostly functional using the new APIs. It still has a few specials, but they are actually now built on top of the new APIs through extensions that it allows.

Issue #2211, opened 2021-04-20
Websocket integration for the experimental annotation API


christophscheunemann:
**Websocket for the experimental annotation API**
* Websocket is a required feature for the experimental annotation API and needs to be implemented

reckart:
We already have general websocket support enabled at the level of Tomcat and Spring. What do you want to do specifically?

reckart:
At least I believe we have it enabled. If not, it should be enabled on this PR: https://github.com/inception-project/inception/pull/2051

Issue #2196, opened 2021-04-14
Features for the experimental annotation API


christophscheunemann:
**Extension for the advanced annotation API is required**
* Basic implementation for all the features the API requires
* Testing via Custom Editor (Custom Editor will call the API)

PR #2307, opened 2021-05-26
Feature/2196 features for the experimental annotation api

christophscheunemann:
**Advanced Features and Brat Integration of the API**
- Issues:  #2196, #2211, #2281


christophscheunemann:
@jcklie Too bad, because Gson worked quite well :D But sure I'll use Jackson then

Issue #2358, opened 2021-06-09
Unrecognized field "ignoreCase" in external recommender traits


jcklie:
**Describe the bug**

`Unrecognized field "ignoreCase" (class de.tudarmstadt.ukp.inception.recommendation.imls.weblicht.traits.WeblichtRecommenderTraits), not marked as ignorable (5 known properties: "chainInputFormat", "chainInputLanguage", "lastKeyUpdate", "apiKey", "url"]) at [Source: (String)"{"ignoreCase":false}"; line: 1, column: 20] (through reference chain: de.tudarmstadt.ukp.inception.recommendation.imls.weblicht.traits.WeblichtRecommenderTraits["ignoreCase"])`

And

`Unrecognized field "ignoreCase" (class de.tudarmstadt.ukp.inception.recommendation.imls.external.ExternalRecommenderTraits), not marked as ignorable (2 known properties: "remoteUrl", "trainable"]) at [Source: (String)"{"ignoreCase":false}"; line: 1, column: 20] (through reference chain: de.tudarmstadt.ukp.inception.recommendation.imls.external.ExternalRecommenderTraits["ignoreCase"])`

And

`Unrecognized field "ignoreCase" (class de.tudarmstadt.ukp.inception.recommendation.imls.lapps.traits.LappsGridRecommenderTraits), not marked as ignorable (one known property: "url"]) at [Source: (String)"{"ignoreCase":false}"; line: 1, column: 20] (through reference chain: de.tudarmstadt.ukp.inception.recommendation.imls.lapps.traits.LappsGridRecommenderTraits["ignoreCase"])`

**To Reproduce**

tbd (Kibana)

**Expected behavior**
No error

**Please complete the following information:**
 - Version and build ID: 0.19.6

reckart:
I wonder how this "ignoreCase" property got there. Sounds like we try loading the JSON traits of a particular recommender via the traits object of another recommender. I guess the case is something like this

* save recommender of type X
* then change recommender to a different type

if at that point the traits editor is rebuild and loading re-loading the traits from the JSON, it may crash.

I suppose all our traits classes should be annotated with `@JsonIgnoreProperties(ignoreUnknown = true)`.

reckart:
In most of the traits-based classes, I think we never hit this because we normally do not allow the "type" of a thing (e.g. of a feature, layer, etc). to be changed post-hoc. Another approach might be to lock the "type" of the recommender after the recommender has been created (i.e. when the id != null).

PR #2366, opened 2021-06-10
#2358 - Unrecognized fields when switching recommender type

jcklie:
**What's in the PR**
We now ignore unknown fields in recommender traits to not fail when switching recommender types and then reusing old traits with potentially unknown fields in them.

**How to test manually**
* Change from a saved string matcher to a remote one and see whether you get the error

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart (approved):

reckart:
Thx!

Issue #2377, opened 2021-06-14
Using activity link after layer creation NPE


UWinch:
**Describe the bug**
There will be a nullpointer exception when visiting a last seen annotation document via activity link if a new layer was created since the last visit:

<details>

```
2021-06-14 17:33:43 ERROR [admin] ApplicationPageBase - admin: Error: NullPointerException: 
2021-06-14 17:33:43 ERROR [admin] AjaxComponentRespondListener - Unable to trigger respond listener callback
java.lang.reflect.UndeclaredThrowableException: null
	at com.sun.proxy.$Proxy235.getAnnotationIndex(Unknown Source) ~[?:?]
	at org.apache.uima.fit.util.CasUtil.selectCovered(CasUtil.java:609) ~[uimafit-core-3.2.0.jar:3.2.0]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.rendering.SpanRenderer.render(SpanRenderer.java:105) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.rendering.PreRendererImpl.render(PreRendererImpl.java:102) ~[classes/:?]
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
	at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
	at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
	at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
	at org.apache.wicket.proxy.LazyInitProxyFactory$JdkHandler.invoke(LazyInitProxyFactory.java:521) ~[wicket-ioc-9.3.0.jar:9.3.0]
	at com.sun.proxy.$Proxy233.render(Unknown Source) ~[?:?]
	at de.tudarmstadt.ukp.clarin.webanno.api.annotation.AnnotationEditorBase.render(AnnotationEditorBase.java:156) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.brat.annotation.BratAnnotationEditor.render(BratAnnotationEditor.java:762) ~[classes/:?]
	at de.tudarmstadt.ukp.clarin.webanno.brat.annotation.BratAnnotationEditor.bratRenderCommand(BratAnnotationEditor.java:678) ~[classes/:?]
[...]

PR #2381, opened 2021-06-15
#2377 Using activity link after layer creation NPE

UWinch:
**What's in the PR**
* typesystem.getlayer does not throw an exception when the type is not found instead it returns null, therefore check for null so types are not rendered if they are not in the typesystem 

**How to test manually**
* see issue


UWinch:
@reckart can you have a look if this still works as intended in https://github.com/inception-project/inception/pull/2205

reckart:
It definitely looks very reasonable. The `catch(IllegalArgumentException)` would be correct if we were using the `CasUtil.getType()` from uimaFIT, but not if we simply fetch the type from the type system... thanks for looking into this!

reckart:
However... there is one question though... why isn't the CAS upgraded when opening a document via the activity history? I would have expected that when clicking on the activity link, essentially the same things happen as when opening a document via the "open document" dialog - and that includes a CAS upgrade with the current project typesystem. And if that upgrade is performed, then the error about not finding types shouldn't happen either.

UWinch:
Yes, I opened another issue for this https://github.com/inception-project/inception/issues/2380. This also happens in the 0.19.6 version but I have not found the problem yet.

Issue #2382, opened 2021-06-16
Allow unlocking of liquibase from command line/tray icon


jcklie:
**Is your feature request related to a problem? Please describe.**
Sometimes the liquibase migration can lock itself. Then, one needs to release the liquibase lock. Right now, it requires an SQL editor.

**Describe the solution you'd like**
Allow unlocking of liquibase from command line/tray icon

https://forum.liquibase.org/t/lock-problems-with-hsqldb/2765
https://docs.liquibase.com/commands/community/releaselocks.html
https://github.com/liquibase/liquibase/blob/9fd6ea48f789da2e1c0551e0e5c86cd0d25825b5/liquibase-core/src/main/java/liquibase/integration/commandline/Main.java#L1703

reckart:
For the moment, I'd tend to allow it only from the command line - e.g.

<img width="742" alt="Screenshot 2021-06-16 at 22 42 49" src="https://user-images.githubusercontent.com/1410238/122291101-ba6d4f00-cef4-11eb-997e-634050fb7a98.png">

I wonder - when the user requests to release the lock, should we only release the lock and then immediately shut down again to force the user to not simply run INCEpTION always with lock-force-removal enabled?

PR #2384, opened 2021-06-16
#2382 - Allow unlocking of liquibase from command line/tray icon

reckart:
**What's in the PR**
- Hook in between DB initialization and Liquibase
- If there is a lock, generate an exception immediately
- Provide helpful message how to recover when running in standalone mode
- Actual unlocking when user requests is is not yet done
- Added system property which can be set to force-remove the lock

**How to test manually**
* Run INCEpTION with MySQL (for easier DB access)
* Manually enter locking information in the table `DATABASECHANGELOGLOCK`
* Start the application in desktop mode and see message
* Start again with `-DforceReleaseLock=true` and see message
* Start a second time with `-DforceReleaseLock=true` and see message
* Start normally

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-app-webapp/src/main/java/de/tudarmstadt/ukp/inception/app/config/LiquibaseLockManager.java
+++ b/inception/inception-app-webapp/src/main/java/de/tudarmstadt/ukp/inception/app/config/LiquibaseLockManager.java
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Technische Universität Darmstadt under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The Technische Universität Darmstadt 
+ * licenses this file to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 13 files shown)


inception/inception-app-webapp/src/main/java/de/tudarmstadt/ukp/inception/app/config/LiquibaseLockManager.java
```diff
@@ -0,0 +1,118 @@
...
+            log.error("Unable to list Liquibase locks: ", e.getMessage());
+        }
+    }
+
+    public List<DatabaseChangeLogLock> listLocks()
+        throws DatabaseException, LockException, SQLException
+    {
+        Database database = null;
+        try {
+            Connection c = dataSource.getConnection();
```
jcklie, line 81: Is this not auto closable ? ;(

reckart: Unfortunately, no.

Issue #2252, opened 2021-05-02
Issue with MTASDocumentIndexing tasks interaction


reckart:
**Describe the bug**
A clear and concise description of what the bug is. Maybe related to recent switch to the `SchedulerService` in the `SearchService`. Maybe some kind of race-condition.

**To Reproduce**
Unit test seems to fail spuriously: `de.tudarmstadt.ukp.inception.search.index.mtas.MtasDocumentIndexTest.testLimitQueryToDocument`

**Expected behavior**
Should never fail.

**Screenshots**
```
Expecting:
  [SearchResult[documentId=2,documentTitle=Raw text document 1,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Galicia is],
    SearchResult[documentId=3,documentTitle=Raw text document 2,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Portugal is],
    SearchResult[documentId=3,documentTitle=Raw text document 2,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Portugal is]]
to contain exactly in any order:
  [SearchResult[documentId=2,documentTitle=Raw text document 1,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Galicia is],
    SearchResult[documentId=3,documentTitle=Raw text document 2,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Portugal is]]
but the following elements were unexpected:
  [SearchResult[documentId=3,documentTitle=Raw text document 2,tokenStart=1,tokenLength=1,offsetStart=4,offsetEnd=11,leftContext=The ,text=capital,rightContext= of Portugal is]]
when comparing values using field/property by field/property comparator on all fields/properties
Comparators used:
- for elements fields (by type): {Double -> DoubleComparator[precision=1.0E-15], Float -> FloatComparator[precision=1.0E-6]}
- for elements (by type): {Double -> DoubleComparator[precision=1.0E-15], Float -> FloatComparator[precision=1.0E-6]}
	at de.tudarmstadt.ukp.inception.search.index.mtas.MtasDocumentIndexTest.testLimitQueryToDocument(MtasDocumentIndexTest.java:392)
Standard Output
[...]

PR #2460, opened 2021-07-06
#2252 - Issue with MTASDocumentIndexing tasks interaction

reckart:
**What's in the PR**
- Better control of the interaction between incoming and already enqueued tasks
- Incoming re-indexing tasks trump enqueued (re)indexing tasks
- Incoming source/anno indexing tasks are queued after re-indexing tasks

**How to test manually**
* Try stressing the indexing / search functionality

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotatorStateUtils.java
+++ b/inception/inception-api-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/api/annotation/model/AnnotatorStateUtils.java
@@ -17,7 +17,10 @@
  */
 package de.tudarmstadt.ukp.clarin.webanno.api.annotation.model;
 
+import static org.apache.commons.lang3.time.DurationFormatUtils.formatDurationHMS;
+
 import java.io.IOException;
+import java.text.SimpleDateFormat;
 import java.util.Optional;
 
 public class AnnotatorStateUtils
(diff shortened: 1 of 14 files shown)


antonyscerri:
I tried the PR and i still saw duplicate search results. With a corpus of 10 document i was search a phrase with several matches in a number of the document (total 13 hits), in some cases multiple matches per document. I think marked one span, then clicked the create button to mark the others. Hit the search button a few times until it finished indexing and then used the delete button to remove them, running the search. I redid this a couple of times, occasionally clicking a result to switch to a different instance. On one occasion i ended up with 14 results. 

I confirmed the document was indexed twice by searching for other phrases in it and getting identical duplicate hits, to ensure it wasnt some kind of false caching situation.

When the extra hits showed up i was able to click the reindex option and wait and then got back the expected 13 results. It was not easy to reproduce it compared to the previous code so i think its probably better. I was trying to pin down a shorter reproducible sequence and i may have found it. I believe if you create or delete a set of annotations matching a query, and right after clicking either button you click on a search result for a document other than the one currently being viewed this may cause the problem. In theory this will have triggered the indexing due to the bulk operation AND attempted a second one as part of the CAS upgraded when trying to view the new document.

I dont know if this is relevant but during the indexing because i still see the errors about missing labels i noted the different in some log lines showing the current username, and others showing SYSTEM:

2021-07-06 20:25:29 ERROR [tony] ConceptFeatureSupport - No label for feature value [http://www.wikidata.org/entity/Q948441]
2021-07-06 20:25:32 ERROR [SYSTEM] ConceptFeatureSupport - No label for feature value [http://www.wikidata.org/entity/Q5090513]
2021-07-06 20:25:32 ERROR [SYSTEM] ConceptFeatureSupport - No label for feature value [http://www.wikidata.org/entity/Q392003]
2021-07-06 20:25:35 ERROR [tony] ConceptFeatureSupport - No label for feature value [http://www.wikidata.org/entity/Q1271432]

This used to occur more often with the old code, and i'm pretty sure i saw one such mix when i followed that short sequence of actions.

Issue #2467, opened 2021-07-09
Search Delete of Stacked Annotations


antonyscerri:
**Describe the bug**
If you have multiple annotations in the same layer with a different value attribute covering the same span of text, so stacked. You are able to search for the individual ones using the value feature. However if you try to then delete them it depends on their stacked order whether it will work or not, the one at the bottom seems to be the one you can delete (which corresponds to the first added to the document).  

**To Reproduce**
1. Open a project with a Named Entity layer, with a value feature and corresponding tagset with multiple values
2.Open a document, find a span of text without any annotations, to keep it simple you may want a single document and/or unique string
3. Create an annotation over the chosen span, and select your first value feature tag
4. Create a second annotation over the same chosen span, and select your second value feature tag
5. Run a search for the given phrase and the second value feature tag, and confirm you get the hit (note it is unnecessary to match on the layer feature value as the following steps only consider the matching spans as the hit and not the specific elements in the query).
6. Click the corresponding annotation marker (for the second value feature tag)
7. Click the delete button, nothing should happen
8. Repeat 5-7 using the first value feature tag and this time the annotation should get deleted.

**Expected behavior**
You should be able to delete the matched annotation as per the selected one when there are multiple annotations at the same location. This is assuming as noted above the query hits used for the selection only consider the span locations and not what else was used in the query wrt to matching any annotation, that relies solely on the selected annotation. 

**Please complete the following information:**
 - Version and build ID: 0.19.3 (2021-04-25 09:58:11, build 238b0e91)

antonyscerri:
I believe the cause may lie with:
https://github.com/inception-project/inception/blob/f3bed56c384d1fbded0a8d50d2eb46279ee23a68/inception/inception-ui-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/search/sidebar/SearchAnnotationSidebar.java#L642

It appears to be selecting the first only matching annotation.

reckart:
Yep, looks like - are you interested to submit a PR and in a position that would allow you doing so?

antonyscerri:
Possibly, i haven't checked out the project development guidelines but just looking over the code it didnt appear there were necessarily test coverage for everything so i'm not sure how much that would entail (the fix itself should be fairly minimal). The main focus right now was completing an annotation exercise but having encountered these issues pinning down the cause let me look into immediate work arounds where possible whilst remaining on the current stable release. I'll have a look this weekend at the dev guidelines and see about a PR.

[...]

Issue #2442, opened 2021-07-03
Duplicates created from search matches when layer uses link feature


antonyscerri:
**Describe the bug**
When creating multiple annotations from a search (based on words) and the layer being used for the selected annotation is configured to use a link feature it creates two annotations per span in the search results. In the case of the selected annotation it also gets two new instances. If you enable "Override existing Annotations" then you get a single new instance and the selected one gets no additional ones.

When you try the delete you have to disable the "Delete only matching feature values" otherwise it fails to match the newly created ones.

**To Reproduce**
1.Create a layer using a link feature
2. Open a document with multiple instance of a word
3. Select the layer with the link feature
4. Mark the first instance of the word (all values can be left as-is)
5. Open search and search on the chosen word
6. Click "create"
7 Clicking create additional times will create 2 more instances per match

**Expected behavior**
The same behaviour as when you do not use a link feature

**Please complete the following information:**
 - Version and build ID: 0.19.3 (2021-04-25 09:58:11, build 238b0e91)

**Additional context**
I suspect its some kind of equality check on the object when using a link feature, although this doesn't quite explain why you get double every time unless its causing some issue with the search and matching results.

**Todo**
- [ ] Do not consider link features when comparing annotations for bulk operations
- [ ] Do not fill in link features in bulk-created annotations

antonyscerri:
It should also be noted if you create several sets of annotations, click create twice for example. If you then search click on one of the newly created annotations (pick the bottom one in the stack for example). Then disable "matching feature values" option and click delete you will notice that a number of them will be removed per match but not all. You can then click a remaining one and then click delete again to remove another set.

antonyscerri:
I've setup a local dev env and tried using the latest code. I was able pin it down more precisely whilst setting things up, it turns out the issue occurs when creating annotations where the layer has a link feature WITH a default slot value defined. Which in the annotation editor then displays "Click to activate" against the default slot role, which you do not have to have selected anything for. So in step one above it also requires defining the tag set and picking a default slow value.

Also whilst testing the indexing seemed to fail and get stuck whilst I was testing it. This only occurred once out of three startups and testing runs i went through to pin down the issue.

org.apache.lucene.store.LockObtainFailedException: Lock held by this virtual machine: /Users/scerria/.inception/repository/project/0/indexMtas/write.lock
[...]

PR #2476, opened 2021-07-12
Bugfix for #2467 and partially addresses #2442

antonyscerri:
**What's in the PR**
This is a fix for bug #2467 and it also partially address #2442 (I realised it also only looked at the first annotation and so would create a duplicate even if one of the others was a match).

**How to test manually**
Assuming you use the named entity layer. Using a search for words you should be able to create multiple sets using a different value feature tag, with repeated calls with the same tag not creating duplicates. You will also then be able to delete any matching annotation using any of the tag values you have created (assuming you have more than one stacked up).

Note the exact behaviour on creating multiple annotations when one exists with matching values alongside the override isn't really defined well so I went with just trying to avoid creating duplicates when not overriding. However if override is on then it will set all matches to the template annotations as before, this may result in duplicates if there were multiple distinct ones beforehand.

**Automatic testing**
No matching test class

### DIFF
--- a/inception/inception-ui-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/search/sidebar/SearchAnnotationSidebar.java
+++ b/inception/inception-ui-search/src/main/java/de/tudarmstadt/ukp/inception/app/ui/search/sidebar/SearchAnnotationSidebar.java
@@ -36,6 +36,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.uima.cas.CAS;
 import org.apache.uima.cas.Type;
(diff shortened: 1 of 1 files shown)


reckart:

reckart:

reckart:

reckart:
I tried to reproduce  #2442 on this branch, but it doesn't seem to happen here. Why do you say that  #2442 is only partially addressed?

I also noted that if the span with the link feature has already filled slots, then the copies also end up with the same slot fillers - and I wonder if this is a good idea or not...

antonyscerri:
Its partially addressed #2442 because i added the loop to check if an existing annotation at the given span matched and then it won't create one. This did not explicitly address the issue of using a link feature and the related matching check with FSArray (which with a default slot selected will always be an intance, not null, with zero length unless the user has actually picked an annotation to link to). So I basically added the same looping check as in the delete for completeness, with the caveat around the exact behaviour around overriding and if a link feature is used whether that should even be copied (whether it should be removed, or just not allowed).

antonyscerri:
I should have also mentioned that same change to the create behaviour will prevent duplicates in other cases now.

reckart:

reckart:

Issue #2455, opened 2021-07-06
Collect project corpus statistics


jcklie:
**Is your feature request related to a problem? Please describe.**
I want to see statistics about my corpus. This should be on document level and aggregated for the whole project. Interesting would be e.g.

**Describe the solution you'd like**

Per document:
- Number of tokens
- Number of sentences
- Number of annotations per layer
- Number of annotations per layer grouped by feature value

For these, min/max/median/five-number summary/average/standard deviation would be nice

Then aggregate them to project level.

-----

Other ideas:

- Number of tokens per sentence
- Number of annotations per sentence

jcklie:
I talked with Richard and we came to the conclusion that the way is not going to collect statistics over UIMA, but via the search that has already indexed the annotations. There is a description in [here](https://textexploration.github.io/mtas/search_component_stats_tokens.html) .

An example in how to do that in code is [here](https://textexploration.github.io/mtas/installation_lucene.html] *Advanced search**. We use Lucene in INCEpTION, not SOLR. The goal is to have a function in the search service similar to `query` called `queryStatistics` that gets not the query results, but the statistics of them.

I would do the following: 

1. Copy the test function from  `de.tudarmstadt.ukp.inception.search.index.mtas.MtasDocumentIndexTest.testAnnotationQuery()`and call it something like `testAnnotationQueryStatistics`
2. To get started. add a function to the search service called something like `getTokenPerDocumentStatistics` and implement it like the query method, you need to plumb it/create copies of functions all the way until you reach the `MtasDocumentIndex ` where the query is actually executed.
3. Then you can use the lucene code above to get the statistics from the Lucene `ComponentToken` .

When that works, you can try to add another method to the search service for other statistics. Once you get a feel for how it works, you can try to unify it into one `queryStatistics` method.

@reckart Did I understand it correctly?

reckart:
Sounds good. And open a PR and commit frequently so we can keep in touch about the approaches taken.

xegulon:
Having the live statistics as we annotate would be great.
For example, I'm annotating sentences for a classifier, and I would like to have the numbers and proportions of the different classes in my dataset, as I annotate.

PR #2480, opened 2021-07-13
#2455 - Corpus Statistics via Search

Falko1:
A new approach for getting the project statistics, this time via the search functionality.  This adds the backend functionality.

**Automatic testing**
* [x] PR includes unit tests


inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
```diff
@@ -108,6 +109,18 @@ long determineNumOfQueryResults(User aUser, Project aProject, String aQuery,
...
+    /*
+    public HashMap<String, HashMap<String, Double>> getProjectTextStatistics(User aUser,
+            Project aProject, String aStatistic, SourceDocument aDocument,
+            AnnotationLayer aAnnotationLayer, AnnotationFeature aAnnotationFeature,
+            String[] aPunctuationMarks)
+        throws IOException, ExecutionException;
+
+     */
+
+    public Map<String, Map<String, Object>> getProjectStatistics(User aUser, Project aProject, String aStatistic,
```
jcklie, line 121: I would create a class for this `Map<String, Map<String, Object>>` instead of using this nested type

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
```diff
@@ -108,6 +109,18 @@ long determineNumOfQueryResults(User aUser, Project aProject, String aQuery,
...
+             Double aLowerDocSize, Double aUpperDocSize)
```
jcklie, line 122: Why are these Double and not double?

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticRequest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public StatisticRequest(Project aProject, User aUser, String aStatistic)
+    {
+        this(aProject, aUser, aStatistic, null, null);
+    }
+
+    public StatisticRequest(Project aProject, User aUser, String aStatistic,
+            Double aLowerDocumentSize, Double aUpperDocumentSize)
+    {
+        super();
```
jcklie, line 40: I dont think that you need the super()

jcklie (commented):
Looks very nice so far, it would be good maybe if you also had a function in the search service that could get statistics for a query, not only the aggregated project statistics.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
Falko1: By accident. Changed it to double

Falko1: Well, just realised that it is actually needed. To also consider unbounded documents the mtas method requires null for upperDocSize. Thus we need Double and not double.

reckart: It seems quite curious that a document size can be a float number - any idea what unit the document size is measured in? There is also OptionalDouble that could be considered here.

Falko1: I'm pretty sure that the document number is the number of tokens of a document :
grafik
( taken from the mtas documentation https://textexploration.github.io/mtas/search_component_stats_tokens.html )

I need this method which wants a Double; a Null indicates unboundedness in this direction.
grafik

reckart: You do no necessarily need to use a Double on our own method signatures. E.g. you could use an `OptionalInt aMinTokens` and instead of passing this value directly to the MTAS method, you could do `Double minTokens = aMinTokens.isPresent() ? (double) aMinTokens.get() : null`

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
         return _executeQuery(this::doCountResults, aRequest);
     }
 
+    @Override
+    public StatisticsResult getAnnotationStatistics(StatisticRequest aStatisticRequest)
+        throws IOException, ExecutionException
+    {
+        ArrayList<Integer> fullDocSet = getUniqueDocuments(aStatisticRequest);
```
jcklie, line 409: List

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
Falko1: I can do it this way if you want. But what is the additional value of using OptionalInt over Double? For me, it seems to be more complicated.

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+                    }
+                }
+                // source document without annotation layer? then user is not relevant
+                else if (!annotatableDocuments.containsKey(Long.valueOf(sourceID))) {
+                    fullDocSet.add(i);
+                }
+            }
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
jcklie, line 504: This needs a bit more care, swallowing exceptions is not good

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+                // method
+                // getSearcherManager().release(searcher);
+                getSearcherManager().release(searcher);
+                searcher = null;
+            }
+        }
+        return fullDocSet;
+    }
+
+    public MtasSpanQuery parseQuery(String query) throws ExecutionException, Exception
```
jcklie, line 519: aQuery

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+            // The exceptions thrown by the MTAS CQL Parser are inheriting from
+            // java.lang.Error...
+            throw new ExecutionException("Unable to parse query [" + query + "]", e);
+        }
+        return mtasSpanQuery;
+    }
+
+    @Override
+    public Map<String, Double> getLayerStatistics(StatisticRequest aStatisticRequest,
+            String aFeatureQuery, ArrayList<Integer> fullDocSet, boolean perSentence)
```
jcklie, line 541: aFullDocSet, aPerSentence

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+        Map<String, Object> resultsMap = null;
+        Map<String, Double> resultsMapDouble = new HashMap<String, Double>();
+        Map<Long, Long> annotatableDocuments = listAnnotatableDocuments(
+                aStatisticRequest.getProject(), aStatisticRequest.getUser());
+        try {
+            searcher = getSearcherManager().acquire();
+            IndexReader reader = searcher.getIndexReader();
+
+            // what does this parameter do?
+            ArrayList<Integer> fullDocList = new ArrayList<Integer>();
```
jcklie, line 554: List = ArrayList

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+            MtasSpanQuery[] spanQuerys = new MtasSpanQuery[1];
+
+            String[] functionKeys = new String[1];
+            String[] functions = new String[1];
+            String[] functionTypes = new String[1];
+
+            MtasSpanQuery parsedQuery = parseQuery(aFeatureQuery);
+            fieldStats.spanQueryList.add(parsedQuery);
+
+            if (perSentence == false) {
```
jcklie, line 566: !perSentence

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+            resultsMap = results.rewrite(true);
+            for (String str : resultsMap.keySet()) {
+                if ((resultsMap.get(str) instanceof Number)) {
+                    resultsMapDouble.put(str, Double.valueOf(resultsMap.get(str).toString()));
+                }
+            }
+            return resultsMapDouble;
+        }
+        catch (Exception e) {
+            e.printStackTrace();
```
jcklie, line 606: Exception handling

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -909,6 +1130,15 @@ private void indexDocument(String aDocumentTitle, long aSourceDocumentId,
         doc.add(new StringField(FIELD_TITLE, aDocumentTitle, Field.Store.YES));
         doc.add(new StringField(FIELD_USER, aUser, Field.Store.YES));
         doc.add(new StringField(FIELD_TIMESTAMP, timestamp, Field.Store.YES));
+        // FieldType fieldTypeForContent = new FieldType();
```
jcklie, line 1133: What is this?

inception/inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -456,6 +526,51 @@ public void testAnnotationQuery() throws Exception
...
+
+        String query = "moon";
+
+        StatisticsResult statsResults = searchService.getProjectStatistics(user, project, statistic,
+                lowerDocSize, upperDocSize);
+
+        StatisticsResult queryStatsResults = searchService.getQueryStatistics(user, project,
+                statistic, query, lowerDocSize, upperDocSize);
+
+        assertThat(0).isEqualTo(0);
```
jcklie, line 571: Is that intended? I dont see assertions here

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Falko1: Then I need to do this cast because the mtas method requires ArrayList:
grafik

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Falko1: Forgot to delete code from my first approach.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
reckart: It is. But the `Double` raises many questions such as "what is the unit?" and "why can it be null?". If the unit is a token, then it raises the question "why can there be fractions of a token?". An `OptionalInt aMaxTokens`would not raise these questions, because the unit is clear, it is clear that the limit is optional, and there is no question about fractions of tokens.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/SearchService.java
```diff
@@ -109,5 +109,13 @@ long determineNumOfQueryResults(User aUser, Project aProject, String aQuery,
             AnnotationFeature aAnnotationFeature)
         throws ExecutionException, IOException;
 
+    public StatisticsResult getProjectStatistics(User aUser, Project aProject, String aStatistic,
+            Double aLowerDocSize, Double aUpperDocSize)
+        throws IOException, ExecutionException;
+
+    public StatisticsResult getQueryStatistics(User aUser, Project aProject, String aStatistic,
```
jcklie, line 116: Is it possible to add some documentation what this method does? Especially the Double are difficult to understand

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticsResult.java
```diff
@@ -0,0 +1,97 @@
...
+    private Double upperDocSize;
+    private Double lowerDocSize;
+    private User user;
+    private Project project;
+
+    private Map<String, Map<String, Double>> allResults;
+    private Map<String, Map<String, Double>> nonTrivialResults;
+
+    public StatisticsResult(StatisticRequest aStatisticRequest,
+            Map<String, Map<String, Double>> allResults,
```
jcklie, line 38: Do we know which metrics we have? If yes, we could accessors here to get a statistic for a certain metric, e.g. getMean(String layerName, String featureName) or so.

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
         return _executeQuery(this::doCountResults, aRequest);
     }
 
+    @Override
+    public StatisticsResult getAnnotationStatistics(StatisticRequest aStatisticRequest)
+        throws IOException, ExecutionException
+    {
+        ArrayList<Integer> fullDocSet = getUniqueDocuments(aStatisticRequest);
+        Map<String, Map<String, Double>> allStats = new HashMap<String, Map<String, Double>>();
```
jcklie, line 410: Do we know which metrics we have? If yes, we could use an object here that wraps all these metrics instead of using a Map<String, Double>

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+        Map<String, Map<String, Double>> nonNullStats = new HashMap<String, Map<String, Double>>();
+        List<AnnotationLayer> layers = schemaService
+                .listAnnotationLayer(aStatisticRequest.getProject());
+        List<AnnotationFeature> features;
```
jcklie, line 414: Why do you declare `features` outside the loop?

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+
+        return new StatisticsResult(aStatisticRequest, allStats, nonNullStats);
+    }
+
+    @Override
+    public ArrayList<Integer> getUniqueDocuments(StatisticRequest aStatisticRequest)
+        throws IOException
+    {
+        IndexSearcher searcher = null;
+        Map<Long, Long> annotatableDocuments = listAnnotatableDocuments(
```
jcklie, line 479: What is this mapping and why is it Long, Long?

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+        ArrayList<Integer> fullDocSet = new ArrayList<Integer>();
+
+        try {
+            searcher = getSearcherManager().acquire();
+            IndexReader reader = searcher.getIndexReader();
+
+            for (int i = 0; i < reader.maxDoc(); i++) {
+                String sourceID = reader.document(i).get(FIELD_SOURCE_DOCUMENT_ID);
+                String annotationID = reader.document(i).get(FIELD_ANNOTATION_DOCUMENT_ID);
+                if (Long.valueOf(annotationID) != -1L) {
```
jcklie, line 491: Can you write 

if (Long.valueOf(annotationID) == -1L) {
continue;
}
?

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -396,6 +402,221 @@ public long numberOfQueryResults(SearchQueryRequest aRequest)
...
+        Map<String, Double> resultsMapDouble = new HashMap<String, Double>();
+        Map<Long, Long> annotatableDocuments = listAnnotatableDocuments(
+                aStatisticRequest.getProject(), aStatisticRequest.getUser());
+        try {
+            searcher = getSearcherManager().acquire();
+            IndexReader reader = searcher.getIndexReader();
+
+            // what does this parameter do?
+            ArrayList<Integer> fullDocList = new ArrayList<Integer>();
+            CodecComponent.ComponentField fieldStats = new CodecComponent.ComponentField(
```
jcklie, line 555: Can you maybe add some comments how this works and maybe link to the documentation of MTAS?

jcklie:
Build is failing

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticsResult.java
Falko1: We do not know them apriori, but we can access them after a StatisticsResult object is created. I wrote a method getStatistic(String aMetric, String aLayerName, String aFeatureName) where the statistic is an input, too.

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Falko1: I wanted to make the code flexible, so the user can decide which metrics should be calculated. Alternatively, we can always calculate all metrics but this could be problematic for the speed.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticsResult.java
jcklie: I thought that we know what metrics the MTAS can give us, so we do not need a map but just can use a data object. Can you explain why we do not know the metrics ahead of time?

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Falko1: listAnnotatableDocuments has been there before (MtasDocumentIndex around line 800). I assume that the document IDs can become so large that int will not always suffice. 
The keys of annotatableDocuments are the IDs of source documents which have a annotation document linked to them. The ID of the annotation document is the corresponding value of this key.

inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
Falko1: No, I can't. In the case of Long.valueOf(annotationID) == -1L we can not simply jump to the next document. We know then that this is a source document. This will not be included in fullDocSet, unless it has no annotation document linked to it. This special case is checked by the else if statement.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticsResult.java
Falko1: Also see my comment on the next question.
Yes, we know what metrics MTAS can give us. 
I imagined it like this: the user ticks boxes in inception to choose which metrics should be calculated for her. These boxes translate to a regex which tells MTAS which statistics to calculate. 
I have no idea if runtime speed is an issue here at all (I only tested with 10-word corpora). If it is no issue, then we can always compute all statistics and only show those to the user which she actually wants to see. 
I think my current approach is more flexible but maybe this flexibility is unnecessary.

jcklie: Can you maybe test it on a large corpus? We have a large project generator somewhere 
@reckart Do you know where the script is we use to generate and test large projects?

reckart: @UWinch did you use a generator script to build the "tiny" project?

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
```diff
@@ -0,0 +1,157 @@
...
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.inception.search;
+
+public class LayerStatistics
+{
+    /** The stats to be calculated */
+    public static final String STATS = "n,sum,min,max,mean,median,standarddeviation";
+
+    private long total;
```
jcklie, line 25: These can all be final I think

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
```diff
@@ -0,0 +1,157 @@
...
+
+    private double maximumPerSentence;
+    private double minimumPerSentence;
+    private double meanPerSentence;
+    private double medianPerSentence;
+    private double standardDeviationPerSentence;
+
+    private long noOfDocuments;
+
+    public LayerStatistics()
```
jcklie, line 40: Why the empty constructor and no setter?

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
```diff
@@ -0,0 +1,157 @@
...
+        maximumPerSentence = aMaximumPerSentence;
+        minimumPerSentence = aMinimumPerSentence;
+        meanPerSentence = aMeanPerSentence;
+        medianPerSentence = aMedianPerSentence;
+        standardDeviationPerSentence = aStandardDeviationPerSentence;
+
+        noOfDocuments = aNoOfDocuments;
+    }
+
+    public boolean isLayerTrivial()
```
jcklie, line 65: Can you add a comment what this does / why its needed?
also, its just `return maximum == 0L;`

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public long getNoOfDocuments()
+    {
+        return noOfDocuments;
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (o == this) {
```
jcklie, line 136: I would use intellij to generate it with equalsbuilder instead and also would override hashCode at the same time.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
```diff
@@ -0,0 +1,157 @@
...
+                && (ls.getMinimum() == getMinimum()) && (ls.getMedian() == getMedian())
+                && (ls.getMean() == getMean())
+                && (ls.getStandardDeviation() == getStandardDeviation())
+                && (ls.getNoOfDocuments() == getNoOfDocuments())
+                && (ls.getMaximumPerSentence() == getMaximumPerSentence())
+                && (ls.getMinimumPerSentence() == getMinimumPerSentence())
+                && (ls.getMedianPerSentence() == getMedianPerSentence())
+                && (ls.getMeanPerSentence() == getMeanPerSentence())
+                && (ls.getStandardDeviationPerSentence() == getStandardDeviationPerSentence()));
+    }
```
jcklie, line 156: I would save layer and feature name also in this to make it simpler to find out for what these stats are. Maybe you do not need it, then you can ignore my comment

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    private List<AnnotationFeature> features;
+    private String query;
+
+    public StatisticRequest(Project aProject, User aUser)
+    {
+        this(aProject, aUser, null, null, new ArrayList<AnnotationFeature>(), null);
+    }
+
+    public StatisticRequest(Project aProject, User aUser, OptionalInt aMinTokenPerDoc,
+            OptionalInt aMaxTokenPerDoc, List<AnnotationFeature> aFeatures, String aQuery)
```
jcklie, line 45: I think OptionalInt should not be null but OptionalInt.empty()

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticRequest.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public List<AnnotationFeature> getFeatures()
+    {
+        return features;
+    }
+
+    public void addFeature(AnnotationFeature aFeature)
+    {
+        if (features.contains(aFeature)) {
```
jcklie, line 83: ? You could also use a set here instead of a list

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticsResult.java
```diff
@@ -0,0 +1,193 @@
...
+
+import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationFeature;
+import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationLayer;
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+public class StatisticsResult
+{
+
+    private OptionalInt minTokenPerDoc;
```
jcklie, line 32: These can all be final?

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/StatisticRequest.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.List;
+import java.util.OptionalInt;
+
+import de.tudarmstadt.ukp.clarin.webanno.model.AnnotationFeature;
+import de.tudarmstadt.ukp.clarin.webanno.model.Project;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+public class StatisticRequest
+{
+    private final Project project;
```
jcklie, line 30: These can mostly be final?

inception/inception-search-mtas/src/test/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndexTest.java
```diff
@@ -456,6 +536,192 @@ public void testAnnotationQuery() throws Exception
...
+        await("Waiting for indexing process to complete") //
+                .atMost(60, SECONDS) //
+                .pollInterval(5, SECONDS) //
+                .until(() -> searchService.isIndexValid(aProject)
+                        && searchService.getIndexProgress(aProject).isEmpty());
+    }
+
+    // executing this method takes around 30 minutes!
+    @Test
+    public void testStatisticsWithRealProject() throws Exception
```
jcklie, line 675: Please add an @Ignore here or better delete the test, because noone except you has the data. You could also instead use data from the internet,see e.g. for recommender tests that load it via DKPro

jcklie (changes requested):
Still needs some work, but it is almost done.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
Falko1: This is just a kind of workaround. I need to access STATS in mtasDocumentIndex to compute the values to put in the actual constructor. Therefore I added this empty constructor. Maybe there is a smarter way to do this.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
Falko1: Well, if the maximum is 0 then there is no document which has an annotation from this feature so all statistics for this feature are 0. 
It was useful for the previous approach where I cycled through all of the features. I can remove it together with the nonTrivialResults from StatisticsResult.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
jcklie: STATS is static final, you why cant you write `LayerStatistics.STATS`?

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
Falko1: LayerStatistics is just a helper class. Everything in it is only accessed through StatisticsResult where layer and feature are saved.

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
Falko1: You are right, missed this somehow...

inception/inception-search-core/src/main/java/de/tudarmstadt/ukp/inception/search/LayerStatistics.java
Falko1: Removed it.

jcklie (approved):
Looks good to me, @reckart  wdyt?

jcklie:
@Falko1 Maybe the build is failing

Issue #2534, opened 2021-08-05
Add word conversion colab notebook


UWinch:
**Is your feature request related to a problem? Please describe.**
Add an example of converting .docx files with comments as annotations to UIMA CAS Xmi, so they can be imported into INCEpTION.

reckart:
How about we add a default layer "Comment" in INCEpTION with a multi-line string feature - and then use that in the example?

UWinch:
I would like to tell people that in general, they need to first configure their custom layer. Using a default layer comment to which the comments are mapped might trick people into thinking the layers are automatically generated from the input document.

I would not mind adding a default comments layer independent from this issue though.

PR #2535, opened 2021-08-05
#2534 Add word conversion colab notebook

UWinch:
**What's in the PR**
*  Colab notebook converting .docx with comments to UIMA CAS xmi

**How to test manually**
* Run the notebook
* Import the resulting file into INCEpTION


reckart (approved):
Two minor comments:
* `UIMA CAS XMI` (in particular `XMI`) should always be written all-uppercase.
* Since comments in a document could become large, I think it would be good to suggest that the "multi-line" flag should be turned on for the "value" feature.

Issue #2522, opened 2021-08-02
Opening sidebar changes document


UWinch:
**Describe the bug**
If the annotation page was first accessed via a recent activity link and the user navigates to another document and then opens the sidebar, the original document is opened again.

**To Reproduce**
1. Follow a recent activity to the annotation page
2. Navigate to the next document
3. Open e.g. the recommendation sidebar
4. The previous document is opened

**Expected behavior**
The document should not change when the sidebar is opened.

**Please complete the following information:**
 - Version and build ID: 21.0-SNAPSHOT
 - OS: Linux
 - Browser: Chrome

UWinch:
Activity links are missing the context-path (will e.g. link to `localhost:8080` instead of `localhost:8080/myinception`).

UWinch:
Another problem is with urls that use the document name instead of the document id e.g. when importing and opening documents from document repositories via the search page

PR #2540, opened 2021-08-10
#2522 - Opening sidebar changes document

reckart:
**What's in the PR**
- Use alternative redirection solution
- Fix issue when document is reference by name instead of ID

**How to test manually**
* Open document via the recent activity (annotation or curation), switch to another document, then open sidebar (should not switch back to the first document)
* Open document via the external search

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


UWinch (approved):
Works nicely, tested with both as well. Great find with the alternative redirection exception!

Issue #2562, opened 2021-08-27
Project name no longer unique


reckart:
**Is your feature request related to a problem? Please describe.**
As a user with project-creator role, I can sometimes not create a project with the name that I want because the system claims that the project already exists - but I cannot see it.

**Describe the solution you'd like**
Considering we have a unique project URL slug now, it would be nice if the project name would no longer need to be unique so that this problem is removed.

PR #2563, opened 2021-08-27
#2562 - Project name no longer unique

reckart:
**What's in the PR**
- Drop unique index on name
- Change hashCode / equals on Project from name to slug
- Automatically add slugs to all project which do not have one yet during startup
- No longer create unique name when importing a project and the name was already taken
- No longer check if project name is already used when setting project name in project details / dashboard
- Adjustments to the remote APIs: the project "name" parameter in the remote API requests and responses is actually the URL slug and the "title" parameter is the project name. In the legacy remote API responses, the slug replaces the name.
- The first line of a project description is now shown as a short description in the project overview page

**How to test manually**
* Start up with external or embedded DB and check if the slugs are auto-generated
* Check if the slugs were actually persisted by hovering with the mouse cursor over the links in the project dashboard
* Try giving two projects the same name manually
* Export a project, then re-import it 

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie (approved):
Works for me

Issue #2513, opened 2021-07-26
Keyboard shortcut for jumping between annotations


annefried:
**Is your feature request related to a problem? Please describe.**
It is often useful to simply pre-mark annotation spans and upload these annotations for correction / validation. This certainly has overlaps with the recommender facility, but is nevertheless a common scenario. At the moment, annotators have to click the pre-marked spans before labeling them.

**Describe the solution you'd like**
A keyboard shortcut should be definable to automatically selecting the next / previous annotation on the same layer (including relations/links attached to the layer).

**Describe alternatives you've considered**
Implemented this feature in the unfortunately discontinued Swan project ;)

reckart:
@annefried The PR introduces two buttons in the right sidebar to navigate between annotations on the same layer as the currently selected annotation. These buttons have keyboard bindings (shift left/right). Works for spans and relations. Note that the order depends on the usual internal sort order (begin increasing, end decreasing) and may not coincide with the position of the labels on screen. In particular for relations, navigating to the next relation may cause the focus to actually move to the left of its current position and vice versa. Links/relations are on different layers - they are not considered in the navigation. The two requirements "on the same layer" and "follow links/relations" are contradictory in this case. I think a better keyboard-based navigation would be better to implement more on the browser/JS side maybe in a visualizer that is not brat-based.

PR #2596, opened 2021-09-13
#2513 - Keyboard shortcut for jumping between annotations

reckart:
**What's in the PR**
- Fix accumulating event listeners when using keyboard shortcuts
- Added buttons to navigate between annotations on the current layer
- Updated documentation

**How to test manually**
* Use the arrow buttons in the sidebar to navigate between annotations
* Use the SHIFT+LEFT and SHIFT+RIGHT key shortcuts to navigate between annotations

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


UWinch:
I tested with a layer layer that has a string feature without tagset. Sometimes, when moving to the next annotation, the page keeps on loading. I don't get an error message.

reckart:
I tried all kinds of different things with relations, spans, single features, multiple features, with pressing enter and without etc. but I could not reproduce your issue. I found a couple of other issues though and fixed them (not really related to this feature though). As far as I can tell, the jumping works pretty good. I even tried in different browsers. The only thing is that when I jump before pressing enter on a feature, the value is not saved. But I think that is acceptable.

Issue #2610, opened 2021-09-19
Remember sidebar state


reckart:
**Is your feature request related to a problem? Please describe.**
When I re-open a document, I always have to re-select and re-expand the sidebar.

**Describe the solution you'd like**
It would be nice if the project remembered the sidebar state.

PR #2611, opened 2021-09-19
#2610 - Remember sidebar state

reckart:
**What's in the PR**
- Store the sidebar state (selected tab and expansion state) as a per-user/per-project preference.

**How to test manually**
* Open a sidebar on the annotation page, leave the page, come back
* Try with and without leaving the sidebar in an expanded state

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/sidebar/AnnotationSidebarState.java
+++ b/inception/inception-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/sidebar/AnnotationSidebarState.java
@@ -17,33 +17,28 @@
  */
 package de.tudarmstadt.ukp.clarin.webanno.ui.annotation.sidebar;
 
-import org.apache.wicket.extensions.markup.html.tabs.AbstractTab;
-import org.apache.wicket.model.IModel;
-
-import de.agilecoders.wicket.core.markup.html.bootstrap.image.IconType;
-
-public abstract class SidebarTab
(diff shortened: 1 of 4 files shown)


inception/inception-ui-annotation/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/annotation/sidebar/AnnotationSidebarState.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package de.tudarmstadt.ukp.clarin.webanno.ui.annotation.sidebar;
+
+public class AnnotationSidebarState
+{
+    private String selectedTab;
+    private boolean expanded;
+
+    public void setSelectedTab(String aFactoryId)
```
jcklie, line 25: Why aFactoryId

reckart: The id of the AnnotationSidebarFactory that created the selected sidebar. We cannot remember the index of the open tab because that might become invalid e.g. if a user disables/enables a feature flag that causes sidebars to (dis)appear.

jcklie: Thank your for the explanation. I found it just strange that the getter variable and the member variable do not have the same name.

Issue #2598, opened 2021-09-16
Collect project corpus statistics - UI


jcklie:
**Is your feature request related to a problem? Please describe.**
We can now collect corpus statistics in the backend. We want now a UI to show these to users.

PR #2613, opened 2021-09-20
#2598 UI for the corpus statistics

Falko1:
Was: In this PR the UI for the corpus statistics will be build. It will be added to the annotation sidebar below search.

**What's in the PR**
- This PR adds a statistics sidebar to inception. It is located under Annotation directly below the search sidebar. 
- The statistics sidebar consists of two dropdowns where you can choose the granularity and the metric. 
- Next to the dropdowns there is a question mark which links to the User Guide. I added a detailed description of the sidebar in the asciidoc but it is not included yet in the official User Guide.
- A "hide empty layers"-checkbox 
- With the calculate button the calculation starts.
- The results of the calculation are shown in a table below the calculate button. It consists of 2 columns. 
- A format dropdown and an export button below the table to export the results in the desired format.

**How to test manually**
* For a detailed description, go to the User Guide -> Annotation -> Statistics (if it is updated already) or see the images below.
* Navigate to Annotation and click the clipboard icon to reach the statistics sidebar.
* Choose a granularity. per Document: the inputs for the metric are the raw counts per Document. per Sentence: before the raw counts are plugged into the metric, each of them is divided by the number of sentences in its respective documents.
* Choose a metric. They should be self-explanatory. 
* Choose whether to omit layer/features which do not have an annotation.
* Click on Calculate. This may take a while depending on project size.
* The results for your chosen metric are displayed in the table.
* If you want to change your choices of metric/granularity you have to click calculate again.
* Below the table you can export the whole table (also the bigger part which is not shown in the UI). For this, choose a format from the dropdown.

grafik
grafik

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
Whatever I try to select, I get a NPE when I calculate

reckart:
The new sidebar seems to work nicely, I can configure different granularities and statistics, calculate them and the table shows up. Export also seems to work.

The build currently fails because of a used undeclared dependency.

The dropdown fields in the UI look differently than elsewhere in the application. Compare e.g. the dropdowns in the statistics sidebar with the layer dropdown in the right sidebar. It looks like the bootstrap styling is missing. Also, we usually put a label left of the dropdown and align items in a grid.

To save space, the help link could go into the card header (floating and the far end right of the header "statistics").

When I reduce the window width, the text of the "Calculate" button disappears. We have such an effect in other places as well, but then we also have an icon in the button which does not disappear. Also, we usually right-align buttons within their context, but the calculate button is on the left.

The icon in the "Export" button is right of the button text - we usually have it on the left.

Numbers should be right-aligned in the table.

The table seems not to consider the layer visibility configured in the preferences dialog. I have a project where only two layers are visible (POS and Dependency) but I see many other layers in the table as well.

If I increase the width of the left sidebar in the preferences dialog, the width of the table does not change. The width of the export fields below the table adapt to the new width, but not the table itself.

The export stuff could either go into a dropdown or at least stay in a fixed position at the bottom of the sidebar (CSS: card-footer).

When I reduce the height of the browser window, the sidebar content breaks out of the overall layout. The table should be in a div with the classes "scrolling flex-content" to avoid this. Adding the "flex-content" class to a div around the table should also push the export stuff to the bottom.

Issue #2612, opened 2021-09-19
Add possibility to enforce singelton document level annotations


jcklie:
**Is your feature request related to a problem? Please describe.**
Sometimes it is desired to only allow a single instance of a document level annotation per document. Ideally, that would then already been created beforehand.

**Describe the solution you'd like**
Add a flag to document level annotation layer settings to enable this.

PR #2615, opened 2021-09-20
#2612 - Add possibility to enforce singelton document level annotations

reckart:
**What's in the PR**
- Add option to mark a document metadata layer as a singleton
- Route DocumentOpenedEvents to the UI so the DocumentMetadataSidebar can react and initialize the singletons
- Put delete button direclty into the sidebar without wrapping it in a dropdown

**How to test manually**
* Mark a document metadata layer as singleton
* Open annotation page, open metadata sidebar, enter data, switch between documents

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
It works, the only issue I had was that I needed to open a document twice to create the singleton annotation. I think after the second open it was available in all documents thoug.

https://gyazo.com/5d87c97f4d54516db7800b0e4809379d

reckart:

UWinch:
It is no longer possible for me to create other annotations (except the document metadata), the right sidebar is missing. This is also the case if no document metadata layer is configured.

reckart:
The right sidebar should only be missing if there is no "regular" layer and only document-metadata layers. As long as any span/relation/chain layer exists, the right sidebar should be visible.

reckart:
@UWinch the sidebar missing issue should be fixed now.

PR #2640, opened 2021-09-30
Added kubernetes artifacts

MATRIX4284:
**What's in the PR**
* Kubernetes config file to run INCEpTION

**How to test manually**

```
kubectl create -f inception.yaml
```

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Thanks for the PR :) Could you please fill in the description template so we have an idea what the PR actually is?

MATRIX4284:
Added kuberetes artifact for inception so that it can e run as cloud native

MATRIX4284:
> Thanks for the PR :) Could you please fill in the description template so we have an idea what the PR actually is?

I provided the kubernetes artifact .

reckart:
Ok, nice. How can one use it?

MATRIX4284:
Hi ,

It is executed by this command below.

kubectl create -f inception.yaml

I am giving the manual as a part of next pr.

Thanks & Regards,
Ka

reckart:
The PR looks very helpful indeed.

However, I cannot tell if the location where the PR places the file is a good one. 

Is it something that should be used out of the box? Probably not because there seem to be lots of settings that a particular user would want to change for their particular deployment. So a new module (as in the PR) doesn't seem quite right to me.

So it may be more an example file like the `docker-compose.yml`? In that case it would probably go next to the `docker-compose.yml` in the admin documentation: `inception/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide/scripts/` - and then a short piece of text explaining what the file does and how one would use it would be very useful - some `installation_kubernetes.adoc` file in `/inception/inception-doc/src/main/resources/META-INF/asciidoc/admin-guide`.

MATRIX4284:
> I have enabled Kubernetes on my Docker Desktop and tried the command that you said, but it doesn't seem to work:
> 
> ```
> % kubectl create -f inception.yaml
> persistentvolume/inception-pv created
> persistentvolumeclaim/inception-pvc created
> service/inception-svc created
> The Deployment "inception" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "mapr"
> ```

Hi there was a typo in volume name I fixed that ..now it will run without errors

reckart:
Yes, as I said, I am looking into setting up a multi-platform image using docker buildx. Actually, I have meanwhile already managed to build an M1 compatible image. Next I need to look into how the multi-image setup can be pushed as part of the Maven build. Normally, I am using a Maven plugin to do that and that picks up the dockerhub credentials from my machine. When I try the same thing using docker buildx, it currently fails because of missing credentials. Baby steps...

MATRIX4284:
Hi,

I will look into it and get back to you.

Thanks & Regards,
Kaustav

blochs:
You may want to look into an init container and mounting the volume for the config's required to get this to run in Kubernetes. In addition, having an Ingress config in the code might be helpful as well.

reckart:
Trying this again now using Docker Desktop 4.25.1 (128006) and its built-in Kubernetes v1.28.2 - and ... it actually launches! 🚀

reckart:
Superseded by https://github.com/inception-project/inception/pull/4325

Issue #2443, opened 2021-07-03
Search Result Ordering


antonyscerri:
**Is your feature request related to a problem? Please describe.**
When searching the matches sometimes seem to change order. I think it is related to the creation/deletion of annotations and the order they are then added to the index. If the corpus is naturally sorted by doc id, then the results could be too.

**Describe the solution you'd like**
Stable ordering of results for the same query when changes to documents do not effect the matches themselves.

PR #2660, opened 2021-10-18
#2443 Keep Order of Search Results

Falko1:
**What's in the PR**
- The sortLeaves method in MtasDocumentIndexTest
- In essence, this method takes care that the search order of the documents is the same as the document id.
- This leads to a stable ordering of the search results. 
- Example 1: The search results ordering is kept even if lowLevelPaging is active.
- Example 2: The search results are ordered according to document id and the currently displayed one does not come last 

**How to test manually**
* Play around with the search (e.g. (de-) activate lowLevelPaging, make the same search again after annotating some document, ...) and observe that the search results are always ordered according to document id.
* Of course, this is not valid if some other kind of grouping is active. 

**Automatic testing**
* [x] PR includes unit tests

**Documentation**


inception/inception-search-mtas/src/main/java/de/tudarmstadt/ukp/inception/search/index/mtas/MtasDocumentIndex.java
```diff
@@ -817,14 +820,51 @@ else if (annotationDocumentId != -1
...
+                        String rawSourceDocumentId = document.get(FIELD_SOURCE_DOCUMENT_ID);
+                        if (rawSourceDocumentId == null) {
+                            continue;
+                        }
+
+                        allDocIds.add(Long.valueOf(rawSourceDocumentId));
+                    }
+                }
+            }
+            if (allDocIds.size() == 1) {
```
reckart, line 851: If you test for the size to be one here, I assume that the loop before must collect exactly one result. But if that is the case, why collect into a list and not into a simple variable - and why not stop the loop when the first result has been found. Could you explain/code-comment how the algorithm is intended to work?

reckart:
@Falko1 looks like the test fails now.

reckart:
Hm... looks like I broke it :(

Issue #2792, opened 2022-01-03
Use esbuild for pdfanno


jcklie:
**Describe the refactoring action**
Building PDFAnno is slow and we do not use eslint output, so switch to esbuild

**Expected benefit**
Faster builds

PR #2793, opened 2022-01-03
#2792 - Use esbuild for pdfanno

jcklie:
**What's in the PR**
- Use esbuild-loader
- Remove eslint

**How to test manually**
* Clean project, build INCEpTION, feel that its faster and check PDF editor whether it still works

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


jcklie:
This is strangely enoug slower now, so we better do not merge it.

Issue #2790, opened 2022-01-03
Build command for npm packages does not work on windows


jcklie:
**Describe the bug**
Build command for npm packages does not work on windows

`'.' is not recognized as an internal or external command,`

**To Reproduce**
1. Run `mvn  install -DskipTests=true -D checkstyle.skip=true -D dependency-check.skip=true -X` on Windows
2. See the error

**Expected behavior**
It builds

**Screenshots**

**Please complete the following information:**
 - Version and build ID: `release/22.x`
 - OS: Windows

PR #2794, opened 2022-01-03
Revert "#2790 - Build command for npm packages does not work on windows"

jcklie:
Reverts inception-project/inception#2791


reckart:
I have committed changes to run simply with the esbuild/sass commands and without `npm bin`.

Issue #3194, opened 2022-07-17
Jumping to location may not bring target into view


reckart:
**Describe the bug**
When working with large pages (i.e. the brat editor is larger than the editor area and we have a scrollbar), an active learning target or search hit that I select may be highlighted outside of the currrently visible part of the scrolling window.

**To Reproduce**
1. Set a large page size like 100 rows
2. Search for words that occur outside of the currently visibly part of the document
3. Click on a match
4. Nothing happens

**Expected behavior**
It would be nice if the view would scroll down to the match.

**Please complete the following information:**
 - Version and build ID: 23.8

PR #3195, opened 2022-07-17
#3194 - Jumping to location may not bring target into view

reckart:
**What's in the PR**
- Track the edit highlight or the match focus and try to keep them in view after rendering

**How to test manually**
* Use the brat editor in different cases for annotating and curating and see if the scrolling interferes with normal work or if it helps

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
Is it is, when a match highlight from the search bar is active, when the user performs some action like annotating a span in a different part of the document, the scroll always jumps back to the search highlight - not good. Jumping and highlight need to be decoupled.

PR #3322, opened 2022-08-29
Fix German word 'Possessivpronomen'

stweil:
Signed-off-by: Stefan Weil <sw@weilnetz.de>

**Documentation**


stweil:
@reckart, maybe this is too trivial for an issue, so I decided against creating one.

inception/inception-schema/src/main/java/de/tudarmstadt/ukp/inception/schema/exporters/LegacyProjectInitializer.java
```diff
@@ -124,8 +124,8 @@ public void initialize(Project aProject, String[] aPostags, String[] aPosTagDesc
                         "attribuierendes Indefinitpronomen mit Determiner \nBsp: [ein] wenig [Wasser], [die] beiden [Brüder] ",
                         "substituierendes Indefinitpronomen \nBsp: keiner, viele, man, niemand ",
                         "irreflexives Personalpronomen \nBsp: ich, er, ihm, mich, dir",
-                        "attribuierendes Possessivpronome \nBsp: mein [Buch], deine [Mutter] ",
-                        "substituierendes Possessivpronome \nBsp: meins, deiner",
```
stweil, line 128: It would also be possible (and in my opinion desirable) to remove the space before the newline, but that pattern seems to be used widely here, so I did not change that.

reckart: This is legacy code - I think it is only used when importing **oooooold** projects (prior to WebAnno 2.0) - probably not worth polishing.

Issue #3326, opened 2022-08-30
HTML errors and warnings


stweil:
**Describe the bug**
The INCEpTION web pages include HTML 5 code with errors and warnings:

```
Error: Stray end tag meta.
Error: An img element must have an alt attribute, except under certain conditions. For details, consult [guidance on providing text alternatives for images](https://www.w3.org/WAI/tutorials/images/).
Warning: Consider adding a lang attribute to the html start tag to declare the language of this document.
Warning: The type attribute is unnecessary for JavaScript resources.
```

Wrong HTML code typically works differently on different web browser, so can be the reason why certain web browsers don't work..

**To Reproduce**
I installed the latest Docker image (`INCEpTION -- 24.1 (2022-08-23 13:26:20, build 631d180b)`).
Use the [W3C validator](https://validator.w3.org/) to validate selected pages either by URI (normally only possible for login page) or by [direct input](https://validator.w3.org/#validate_by_input) (possible for any page).

**Expected behavior**
Ideally all web pages should pass without any HTML error or warning.

stweil:
More reports for the first page after successful login:

```
Error: Element [span](https://html.spec.whatwg.org/multipage/#the-span-element) is missing one or more of the following attributes: role.
Error: Element [div](https://html.spec.whatwg.org/multipage/#the-div-element) not allowed as child of element [span](https://html.spec.whatwg.org/multipage/#the-span-element) in this context. (Suppressing further errors from this subtree.)
Error: No p element in scope but a p end tag seen.
Error: Element [p](https://html.spec.whatwg.org/multipage/#the-p-element) not allowed as child of element [small](https://html.spec.whatwg.org/multipage/#the-small-element) in this context. (Suppressing further errors from this subtree.)
```

stweil:
There remain unsolved HTML errors even after the latest commits. Therefore I suggest to keep this issue open.

reckart:
I'll keep it open for further PRs on v25 but will close it when v25 gets released. A new issue can be opened for issues that remain after v25.

PR #3327, opened 2022-08-30
#3326 - Fix some warnings and errors in HTML code

stweil:
-  #3326 - Remove unnecessary type attribute for JavaScript resources in HTML code
-  #3326 - Add alt attribute to img element in HTML code
-  #3326 - Remove wrong </meta> in HTML code

### DIFF
--- a/inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
+++ b/inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
@@ -30,7 +30,7 @@
 </head>
 <body data-bs-no-jquery="data-bs-no-jquery">
   <div>
-    <img id="spinner" src="images/spinner.gif" />
+    <img id="spinner" alt="" src="images/spinner.gif" />
   </div>
   
   <header class="page-header">
--- a/inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
+++ b/inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
@@ -18,10 +18,10 @@
 -->
 <html xmlns:wicket="http://wicket.apache.org">
 <head>
-  <meta charset="utf-8"></meta>
-  <meta http-equiv="X-UA-Compatible" content="IE=edge"></meta>
-  <meta name="viewport" content="width=device-width, initial-scale=1"></meta>
-  <meta name="google" content="notranslate"></meta>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="google" content="notranslate">
   
   <link rel="icon" href="favicon.png" type="image/x-icon"/>  
   <title><wicket:message key="product.name" /> - <wicket:message key="page.title" /></title>
--- a/inception/inception-doc/src/main/resources/META-INF/asciidoc/docinfo.html
+++ b/inception/inception-doc/src/main/resources/META-INF/asciidoc/docinfo.html
@@ -30,7 +30,7 @@
   }
 </style>
 
-<script type="text/javascript">
+<script>
     $(function () {
         // Add a new container for the tocify toc into the existing toc so we can re-use its
         // styling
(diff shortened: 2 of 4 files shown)


inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
```diff
@@ -18,10 +18,10 @@
 -->
 <html xmlns:wicket="http://wicket.apache.org">
 <head>
-  <meta charset="utf-8"></meta>
```
reckart, line 21: Have you tried building and running INCEpTION with this change? This is not valid XML - I think the Apache Wicket framework we use to render pages may not like it.

inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
```diff
@@ -30,7 +30,7 @@
 </head>
 <body data-bs-no-jquery="data-bs-no-jquery">
   <div>
-    <img id="spinner" src="images/spinner.gif" />
+    <img id="spinner" alt="" src="images/spinner.gif" />
```
reckart, line 33: Isn't adding an empty `alt` attribute like cheating?

stweil: From https://www.w3.org/WAI/tutorials/images/: "[Decorative images](https://www.w3.org/WAI/tutorials/images/decorative/): Provide a null text alternative (alt="") when the only purpose of an image is to add visual decoration to the page, rather than to convey information that is important to understanding the page."

If the image is not a decorative image, a reasonable text would be better.

reckart: The image indicates if an AJAX request to the server is in progress.

reckart:
It appears the PR may not be configured to allow committers to the target project to make additions to it? I assume that is why I do not see the usual button to update the PR when it is out-of-date with the target branch.

inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
stweil: It is valid HTML 5, so the framework should handle it. See https://cwiki.apache.org/confluence/display/WICKET/Wicket%27s+XHTML+tags#Wicket'sXHTMLtags-Elementwicket:header-items for examples.

stweil: But I did not try building and running INCEpTION up to now (only installed Docker image).

reckart: I'll try. From my memory, I tried this ages ago and it didn't work - maybe I mis-remember or maybe it was fixed since then. We'll see.

reckart: Look good - no crashing and burning ;)

stweil:
I did not disable modifications from your side, but rebased the commits now to synchronize it with your main branch.

reckart:

reckart:
If you want to indicate that you plan to add more commits to a PR, you can turn the PR into a "draft PR". Otherwise, I assume that a PR that looks good can simply be merged.

inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
stweil: Then the question is whether some text would be helpful for a blind person. Personally I think no text would be acceptable. Overwise "busy" or a similar text could be used, but that needs localisation of course.

stweil:
I don't plan to add more commits, because I have no quick solutions for the remaining HTML errors.

inception/inception-ui-core/src/main/java/de/tudarmstadt/ukp/clarin/webanno/ui/core/page/ApplicationPageBase.html
reckart: I was wondering about localization as well - that would also affect the `lang` tag on the page. But I think fixing such issues should at the moment be considered without localization since we don't really yet support any language but English. When/if localization is addressed, this would need to be revisited.

reckart (approved):

Issue #3611, opened 2022-12-04
Upgrade to Spring 6, Boot 3, Wicket 10, etc.


reckart:
**Describe the refactoring action**
Update to the current generation of Spring and Spring Boot which use the new Jakarta namespaces for J2EE APIs which entails a host of additional upgrades.

- Spring 5 -> 6
- Spring Boot 2 -> 3
- Wicket 9 -> 10
- Hibernate 5 -> 6

**Expected benefit**
Avoid technical debt coding against old APIs.

**Upstream issues**
* https://issues.apache.org/jira/browse/WICKET-6882
* https://github.com/l0rdn1kk0n/wicket-bootstrap/issues/980
* https://github.com/sebfz1/wicket-jquery-ui/issues/353

PR #3612, opened 2022-12-04
#3611 - Upgrade to Spring 6, Boot 3, Wicket 10, etc.

reckart:
**What's in the PR**
- Java 11 -> 17
- spring 5.3.24 -> 6.1.8
- spring-boot 2.7.5 -> 3.3.0
- spring-data 2.7.5 -> 3.3.0
- spring-security 5.7.5 -> 6.3.0
- tomcat 9.0.69 -> 10.1.24
- servlet-api 4.0.1 -> 6.0.0
- hibernate 5.6.14 -> 6.5.2
- hibernate-validator 6.2.5 -> 8.0.0
- wicket 9.x -> 10.1.0
- wicketstuff 9.x -> 10.1.0
- wicket jquery 9.x -> 10.1.0
- wicket-jquery-selectors 4.0.4 -> 4.0.5
- wicket bootstrap 6.0.x ->  7.0.5
- wicket springboot 3.x -> 4.0.0
- wicket-webjars 4.0.3 -> 4.0.4
- Use Jakarta dependencies instead of old javax dependencies

**Problems**
- [x] All DB entities with a `@Lob` String need to be set to `@Column(length = 16_777_216)` to make current MySQL scheme validation pass because Hibernate now uses the length to determine the best column type (before `@Lob` was always `LONGTEXT`) - that is maybe not the best idea. Might need to migrate the DB to use other types. Need to check embedded DB as well.

**How to test manually**
* No specific test procedure. Pervasive change, just try using the app.

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**


reckart:
INCEpTION does not use Jetty at all, it uses Tomcat. If Jetty is used, then only by third-party libraries for unit tests.

Issue #4026, opened 2023-05-26
Support for error tracking with Sentry


torgeirl:
**Is your feature request related to a problem? Please describe.**
As a system administrator I would like to be able to use Sentry as an error tracker the Inception deployment I maintain.

**Describe the solution you'd like**
This would be possible if the needed [Sentry SDK libraries (MIT License)](https://github.com/getsentry/sentry-java) were included in the standard build of Inception. Since Inception is built using Spring Boot, this would include two new dependencies:

* `sentry-spring-boot-starter` (required)
* `sentry-log4j2` (optional; would allow for transaction performance monitoring)

The later could in addition be supported by also adding [`SentryAppender`](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/#log4j2) to [the list of supported loggers](https://inception-project.github.io/releases/27.8/docs/admin-guide.html#_logging_in_json_format). Otherwise, it would have to be specified as a custom logger.

[Sentry's documentation for Spring Boot](https://docs.sentry.io/platforms/java/guides/spring-boot/)

**Describe alternatives you've considered**
Kibana/ELK is fine for metrics/general traffic monitoring, but compared to Sentry they require more setup to monitor errors, include more noise, and often requires you to already be digging for answers to a reported issue.

**Additional context**
I would happily expand Inception's documentation with instructions on how to set it up if the needed libraries get included.

reckart:
@torgeirl how are you sentry experiments going? Is this something we should keep? If you don't get sentry to run, I would consider removing the dependencies again.

torgeirl:
@reckart: I have been traveling since right before you added the latest dependency, and have therefore not had the change to test it yet. Back at the office this Thursday. Will give you an update either Thursday or Friday!

torgeirl:
@reckart: seems to be working well now, but I like to test some more to be sure. Is there a rough timeline for the 29.0 release ("docs due date")?

reckart:
@torgeirl would you like to write up a few lines of documentation regarding sentry this week? The 29.0 release is scheduled for Tuesday next week.

torgeirl:
> @torgeirl would you like to write up a few lines of documentation regarding sentry this week? The 29.0 release is scheduled for Tuesday next week.

Will do! :+1:

reckart:
Thx. I have squashed-merged it and in the process updated the commit comment according to the usual format. Normally, I prefer regular merges, but then the commit messages should already follow the conventions.

PR #4042, opened 2023-05-30
#4026 - Support for error tracking with Sentry

reckart:
**What's in the PR**
- Add sentry dependencies

**How to test manually**
* Try configuring a connection to Sentry via properties set in the `settings.properties` file

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-app-webapp/pom.xml
+++ b/inception/inception-app-webapp/pom.xml
@@ -522,7 +522,13 @@
       <artifactId>commons-logging</artifactId>
       <scope>provided</scope>
     </dependency>
-
+    
+    <!-- SENTRY DEPENDENCIES -->
+    <dependency>
+      <groupId>io.sentry</groupId>
+      <artifactId>sentry-spring-boot-starter</artifactId>
+    </dependency>
+      
     <!-- DATABASE / HIBERNATE -->
 
     <dependency>
@@ -946,6 +952,8 @@
               <usedDependency>org.apache.logging.log4j:log4j-core</usedDependency>
               <usedDependency>org.apache.logging.log4j:log4j-layout-template-json</usedDependency>
               <usedDependency>commons-logging:commons-logging</usedDependency>
+              <!-- Monitoring - used via reflection / optional -->
+              <usedDependency>io.sentry:sentry-spring-boot-starter</usedDependency>
               <!-- Themes -->
               <usedDependency>com.googlecode.wicket-jquery-ui:wicket-kendo-ui-theme-bootstrap</usedDependency>
               <usedDependency>com.googlecode.wicket-jquery-ui:wicket-jquery-ui-theme-uilightness</usedDependency>
--- a/inception/pom.xml
+++ b/inception/pom.xml
@@ -1532,6 +1532,14 @@
       </dependency>
 
       <dependency>
+        <groupId>io.sentry</groupId>
+        <artifactId>sentry-bom</artifactId>
+        <version>6.20.0</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+
+      <dependency>
         <groupId>javax.servlet</groupId>
         <artifactId>javax.servlet-api</artifactId>
         <version>${servlet-api.version}</version>


reckart:
@torgeirl I have added the Sentry dependencies here. Can you check out, build locally (`mvn clean install -DskipTests` - the JAR will be in `inception/inception-app-webapp/target`) and check if you can connect INCEpTION to your Sentry now?

torgeirl:
> Can you check out, build locally (`mvn clean install -DskipTests` - the JAR will be in `inception/inception-app-webapp/target`) and check if you can connect INCEpTION to your Sentry now?

@reckart: unfortunately I'm having trouble making it work as described by their documentation.

I have tested both with Java System Property (`-Dsentry.dsn=https://examplePublicKey@my.sentry-server.com/0`) and by providing the SDK's configuration through modifying `inception/inception-app-webapp/src/main/resources/application.yml`:

```
# ===================================================================
# INCEpTION Properties
# ===================================================================

(...)

sentry:
  dsn: ${monitoring.sentry.dsn:${SENTRY_DSN:}} # left empty this will disable Sentry
  exception-resolver-order: ${monitoring.sentry.exception-resolver-order:2147483647}
  environment: ${monitoring.sentry.environment:production}
  release: ${monitoring.sentry.release:${project.version}} # handled automatically if Spring Boot Git is configured
  traces-sample-rate: ${monitoring.sentry.traces-sample-rate:1.0}
```

My idea for the later was to use `settings.properties` values similar to how the database configuration is done, ie:
```
(...)

monitoring.sentry.dsn=https://examplePublicKey@my.sentry-server.com/0
monitoring.sentry.exception-resolver-order=-2147483647 # overwrite default from only unhandled to all errors
monitoring.sentry.environment=prod # overwrite the default from `production` to `prod`
monitoring.sentry.traces-sample-rate=0.2 # overwrite the default from 100% to 20% of transactions
```

----

For some Sentry SDKs such as Python, the SDK has to be explicitly initiated. That doesn't seem to be case for the Java/Spring/Spring Boot SDKs, but one section of their documentation suggests a SDK option [must be enabled to enable loading configurations from properties file, system properties or environment variables](https://docs.sentry.io/platforms/java/guides/spring-boot/configuration/#configuration-methods):

> **Configuration methods**
> There are multiple ways to configure the Java SDK, but all of them take the same options. See below for how to use each configuration method and how the option names might differ between them.
> 
> To enable loading configuration from the properties file, system properties or environment variables, `enableExternalConfiguration` has to be set to `true` on `SentryOptions`:
> 
> ```Java
> import io.sentry.Sentry;
> 
> Sentry.init(options -> {
>   options.setEnableExternalConfiguration(true);
> });
> ```

If `sentry.dsn` isn't set that would indeed cause the SDK to not initiate, making me believe this could be the issue. But I'm unsure where and how this should be set. Could you give me some pointers for how I could test this?

reckart:
Just write your properties directly to `settings.properties` without modifying the `application.yml` file and without the `monitoring` prefix. If your configuration examples use YAML, you should just need to "flatten" the key name into the dot notation for it to work in the `settings.properties` file.

```
sentry.dsn=https://examplePublicKey@my.sentry-server.com/0
```

torgeirl:
As far as I can tell, that didn't change anything. But I'll look more into it next week. 👍

reckart:
So looking at the Sentry Spring Boot starter code, as soon as you set `sentry.dsn` to some value in `settings.properties` (or using `-Dsentry.dsn=...` on the command line), their code should spring into action.

reckart:
If I start with `-Dsentry.dsn=someurl` and `-Ddebug`, then I see in the startup report that various Sentry components get activated. No idea what they should do though. I had a short look at whether I can easily run a local sentry instance - I probably can, but their `install.sh` script seems rather complicated and I do not want to have any system-level stuff installed or upgraded by that script... so I preferred not trying it.

torgeirl:
The `sentry-spring-boot-starter` itself seems to be working and once I got the settings correct I'm able to pass on transaction data («bonus» objective) to Sentry. Unfortunately, I still haven't figured out how to get the error tracking (main objective) to work. I think the issue has to do with how Inception handles error, or rather how to replace/extend that handling so it also work with Sentry.

The following is mostly notes to myself for when I continue to work on this in some weeks, but I share it here in case you have some input or follow-up questions.

As far as I can tell, there are three ways to pass on errors to Sentry from a Spring Boot application:
1. the `sentry-logback` integration; it requires `ch.qos.logback.classic.Logger` and since that is excluded from Inception's dependencies I guess it conflicts with another log framework and/or has some security issue. :x: 
2. the `sentry-log4j2` integration; it depends on `spring-boot-starter-log4j2` and requires `spring-boot-starter-logging` to be excluded from `spring-boot-starter` and doesn't seem to be possible. :x: 
3. either using a custom `log4j2.xml` file (preferred) or extending Inception to include a Sentry logger with the available log4j2 resources so that its available when the Sentry SDK gets activated. :grey_question: 

Sentry's documentation has [a suggestion for how to configure a `log4j2.xml` file](https://docs.sentry.io/platforms/java/guides/spring-boot/logging-frameworks/) when using the `sentry-log4j2` integration, but that involves some extra packages we can't use:
```
<Configuration status="warn" packages="org.apache.logging.log4j.core,io.sentry.log4j2">
  <!-- ... -->
```

reckart:
There is a new PR for adding the sentry log4j2 dependency: https://github.com/inception-project/inception/pull/4059

torgeirl:
> There is a new PR for adding the sentry log4j2 dependency: #4059

Thanks, I'll test it when I'm back in the office! 🙏

Issue #4068, opened 2023-06-13
Ability to change interface language (i18n)


bhusson:
The application interface is written in english and we would like to let non english speaker use it.

We would like to be able to provide an alternative language pack (for French), and ease the addition of other languages as well.

Ideally we should be able to switch language from the interface.

PR #4069, opened 2023-06-13
#4068 - Ability to change interface language (i18n)

awachatilf:
**What's in the PR**
* Draft of translation support

**How to test manually**
* Configure the application settings.properties file on french language (locale=fr) and verify that texts are in french

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

**Todo**
* [x] Move the strings extracted from the telemetry configuration page back into the HTML files and translate the entire HTML files - to avoid having to disable HTML escaping on the resource models


inception/inception-curation/src/main/java/de/tudarmstadt/ukp/inception/curation/settings/CurationProjectSettingsMenuItem.java
```diff
@@ -50,7 +53,13 @@ public IconType getIcon()
     @Override
     public String getLabel()
     {
-        return "Curation";
+        IRequestablePage currentPage = PageRequestHandlerTracker.getLastHandler(RequestCycle.get())
```
reckart, line 56: We have a new static method `de.tudarmstadt.ukp.clarin.webanno.support.wicket.resource.Strings.getString(String)` that should be used in backend code like this. However, it requires that the string to be obtained is contained in a `wicket-package.properties` file in the package or parent-package of the code that calls the method.

I am using this new method in very few places at the moment, e.g. in `de.tudarmstadt.ukp.inception.annotatorjs.AnnotatorJsHtmlAnnotationEditorFactory.getDisplayName()`.

inception/inception-telemetry/src/main/java/de/tudarmstadt/ukp/clarin/webanno/telemetry/matomo/MatomoTelemetryTraitsEditor_fr.html
```diff
@@ -0,0 +1,66 @@
+<!--
```
reckart, line 1: I wonder if it would be better to pull out the strings here and put them into a `.properties` file instead of having multiple versions of the `.html` files around.

inception/inception-telemetry/src/main/java/de/tudarmstadt/ukp/clarin/webanno/telemetry/TelemetryStrings.java
awachatilf, line 1: I've tried transferring the translation of the HTML files to the properties files in order to centralise the translations in wicket.properties file as it was suggested before. I've encountered the problem of variable interpolation within the text (such as "product.name" or "save"). I wanted to know if my approach was the right one by creating this class (TelemetryStrings) and naming it this way in order to call the variables in the properties file using the notation ${variable}. Could someone tell me if I did it in the proper way?

reckart: Thanks for the suggestion. It looks a bit strange. The question is why the properties are not interpolated - IMHO that should just work.

reckart: Another consideration - if we need to set `setEscapeModelStrings(false)` in order to pull the strings out of the HTML files, then I'd say it may be better to internationalize the entire files as you had before.

reckart: Ok, so the correct syntax for properties inside properties is `#(propertyname)`. I have removed the TelemetryStrings and fixed the properties files - now it works as expected.

inception/inception-telemetry/src/main/java/de/tudarmstadt/ukp/clarin/webanno/telemetry/matomo/MatomoTelemetryTraitsEditor_fr.html
reckart: Thanks for trying this out. Turns out it was a bad idea because the extracted strings contain HTML markup which would force use disabling HTML escaping on the resource models - which we do not want to do. So better to translate the entire HTML files as you had. I have added a corresponding todo to the initial description above.

reckart:
@awachatilf Looks like this PR is coming along pretty well. I wonder what (if anything) would be missing from your side before we could merge it?

The commit messages in the PR do not match the project convention. We have two options here:

1) you rewrite the commit comments using an interactive git rebase. Mind that depending on what tool you are using to do this, you may have to change the comment character option in git temporarily to something other than `#` because our initial comment line always starts with `#NUMBER - TITLE`...
2) we squash-merge the PR. All commits will become a single commit in this process and I rewrite the commit comment at merge time

I saw that some of the newly introduced properties match what seems should be a "core term", e.g. 👍 

```
curationprojectsettings.menuitem.label=Curation
curationprojectsettings.panel.factory.label=Curation
```

Instead of introducing new properties here, I would have tried introducing a "core term" `curation=Curation` and put that into `WicketApplicationBase.properties` - or at least not introduce multiple properties for essentially the same term. But it is a rather philosophical comment. Having strictly separate properties for every single string is also an approach and has its own benefits. I do not have a particularly strong opinion about either approach at this point.

The English properties files are in UTF-8. The French ones are in ISO-8859-1. I believe it would be good to use UTF-8 for both of them. For this to work properly, we should rename them from `.properties` to `.utf8.properties` though. But let's maybe do that in a separate PR. This one is already getting very very large - and too large PRs are not good.

inception/inception-telemetry/src/main/java/de/tudarmstadt/ukp/clarin/webanno/telemetry/ui/TelemetrySettingsPage.html
```diff
@@ -25,13 +25,10 @@
...
             <div class="flex-content">
               <p>
-                In order to support the development of <wicket:message key="product.name"/>, it would
-                like to collect some data and send it home. Please review the settings below and 
-                opt in or out of the respective data collections. Once you are done, use the
-                <b><wicket:message key="save"/></b> button on the right side to save your choices. If
-                there were any changes to the data collection process, this page will appear again
-                when an administrator logs in. You can always revise your choices by accessing this
-                page via the <i class="fa fa-heartbeat"></i> icon in the page footer.
+                <wicket:message key="settingsIntro">
```
awachatilf, line 28: I think I've found a way to use a translation properties file with HTML content without escaping HTML. I created wicket elements for texts inside tags (like <b>) and translated them separately. Do you think this is the right way to do it? I'd prefer all the translations to remain centralised in the properties files because this will be less complicated to maintain translations later on.

reckart: Actually, that is not necessary. When embedding strings in a page using `wicket:message`, then the strings in the properties files can contain markup. Only if the strings are included through a `ResourceModel`, we would need to turn off escaping to support markup. In fact, we should try to work with `wicket:message` as much as possible and use `ResourceModel` only for cases where the text in a component needs to change dynamically (i.e. where the message key is not known or not fixed).

reckart: I made a commit to change the Telemetry page accordingly.

reckart:
> There are a lot of hardcoded strings in java files and I'm not sure which one should be translated or not. For example, I suppose logger message shouldn't, but I suppose corresponding functions of Wicket Component class (success(), info(), warn(), error() and fatal()) that seem to display popup messages should. For some other functions like getName(), getDescription() or getDisplayname() of some interfaces I'm not really sure if they should be translated. I'm not sure either if Validate.notNull() strings are displayed in the application and should be translated. There are many more, but these seems to be most common cases.

We should for the moment focus on the properties files that are already there and not go through all the Java files (and possibly typescript files) to fix everything. Translating only the properties files we already have will already be a big step. After that, we might go through the code module by module and see if/which strings in Java files should be internationalized. But that should be a separate issue and then best separate PRs for different modules.

In any case, we won't be able to translate *everything*  because many strings are not even under our control (they come from third-party libraries or from the JVM itself).

reckart:
> I should rename properties files key with shorter names as you told me to. (I don't know if you saw Benjamin Husson's proposition of prefixing package keys with local and application ones with global (or something like))

Also not critical. This can happen later. I still would like to make use of some tooling that generates Java files with constants for the keys so we can a compiler-checkable connection to the strings. Latest at that point, we'll notice which strings have awkward (longish) names and can fix them.

reckart:
> Are you sure english properties file are currently in UTF-8? The "ä" in "Universität" first line comments doesn't seem to display well in UTF-8. In any case, if it's possible to encode in UTF-8 I also think it's better. But if you prefer me to do that in another PR there's no problem.

Yes, let's do that separately. Separate issue as well.

reckart:
Closing due to inactivity. I'll try looking into an alternative solution.

Issue #3923, opened 2023-04-05
Error accessing PDF


martin-lentschat:
**Describe the bug**
After uploading a PDF file, I am not able to see or annotate it.
It seems that the server is not able to access the file, the Console is giving me the following error messages: 
-  `Refused to connect to 'http://my.server.name.fr/p/test2/annotate?3-1.1-centerArea-editor-vis&time=1680690334665' because it violates the following Content Security Policy directive: "connect-src 'self'".`
- `Refused to connect to 'http://my.server.name.fr/p/test2/annotate?3-1.1-centerArea-editor-vis&time=1680690334665' because it violates the document's Content Security Policy.`

I eddited the settings.properties file with different `wicket.core.csrf.accepted-origins` but it did not changed anything.

**Observations**
- I tried changing editors, the brats are working (but are not adapted to annotate pdfs) and PDF(legacy) indicates an incompatible file format (pdf2). I believe those behaviour are expected (same thing happens with different pdf files).
- The console also indicates that the connexion to the WebSocket failed. This message appears when entering a project and I do not know if it is important since it does not seems that it is affecting anything.
- When clicking on the link in the console error message, `http://my.server.name...`, the pdf opens in the brower, but is redirected to http***s***. Could this be related ?

**Information:**
 - Version and build ID: 27.3 (2023-03-28 21:05:02, build cdd73a94)
 - OS: Windows
 - Browser: Chrome

martin-lentschat:
So, I found the source of the issue: the `/etc/apache2/conf-available/inception.local.conf` file had not been created.

However a new message appeared in the console:
`The Cross-Origin-Opener-Policy header has been ignored, because the URL's origin was untrustworthy. It was defined either in the final response or a redirect. Please deliver the response using the HTTPS protocol. You can also use the 'localhost' origin instead. See https://www.w3.org/TR/powerful-features/#potentially-trustworthy-origin and https://html.spec.whatwg.org/#the-cross-origin-opener-policy-header.`

I do not know if it is a problem or not, since everything seems to work fine.

reckart:
Great you found theissue :)

Certain browser security features will not be available unless you use https. The browser lets you know this.

martin-lentschat:
Strange thing is that https is not activated anymore. It was before.

I will keep this issue open for a few days if that's ok, and update you with my investigations ^^

reckart:
If you upgraded - depending on what your old version was - it might be that relevant Spring properties have changed their names.

[...]

PR #4374, opened 2023-12-11
Docs/admin apache proxy config

tehaef:
**What's in the PR**
* Added information about forwarding the URL scheme when using Apache HTTPD with mod_proxy and SSL. 
* The existing documentation shows a deprecated approach to setting permissions in the Apache config. Apache 2.4 can be configured to use still support the older approach, and often it is (by use of mod_access_compat). If not, the example might not work at all. Therefore the newer method was added beneath.

**How to test manually**
* -

**Automatic testing**
* [ ] PR includes unit tests

**Documentation**

Please also refer to: https://github.com/inception-project/inception/issues/3923#issuecomment-1840413567


reckart:
Thanks!

Issue #4377, opened 2023-12-12
Configure include and exclude patterns for logged events


reckart:
**Is your feature request related to a problem? Please describe.**
The event log currently captures application events based on an internal and a configurable exclude list. However, there could be cases where e.g. only specific groups of events should be excluded or included.

**Describe the solution you'd like**
* Deprecate the `event-logging.exclude-events` setting.
* Introduce new settings `event-logging.events.include-patterns` and `event-logging.events.exclude-patterns`. Include should be `.*` by default and exclude should match the current `event-logging.exclude-events` defaults.
* If an event matches any include pattern, and does not match any of the exclude patterns, it should be logged.

PR #4490, opened 2024-02-01
#4377 - Configure include and exclude patterns for logged events

serwarde:
**What's in the PR**
* Deprecated exclude events.
* Introduced include and exclude patterns.
* Including everything (.*) by default, and excluding a predefined set of events by default.
* Refactored EventLoggingListener to use the new include and exclude patterns.
* Added tests.

**How to test manually**
* N/A

**Automatic testing**
* [x] PR includes unit tests

**Documentation**

### DIFF
--- a/inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventLoggingListener.java
+++ b/inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventLoggingListener.java
@@ -28,6 +28,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.DisposableBean;
@@ -82,15 +83,17 @@ public class EventLoggingListener
             return eventCache.get(aEventName);
         }
 
-        boolean shouldLog = false;
+        boolean shouldLog;
 
-        if (properties.getIncludePatterns() != null && !properties.getIncludePatterns().isEmpty()) {
+        if (CollectionUtils.isEmpty(properties.getIncludePatterns())) {
+            shouldLog = true;
+        }
+        else {
             shouldLog = properties.getIncludePatterns().stream()
                     .anyMatch(pattern -> Pattern.matches(pattern, aEventName));
         }
 
-        if (shouldLog && properties.getExcludePatterns() != null
-                && !properties.getExcludePatterns().isEmpty()) {
+        if (shouldLog && !CollectionUtils.isEmpty(properties.getExcludePatterns())) {
             shouldLog = properties.getExcludePatterns().stream()
                     .noneMatch(pattern -> Pattern.matches(pattern, aEventName));
         }
--- a/inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventLoggingListener.java
+++ b/inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventLoggingListener.java
@@ -83,10 +83,11 @@ public class EventLoggingListener
         if (!maybeAdapter.isPresent()) {
             return;
         }
+        
 
         var adapter = maybeAdapter.get();
 
-        if (properties.getExcludeEvents().contains(adapter.getEvent(aEvent))) {
+        if (!properties.shouldLogEvent(adapter.getEvent(aEvent))) {
             return;
         }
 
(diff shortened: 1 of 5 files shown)


inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/EventLoggingListener.java
```diff
@@ -86,7 +86,7 @@ public void onApplicationEvent(ApplicationEvent aEvent)
 
         var adapter = maybeAdapter.get();
 
-        if (properties.getExcludeEvents().contains(adapter.getEvent(aEvent))) {
+        if (!properties.shouldLogEvent(adapter.getEvent(aEvent))) {
```
reckart, line 89: IMHO `EventLoggingProperties` should remain a data class. That is I would tend to move "shouldLogEvent" into the logging service instead - maybe as a static / package-private method so it can be tested without having to bring up a Spring Boot context.

inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImpl.java
```diff
@@ -32,14 +33,12 @@ public class EventLoggingPropertiesImpl
...
 
-    private Set<String> excludeEvents = Set.of( //
-            // Do not log this by default - hardly any information value
-            AfterCasWrittenEvent.class.getSimpleName(), //
-            AvailabilityChangeEvent.class.getSimpleName(), //
-            "RecommenderTaskNotificationEvent", //
-            BeforeDocumentOpenedEvent.class.getSimpleName(), //
-            PreparingToOpenDocumentEvent.class.getSimpleName(), //
-            "BrokerAvailabilityEvent", //
+    private Set<String> includePatterns = Set.of(".*"); // Default include everything
```
reckart, line 36: I think and empty set should imply `.*`- so it should probably not be necessary to have to specify `.*` explicitly.

inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImpl.java
```diff
@@ -55,14 +54,34 @@ public void setEnabled(boolean aEnabled)
...
 
     @Override
-    public Set<String> getExcludeEvents()
+    public Set<String> getIncludePatterns()
+    {
+        return includePatterns;
+    }
+
+    @Override
+    public void setIncludePatterns(Set<String> includePatterns)
```
reckart, line 63: Please use `aParameterName` (prefix with `a`) and avoid using `this`.

inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImpl.java
```diff
@@ -55,14 +54,34 @@ public void setEnabled(boolean aEnabled)
...
-        return excludeEvents;
+        this.excludePatterns = excludePatterns;
     }
 
     @Override
-    public void setExcludeEvents(Set<String> aExcludeEvents)
+    public boolean shouldLogEvent(String eventName)
     {
-        excludeEvents = aExcludeEvents;
+        return includePatterns.stream().anyMatch(pattern -> Pattern.matches(pattern, eventName))
```
reckart, line 83: This code compiles the patterns and then matches them against the names every time an event is processed. This is quite slow. Assuming the set of names of events is finite, I would suggest to set up a (static) cache where once you discover if a given event name should be logged or not, you enter that event name with a true/false value. Thus, you only need to check the patterns when an event name is not yet in the cache (`O(n)`). If it is in the cache, then the lookup is `O(1)`.

inception/inception-log/src/test/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertThat(properties.shouldLogEvent(eventPreparingToOpenDocument)).isFalse();
+        assertThat(properties.shouldLogEvent(eventBrokerAvailability)).isFalse();
+        assertThat(properties.shouldLogEvent(eventShutdownDialogAvailable)).isFalse();
+    }
+
+    @Test
+    public void shouldLogEvent_EventNotInExcludeLists_ReturnsTrue()
+    {
+        // Test events
+        String eventAfterDocumentOpened = "AfterDocumentOpenedEvent";
```
reckart, line 69: Maybe add an assert that checks if the event is actually not in the exclude list?

reckart (changes requested):

inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImpl.java
serwarde: Set.of() can't be empty

reckart (approved):
Thanks :)

inception/inception-log/src/main/java/de/tudarmstadt/ukp/inception/log/config/EventLoggingPropertiesImpl.java
reckart: There should be a `Collections.emptySet()` or simply `new HashSet<>()`.

PR #4959, opened 2024-07-24
INCEpTION HelmChart

kzgrzendek:
**What's in the PR**
* This PR contains a HelmChart allowing INCEpTION's deployment in a cloud-native & vendor-agnostic way

**How to test manually**
*  Deploy a local K8S cluster using [kind](https://kind.sigs.k8s.io/docs/user/quick-start/) with a deployed [Ingress Controller](https://kind.sigs.k8s.io/docs/user/ingress/) 
* You must have [Helm 3](https://helm.sh/docs/intro/quickstart/) installed
* From the root of the git repo, just run
  ```sh
   helm upgrade --install inception ./chart/inception
   ```
* Once the INCEpTION pod will be ready, you can access the application from http://localhost/

**Automatic testing**
* There's no auto-test for Helm, but we can talk about including documentation generation, schema validation and helm linting in the CI

**Documentation**
* All the Chart's documentation is inside `chart/inception/README.md` file
* That documentation can be re-generated with [helm-doc](https://github.com/norwoodj/helm-docs), from the root of the git repo : 
   ```sh
   docker run --rm --volume "$(pwd)/chart/inception:/helm-docs" -u $(id -u) jnorwood/helm-docs:latest --sort-values-order=file
   ``` 
* The values schema file can be re-generated with [the Helm `values-schema plugin`](https://github.com/losisin/helm-values-schema-json), from the root of the git repo : 
    ```sh
   helm schema -input chart/inception/values.yaml
   ``` 
* If you're OK with the merge request, we can discuss about updating the official INCEpTION documentation regarding the Kubernete's deployment section

PS @reckart : The CLAs are on their way, the holidays are adding a bit of delay. I think it's best not merging this MR until you have them, but I'll happily exchange about the feature until then :)


reckart:
Is there a simple way to test this using Docker Desktop?

reckart:
Looks like the PR includes a few spurious changes to binary files:

Screenshot 2024-07-30 at 10 10 38

chart/inception/README.md
```diff
@@ -0,0 +1,177 @@
...
+| auth.preAuthentication.enabled | bool | `false` | Enables the PreAuthentication feature. |
+| auth.preAuthentication.headerPrincipal | string | `""` | Name of the header holding the remote user name. |
+| auth.preAuthentication.newUserRole | string | `""` | Role given to new pre-authenticated users. |
+| auth.preAuthentication.customUserRoles | object | `{}` | Roles to map to dedicated pre-authenticated users, in the `username: role` format. |
+
+### INCEpTION deployment settings
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| replicaCount | int | `1` | Number of replicas fir INCEpTION deployment. |
```
reckart, line 106: Should something be said here that multiple replica using the same DB / filesystem location are not possible or does the replica count include replicas of the DB / FS as well?

kzgrzendek:
Thanks for your feedback!

> Is there a simple way to test this using Docker Desktop?

I think Docker Desktop have an option to run an integrated Kubernetes cluster (haven't used it for years). If so, you could probably just run the helm command as described in the PR once the cluster is booted up. Are you on Windows or Linux?

> Looks like the PR includes a few spurious changes to binary files:

 Yes, it's very weird! The checksum of those files is changing each time I'm switching branched - I'll have a look at it.

> Should something be said here that multiple replica using the same DB / filesystem location are not possible or does the replica count include replicas of the DB / FS as well?

Definitely a good idea. The deployment would not scale DB and FS - it could maybe be done probable with a statefulset, but I don't know how INCepTION would behave (is there a "cluster mode"?).
 For now, I'm gonna document this and disable this parameter.

reckart:
> PS @reckart : The CLAs are on their way, the holidays are adding a bit of delay. I think it's best not merging this MR until you have them, but I'll happily exchange about the feature until then :)

Yep 👍

kzgrzendek:
@reckart I have updated the documentation and shopping cart settings as indicated. I don't know why the checksums of some files change with each validation, are these files generated automatically?

reckart:
We don't have automatically generated files in the repo as far as I know. What is more likely is that your system does not recognize the files as binary files by default. It might be necessary to add the file endings to `.gitattributes`.

kzgrzendek:
> We don't have automatically generated files in the repo as far as I know. What is more likely is that your system does not recognize the files as binary files by default. It might be necessary to add the file endings to `.gitattributes`.

You were right, I added those extensions ti the `.gitattribute` file, pushed them again, and now the issue is gone - well done!

kzgrzendek:
About the helmchart, I think it could be nice to build it at each release and store it as a gh-pages artifact, allowing people to retrieve from git?

That process is documented [here](https://dev.to/jamiemagee/how-to-host-your-helm-chart-repository-on-github-3kd)

What do you think about this?

A saw a few projects like jupyterhub or argocd doing the same thing.

reckart:
The INCEpTION build and release process are fully based on Maven. If the helm charts need a build process, can it be integrated with the Maven build? The build must be able to run on Linux, Windows and macos.

kzgrzendek:
> The INCEpTION build and release process are fully based on Maven. If the helm charts need a build process, can it be integrated with the Maven build? The build must be able to run on Linux, Windows and macos.
> 
> It seems there are at least two Maven plugins related to helm:
> 
> * https://github.com/kokuwaio/helm-maven-plugin
> * https://github.com/deviceinsight/helm-maven-plugin
> 
> I am not familiar with either of them.

Ok, I'll have a look at this

kzgrzendek:
@reckart, I have a few questions abouts the packaging : 

The normal release cycle of a HelmChart includes (at least) the following steps :    
1. Linting 
2. Generating validation schema
3. Generating documentation
4. Packaging the chart (with correct `version` and `appVersion`)
5. Pushing the chart in a repository

More steps can be added to further improve the chart's 'stability', but actually those 6 steps are done manually.

Now, I have a few issues with the maven build :  
- None of the plugin you mentioned (nor the other ones I saw) allows to automatize those 6 steps
  - That remarks concerns the steps 2 and 3 in particular - how should we address that?
- Some plugins can only push charts to certain types of repositories (Nexus, Artifactory or Chart Museum)
  - Do you already have an artifact repository? If not, the simplest way imo would be to store it on `github-pages` as I described                                                                     [here](https://github.com/inception-project/inception/pull/4959#issuecomment-2263750885).
  
We should address these two points before going further imo.

PS : I'm not a Maven expert so I maybe missed something. I'd be glad to have your feedback on that.

reckart:
> None of the plugin you mentioned (nor the other ones I saw) allows to automatize those 6 steps
> That remarks concerns the steps 2 and 3 in particular - how should we address that?

If I am not mistaken, the plugins should be able to cover steps 1, 4, and 5 (step 6 seems to be missing from your list). 

If I read your testing instructions from the initial PR comment correctly, step 3 can be done by running a docker image. This could be done using the [Maven Docker Plugin](https://dmp.fabric8.io) or potentially using via Java Code using [testcontainers](https://testcontainers.com). Testcontainers would have the benefit of automatically being able to detect if a Docker deamon is running and being able to skip the documentation build if this is not the case. I don't want to always have to keep Docker running in the background for a build.

The maven helm plugin(s) seem to automatically download, install and run helm. I hope this works on all supported platforms (i.e. Windows, Mac, Linux). Once it has been downloaded, it might be possible to use the Maven Exec Plugin to execute the necessary commands to install the  help schema plugin and to execute the schema build command.

I'd prefer not maintaining generated stuff in the repository. Normally, I always generate auto-generatable stuff during builds only and exclude it from the git repository. Documentation gets uploaded to the GitHub pages as part of the release workflow. And likewise, artifacts get deployed to their respective repos as part of the release workflow. First, the Maven release is run which auto-deploys the Maven artifacts to the Maven repo. Then, I manually run the build in the `inception-docker` module to also deploy the docker image. I suppose, deploying the helm chart somewhere would be another extra manual step unless it can sensibly be integrated into the Maven release process.

We don't have an Artifactory, Nexus or ChartMuseum server for INCEpTION. For Maven artifacts, we use a different server and for Docker, we use GitHub. Can helm charts be package up as a ZIP file? If yes, we could deploy them to the Maven Repository. That's what I do with some NPM modules produced by INCEpTION. What other options are there? Are there public repositories similar to Maven Central, npmjs, pypi - but for Helm?

reckart:
> That process is documented [here](https://dev.to/jamiemagee/how-to-host-your-helm-chart-repository-on-github-3kd)

If I read that correctly, the idea is to use GitHub pages as a make-shift Helm repository in the absence of a central public repository. That could be an option. But in this case, it seems to me, the Helm charts should be maintained not in the INCEpTION repository but in a separate repository, e.g. `inception-helm` and they would then have their own build and release cycle as well. In that case, there would be no integration with the Maven build and the INCEpTION release cycle. Triggering the Helm chart release might be on the release checklist for the INCEpTION release, but it would not automatically take place.

In principle, it would be nice if the release process of INCEpTION would not get more complicated than it already is - i.e. that it would not include more manual steps.

reckart:
I have opened a repo here specifically for the helm chart: https://github.com/inception-project/inception-helm-chart

kzgrzendek:
@reckart I'm closing this PR as I moved it to the new repo you created -> https://github.com/inception-project/inception-helm-chart/pull/2

Thanks!

Issue #4941, opened 2024-07-13
Mapping OIDC groups to INCEpTION-s internal groups 


kzgrzendek:
Hello,

I would like to suggest a feature regarding the mapping of OIDC/OAUTH2 groups (when this authentication mode is enabled) to the internal groups of INCEpTION (`ADMIN`, `USER`, etc.)

## Context

I'm working for the Greater Paris University Hospitals, and we would like to propose INCEpTION as our go-to text annotation tool for our users, as part of our ['Health Data Warehouse' offer](https://eds.aphp.fr/qui-sommes-nous/nos-objectifs) (in French only, unfortunately).

As you can guess, we are tied by very strict legal and industrial obligation, one of them being the obligation to federate the authentication and authorization of our userbase when we provide services. 

One of them implicates that the applications we're deploying, if they need 'internal groups', should be able to map those with pre-existing groups defined in our SSO solution (Keycloak, in our case).

The go-to standard for for doing that being to use OIDC/OAUTH2 to fetch those groups in the user's token or with the `userInfo` endpoint, that's the reason why I'm proposing this feature, and I guess we're not the only one with those obligation.

## Description of the feature
If we simply look at a solution like JupyterHub, we can see that [its OIDC module](https://oauthenticator.readthedocs.io/en/latest/tutorials/provider-specific-setup/providers/generic.html#setup-for-an-openid-connect-oidc-based-identity-provider) is proposing to set : 
- The name of the claim containing the groups (`c.GenericOAuthenticator.claim_groups_key`)
- The groups allowed as users (`c.GenericOAuthenticator.allowed_group`)
- The groups allowed as admins (`c.GenericOAuthenticator.admin_groups`)

Implicitly, all users not members of the groups are denied by the application.

## Edge cases to bear in mind
I can think of a fez edge cases, but you're welcome to add yours as well : 
- If a user is in several groups, granting him different access level to the application, he should have the highest possible access level
  - eg, if a user can be mapped to `USER` and admin `ADMIN` groups, he should me mapped as `ADMIN` (unless INCEpTION supports having users un several groups!)
- If a user can't be mapped to any group, he should have his access denied (returning à HTTP 403 code)
- If a user is promoted to a group, it should be reflected in the application
  - eg, if a user was in an group mapped to `USER`, and is now in a group mapped to `ADMIN`, he should appear as an `ADMIN` in INCEpTION, and no more as a `USER`

I'm hoping you'll consider this feature, and am of course willing to discuss about it  :)

[...]

PR #4982, opened 2024-08-07
#4941 - Mapping OIDC groups to INCEpTION's internal roles

kzgrzendek:
**What's in the PR**
This PR contains the feature allowing to map OAuth2 groups with INCEpTION roles, as described in the [corresopnding feaure request](https://github.com/inception-project/inception/issues/4941)

**How to test manually**
1. Run a local Keycloak, with a dedicated user and 4 groups associated to that user : 
  - `INCEPTION_ADMIN`
  - `INCEPTION_USER`
  - `INCEPTION_PROJECT_CREATOR`
  - `INCEPTION_REMOTE`
2. Add the following lines to your `settings.properties` file, inside your INCEpTION home directory : 
    ```
    # OAuth2 groups mapping settings
    oauth2-groups.enabled=true
    oauth2-groups.admin=/INCEPTION_ADMIN
    oauth2-groups.user=/INCEPTION_USER
    oauth2-groups.project-creator=/INCEPTION_PROJECT_CREATOR
    oauth2-groups.remote=/INCEPTION_REMOTE
    ```
3. Build and run an INCEpTION instance from this branch
4. Try to login via OAuth2, and observe that the newly created user is bind to all the INCEpTION roles, as mapped in the `settings.properties` file. You can play with the mapping/groups to test the behavior.

**Note :**  :
-  The roles are mapped at the user's creation and at every connections
-  If the feature is enabled and the user isn't in any group mapped to INCEpTION roles, and AccessDeniedException will be throw.
- The `OAuth2AdaptaterImpl` class was vefore calling to the `PreAuthUtil` class, whereas it's specified in the documentation that the `External PreAuth` feature isn't compatible with OAuth2 authentication. I made a dedicated `OAuth2Utils` class that contains a method that maps the OAuth2 groups if the mapping feature is enabled, or just adds the `USER` INCEpTION role to the user if the  mapping feature is disabled.
- Corresponding properties have been added with a native SpringBoot `@Configuration` annotation.

**Automatic testing**
- The test class `OAuth2UtilsTest` contains a set of unit tests

* [x] PR includes unit tests

**Documentation**
- I can do the documentation as well if the feature is validated.

I'm of course open to feedback and willing to modify the implementation if needed ;)


inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.AccessDeniedException;
+
+import de.tudarmstadt.ukp.clarin.webanno.security.model.Role;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+@Configuration
+@ConfigurationProperties(prefix = "oauth2-groups")
+public class OAuth2Utils {
```
reckart, line 36: Utils classes in INCEpTION are typically final classes containing only static methods. Configuration classes are called `XxxPropertiesImpl`, e.g. `KnowledgeBasePropertiesImpl` and come with a respective interface, e.g. `KnowledgeBaseProperties`. This class/interface separation is necessary in cases where the properties are injected into Wicket code.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+    
+    private static boolean OAUTH2_GROUPS_ENABLED;
```
reckart, line 38: Fields should be non-static and use camelCase.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    @Value("${oauth2-groups.remote:}")
+    public void setRemoteGroup(String remoteGroup)
+    {
+        OAUTH2_REMOTE_GROUP = remoteGroup;
+    }
+
+
+    public static Set<Role> getOAuth2UserRoles(User aUser, ArrayList<String> oauth2groups)
```
reckart, line 75: Arguments should use interfaces for collections. e.g. `List` instead of `ArrayList`.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+        throws AccessDeniedException
+    {
+        Set<Role> roles = new HashSet<>();
```
reckart, line 78: Use `var` when possible.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+            roles.add(Role.ROLE_USER);
+            return roles;
+        }
+
+        if (oauth2groups == null || oauth2groups.isEmpty()) {
+            throw new AccessDeniedException("OAuth2 groups mapping is enabled, but user ["
+                + aUser.getUsername() + "] doesn't have any groups, or the corresponding claim is empty");
+        }
+
+        oauth2groups.forEach(group -> matchOauth2groupToRole(group, roles));
```
reckart, line 90: Functional programming (without side-effects) would be easier to understand. Can we use `map` instead of `forEach` here?

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+        throws AccessDeniedException
+    {
+        Set<Role> roles = new HashSet<>();
+
+        if (!OAUTH2_GROUPS_ENABLED) {
+            roles.add(Role.ROLE_USER);
+            return roles;
+        }
+
+        if (oauth2groups == null || oauth2groups.isEmpty()) {
```
reckart, line 85: Maybe just return the empty list here.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+            throw new AccessDeniedException("OAuth2 groups mapping is enabled, but user ["
+                + aUser.getUsername() + "] doesn't have any groups, or the corresponding claim is empty");
+        }
+
+        oauth2groups.forEach(group -> matchOauth2groupToRole(group, roles));
+
+        if (roles.isEmpty()) {
```
reckart, line 92: I am not sure if this check is necessary at all since the application should only set users in that have ROLE_USER. If it is necessary, then it should IMHO be at a higher level, not at the level where the groups are extracted from the claim.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+            throw new AccessDeniedException("User ["
+                + aUser.getUsername() + "] doesn't belong to any role");
+        }
+
+        return roles;
+    }
+
+    private static void matchOauth2groupToRole(String oauth2group, Set<Role> userRoles) {
+        
+        if (StringUtils.equals(oauth2group, OAUTH2_ADMIN_GROUP)) {
```
reckart, line 102: With the hard-coded roles here, we'd have to update the code in case a new role is added.

How about setting up a map of inception roles and external roles and make that map configurable through the externalized properties?

inception/inception-app-webapp/src/test/java/de/tudarmstadt/ukp/inception/db/InceptionHsqldbIntegrationTest.java
```diff
@@ -55,6 +58,8 @@ static void tearDownClass()
     }
 
     @Nested
+    @ContextConfiguration
```
reckart, line 61: (Why) did you have to add these annotations here?

inception/inception-security/src/test/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2UtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import de.tudarmstadt.ukp.clarin.webanno.security.model.Role;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = OAuth2Utils.class)
+@DirtiesContext(classMode = ClassMode.AFTER_CLASS)
+@TestPropertySource(
+        properties = """
```
reckart, line 45: Better use `@DynamicPropertySource` to configure the properties programmatically, that allows you to use constants and this way facilitate refactoring.

inception/inception-security/src/test/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2UtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+    
+    @Test
+    void thatUserRoleIsGivenIfMatchingGroupFound()
+    {
+        ArrayList<String> userOAuth2Groups = new ArrayList<>();
+        userOAuth2Groups.add(OAUTH2_GROUP_USER);
+        
+        Set<Role> userRoles = OAuth2Utils.getOAuth2UserRoles(testUser, userOAuth2Groups);
+        
+        assertTrue(userRoles.contains(Role.ROLE_USER));
```
reckart, line 88: Please use AssertJ asserts instead of Unit asserts.

inception/inception-security/src/test/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2UtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+        
+        assertTrue(userRoles.contains(Role.ROLE_REMOTE));
+    }
+    
+    @Test
+    void thatUnauthorizedExceptionIsThrownIfNoRoleIsMapped() {
+        
+        ArrayList<String> userOAuth2Groups = new ArrayList<>();
+        
+        try {
```
reckart, line 118: `assertThatExceptionOfType(...).isThrownBy(() -> { ... })`

inception/inception-security/src/test/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2UtilsTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.springframework.test.context.TestPropertySource;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import de.tudarmstadt.ukp.clarin.webanno.security.model.Role;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = OAuth2Utils.class)
+@DirtiesContext(classMode = ClassMode.AFTER_CLASS)
```
reckart, line 43: Instead of making this a spring integration test, I'd tend towards it being a plain junit test that sets up the necessary beans as POJOs. It seems a bit of overkill to start up an entire spring context for this.

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.access.AccessDeniedException;
+
+import de.tudarmstadt.ukp.clarin.webanno.security.model.Role;
+import de.tudarmstadt.ukp.clarin.webanno.security.model.User;
+
+@Configuration
+@ConfigurationProperties(prefix = "oauth2-groups")
```
reckart, line 35: `security.oauth2.roles...`? Other INCEpTION security properties live under `security...`. Mind, INCEpTION has no groups, only roles. It is incidental that you call these labels `groups` on your side. So from the INCEpTION perspective, we are mapping roles, not groups.

```
security.oauth.roles.enabled=true
security.oauth.roles.claim=groups
security.oauth.roles.mapping.ROLE_USER=ad-user
security.oauth.roles.mapping.ROLE_ADMIN=ad-admin
security.oauth.roles.mapping.ROLE_PROJECT_CREATOR=ad-project-creator
security.oauth.roles.mapping....=...
```

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2AdapterImpl.java
```diff
@@ -148,6 +147,10 @@ private User fetchOrMaterializeUser(OAuth2UserRequest userRequest, OAuth2User us
         if (u != null) {
             denyAccessToDeactivatedUsers(u);
             denyAccessOfRealmsDoNotMatch(realm, u);
+
+            u.setRoles(OAuth2Utils.getOAuth2UserRoles(u, user.getAttribute("groups")));
```
reckart, line 151: The claim should be configurable via the properties as well since this is not a well-defined name. Others might e.g. use a `roles` claim instead.

reckart (commented):
Thanks for the suggestion! I hope you find my feedback useful.

reckart:
Btw. the commit message convention for INCEpTION goes like this:

```
#ISSUE - Title of the issue

- change 1
- change 2
- change 3
```

inception/inception-security/src/main/java/de/tudarmstadt/ukp/inception/security/oauth/OAuth2Utils.java
kzgrzendek: @reckart how would you recommend to proceed? Creating the interface and implementation class for the properties, injecting the PropertiesImpl class in OAuth2AdapterImpl's constructor and moving the mapping logic in a private method of that class?

inception/inception-app-webapp/src/test/java/de/tudarmstadt/ukp/inception/db/InceptionHsqldbIntegrationTest.java
kzgrzendek: Probably a side effect of a refactor, I didn't changed that class - will revert that

reckart:
The PR build fails but I have no idea why. I have opened a support discussion: https://github.com/orgs/community/discussions/138018

reckart:
I have create a new PR for this one since on this one here the GH actions always fail... https://github.com/inception-project/inception/pull/5453

PR #6093, opened 2026-06-05
#6095 - Auto-login causes an infinite redirect loop under a non-root server.servlet.context-path

ogirardot:
## The issue
With `security.login.auto-login=<registrationId>` set, INCEPTION enters an infinite 302 redirect loop **when it is deployed under a non-root servlet context path** (`server.servlet.context-path=/inception`). The browser bounces
forever between:
- `https://<host>/inception/inception/oauth2/authorization/<registrationId>`  ← note the **doubled** `/inception/inception/`
- `https://<host>/inception/login.html`

The manual "Login with <provider>" button on `/login.html` works correctly; only the automatic redirect is broken.

### To Reproduce

1. Deploy INCEPTION behind a reverse proxy with `server.servlet.context-path=/inception`.
2. Configure OAuth2 login (`spring.security.oauth2.client.registration.<id>...`). The manual login button works.
3. Set `security.login.auto-login=<id>`.
4. Open `https://<host>/inception/login.html` → infinite redirect loop on `/inception/inception/oauth2/authorization/<id>`.

### Expected behaviour

Auto-login redirects once to `/inception/oauth2/authorization/<id>` and proceeds through the normal OAuth2 flow, exactly as the manual button does.

## The PR 

**What's in the PR**

This PR adds a test to reproduce the behavior and fixes it in Wicket's RedirectRequestHandler#respond()

**How to test manually**
Launching mvn test will test the behavior and the fix, otherwise you can use the steps to reproduce the issue.

**Automatic testing**
* [X] PR includes unit tests

**Documentation**
* [X] PR does not updates the documentation because it should just work

### DIFF
--- a/inception/inception-io-xml/src/main/java/de/tudarmstadt/ukp/inception/io/xml/CustomXmlFormatFactory.java
+++ b/inception/inception-io-xml/src/main/java/de/tudarmstadt/ukp/inception/io/xml/CustomXmlFormatFactory.java
@@ -73,9 +73,10 @@ public class CustomXmlFormatFactory
         }
 
         stylesheetReferences = new ArrayList<ResourceReference>();
+        var basePath = description.getBasePath().normalize();
         for (var stylesheet : description.getStylesheets()) {
-            var stylesheetPath = description.getBasePath().resolve(stylesheet);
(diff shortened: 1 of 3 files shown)


reckart:
@ogirardot thanks for the fix, in particular also with a test 👍

PR #6096, opened 2026-06-06
Await displayViewer

tejas-ae:
The behavior in displayViewer might be relying on something a little implicit. The promise result gets used as if it were already resolved. this patch adds the missing await so the async path is actually sequenced. Sorry if this is intentional.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


reckart:
The code should be correct as it is. I'm closing this. If you ran into an actual problem with the code, please open a bug report explaining the issue and how to reproduce it.

Thanks!

